PR #140, opened 2016-04-12
Fixed paths to new reference case location

gabriel-happle:
Fixed paths to new reference case location


daren-thomas:
congratulations on first merged pull request @gabriel-happle :smile:

PR #143, opened 2016-04-13
I114 update scripts part 1

JIMENOFONSECA:
LCA embodied working


daren-thomas:
shall we merge it? have you tested it well?

JIMENOFONSECA:
What do you mean "tested well?" Is there a way to test it bad? Please elaborate if there is an specific way...

Sent on the road

daren-thomas:
yes. badly: just test the part you changed. push it to master. break the tool for everyone else.

good: first always update from master. then test the script with the test script. is the output correct? can others verify the output is correct? (communicate what the output should be)

does the other functionality still work? (i will create a master test script to run on a separate computer once a day to make sure master is always running). did you run the script from arcgis?

so. basically this. we are still a young project so this hasn't been formalized yet. we will probably write down a test procedure in the dev docs soon. 

remember the time i refactored and the output was wrong and you had to roll back my changes? i'm trying to avoid this in the future...

JIMENOFONSECA:
then it was well done :-), however as part of the board i believe we should be mutually be exchanging this task of merging and re-testing.

i.e., if you do and pull a request. you should have tested it in your machine. To cross-check i will test it in mine before accepting your pull request. if everything is fine then i approve your merge. this is only for what modifies the core. for new things we are testing, we use the SandBox instead in the master.

we do this to each other's code. then we avoid what happened last time. It completely worked in your machine but not in mine.

what do you think?

daren-thomas:
i think this is a very good plan. let's call it the four-eyes-principle!

daren-thomas:
and write it down in the dev docs!

daren-thomas:
and write it down in the dev docs!

daren-thomas:
Jimeno, I can't merge this, because of merge conflicts that I can't resolve. I don't know which branch contains the newest version of solar_radiation_vertical... You know what? I'm just going to assume that the version in master is newer than the one on the branch for the merge...

PR #148, opened 2016-04-15
I135 update toolbox new reference case

daren-thomas:
updated scripts to use the `InputLocator` (simplify paths, toolbox)
- lca_embodied
- properties
- demand

some scripts still need to be updated to use the shapefiles, so I could not update those yet:
- lca_emissions
- heatmaps


JIMENOFONSECA:
It looks beautifully organised this way, i am ready to merge it. should I? or should i fix first the input to shapefiles of the remaining scripts?

daren-thomas:
thank you :) yes, I think the merge now is better than waiting and then everybody can get on with life... @shanshanhsieh is also waiting for the fix for the MemoryError that is part of this...

PR #152, opened 2016-04-21
this should solve the MemoryError shanshan is having

daren-thomas:
@JIMENOFONSECA, @shanshanhsieh: I believe this change could fix shanshan's `MemoryError` problem. My strategy: Instead of summing all 8760 columns at once, group/sum them separately and then create the (much smaller) `DataFrame`.

Initial tests on @shanshanhsieh's computer show that they work, but TeamViewer decided to hang up on me so I finished the work on my computer.

I have run both versions and the results seem identical. @JIMENOFONSECA, you might want to check yourself?

I propose the OBM (Other Board Member) merges this pull request.


shanshanhsieh:
it is working on my side now. thanks, @daren-thomas !

JIMENOFONSECA:
i am testing it.

JIMENOFONSECA:
the results are equal to the previous version

PR #162, opened 2016-04-26
I159 controlsystems

JIMENOFONSECA:
this partly solves out the problem


JIMENOFONSECA:
@daren-thomas this branch changes the demand and inputs. While testing, it is necessary to also reload the reference-case data from the other repository

PR #170, opened 2016-05-02
implemented toolbox for heatmaps

daren-thomas:
and some fixes in the heatmaps script


JIMENOFONSECA:
The results up to now need some modifications. @daren-thomas we can discussed them on Slack

PR #172, opened 2016-05-03
I171 refactor calcthermalloads

gabriel-happle:


daren-thomas:
this runs through on my machine! I am going to add the `gv.models` approach we talked about so you have an entry point for your models and then merge the pull...

daren-thomas:
OK. It turns out I can't add the `gv.models` approach we talked about, because: `functions.py` now imports directly from `contributions`. I don't think that is a good idea. Neither does python... what should we do about this? Is the storeage tank part of core?

We need to rethink the folder structure of the project. I would like to postpone this until after the workshop...

gabriel-happle:
But this means that we can not write our own CalcThermalLoad functions for now. I guess we are stuck.

daren-thomas:
thanks @shanshanhsieh & @gabriel-happle, good job!

PR #176, opened 2016-05-04
added `log` method to `GlobalVariables`

daren-thomas:
Using this technique, it is possible to change the `print` statements in the scripts to `gv.log` and they will print just the same.

BUT: When run from the toolbox, the `log` method can be replaced with the new `add_message` which will use `arcpy.AddMessage` instead: voila! We now have logging in arcgis for long-running tasks!


daren-thomas:
so far implemented for `graphs` and `demand`

PR #194, opened 2016-05-13
I192 full reports

gabriel-happle:
extensive report function in the contribution. controlled via a parameter in globalvar


cea/functions.py
```diff
@@ -2,13 +2,14 @@
...
-import numpy as np
 import os
+
+import numpy as np
 import pandas as pd
 import scipy
 import scipy.optimize as sopt
-import sandbox.ghapple.reporting as reporting
 
+import contributions.reporting as reporting
```
daren-thomas: please do not import from contributions or sandbox in core! due to the way the python packaging system works, these create a lot of headache...

cea/functions.py
```diff
@@ -633,6 +634,20 @@ def CalcThermalLoads(Name, prop_occupancy, prop_architecture, prop_thermal, prop
                    Tww_sup_0, Waterconsumption, locationFinal, mcpcs, mcphs, mcpww, path_temporary_folder,
                    sys_e_cooling, sys_e_heating, waterpeak)
 
+    if gv.testing:
+        reporting.full_report_to_xls(Name, Af, Aef, sys_e_heating, sys_e_cooling, Occupants, Am, Atot, Awall_all, Cm,
```
daren-thomas: no. please keep function signatures (parameter lists) to max five parameters.

contributions/reporting.py
```diff
@@ -0,0 +1,80 @@
...
+    """ this function is to write a full report to an *.xls file containing all intermediate and final results of a
+     single building thermal loads calculation"""
+
+    # TODO: get names of used functions from glovalbars for report
+    # TODO: write units to column names
+    # TODO: split outputs into more work sheets, e.g. one for water, electricity, etc
+
+    # get date and time
+    now = datetime.datetime.now()
+    string_now = ("%s-%s-%s-%s-%s-%s" % (now.year, now.month, now.day, now.hour, now.minute, now.second))
```
daren-thomas: you can actually format a time (look up `datetime` module, it might be in there) - use the pythonic way to do this!

contributions/reporting.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # create excel work book and work sheets
+    filename = location_final + '\\' + name + '-' + string_now + '.xls'
```
daren-thomas: please use `os.path.join` to create file paths!

contributions/reporting.py
```diff
@@ -0,0 +1,80 @@
...
+def full_report_to_xls(name, area_f, area_ef, sys_e_heating, sys_e_cooling, occupants, am, atot, awall_all,
+                       cm, ll, lw, retrofit, sh_typ, year, footprint, nf_ag, nfp, lcww_dis, lsww_dis, lv, lvww_dis,
+                       tcs_re_0, tcs_sup_0, ths_re_0, ths_sup_0, tww_re_0, tww_sup_0, y, fforma, ta_hs_set, ta_cs_set,
+                       people, ve_schedule, q_int, eal_nove, eprof, edata, qcdataf, qcrefirf, vww, vw, qcdata, qcrefri,
+                       qv_req, hve, htr_is, htr_ms, htr_w, htr_em, htr_1, htr_2, htr_3, i_sol, i_int_sen, w_int, i_ia,
+                       i_m, i_st, uncomfort, ta, tm, qhs_sen, qcs_sen, qhs_lat, qcs_lat, qhs_em_ls, qcs_em_ls, qhc_sen,
+                       ma_sup_hs, ta_sup_hs, ta_re_hs, ma_sup_cs, ta_sup_cs, ta_re_cs, w_sup, w_re, ehs_lat_aux,
+                       qhs_sen_incl_em_ls, qcs_sen_incl_em_ls, t5, tww_re, top, im_tot, thset_corr, tcset_corr, tcs_re,
+                       tcs_sup, ths_re, ths_sup, mcpcs, mcphs, mww, qww, qww_ls_st, qwwf, tww_st, vw_large, vww_large,
+                       mcpww, eaux_cs, eaux_fw, eaux_ve, eaux_ww, eauxf, occupancy, waterconsumption, location_final):
```
daren-thomas: notice how `location_final` got lost in the parameters list? This is why we do short parameter lists!

contributions/reporting.py
```diff
@@ -0,0 +1,80 @@
...
+    # create excel work book and work sheets
+    filename = location_final + '\\' + name + '-' + string_now + '.xls'
+    wb = xlwt.Workbook()
+    wb.add_sheet('Simulation properties')
+    wb.add_sheet('Building properties')
+    wb.add_sheet('Thermal loads hourly')
+    wb.save(filename)
+
+    # create data frames
+    df2 = pd.DataFrame({'Name': name, 'Af_m2': area_f, 'Aef_m2': area_ef, 'Heating system': sys_e_heating,
```
daren-thomas: `df2`? I just read we're creating sheets like "Simulation properties", "Building properties", "Thermal loads hourly" - surely we can name the dataframes according to what data is stored inside, right? `df_simulation_properties` or `building_properties_df` - we should decide on where to put the `df` (prefix or postfix and use that for the whole project.

daren-thomas:
I would prefer not to merge this as is. See comments in the review.

Idea: Instead of creating an uber-function with a list of parameters to report, use `locals()` and accept a dictionary of functions to report. `locals()` is a dictionary containing all local variables.

Next, why `gv.testing = True`? Let's do 
```
gv.report_variables = { 
    'calc-thermal-loads': ['Name', 'Af', 'Aef', 'sys_e_heating', 'syse_cooling'], 
    'other-module-that-needs-logging': ['v1', 'v2', 'v3']
}

Then, we can do `report(location_final, gv.report_variables['calc-thermal-loads'], locals())`

But even that can probably be done nicer (why have location_final hardcoded in the function?)

`report('calc-thermal-loads', locals())` - `report` now has to look up the variables to write and also the configured location (maybe from `gv.report_locations = {'calc-thermal-loads': r'c:\blahblahblah'}`

Keep the `if gv.testing` test inside the report function!
```

BTW: I see you modelled the function after the `results_to_csv` function. That explains the design you chose. I would like us to improve the design of `results_to_csv` too...

cea/functions.py
```diff
@@ -2,12 +2,14 @@
...
 import math
-import numpy as np
 import os
+
+import numpy as np
 import pandas as pd
 import scipy
 import scipy.optimize as sopt
 
+import contributions.reporting as reporting
```
daren-thomas: please don't import from contributions or sandbox... hm... ok. I see your problem - we have to remove that when merging the pull request back into core...

gabriel-happle:
Hey @daren-thomas and @JIMENOFONSECA
I see all of your points and could not agree more! I can change these things in the future, when I have time to research a bit more about advanced python programming.

But can you please merge my pull request anyways? I want to compare these results generated in my branch with these results generated in the master.
I see no other way how this could be done, except with a merge to the master and then update my branch from the master.

contributions/reporting.py
gabriel-happle: What is the pythonic way? Default formats of `datetime` strings include `.` or `:`, which can not be part of file names.

contributions/reporting.py
gabriel-happle: OK

contributions/reporting.py
gabriel-happle: Agree.
I vote for prefix.
And we should decide soon.

contributions/reporting.py
gabriel-happle: Agree

gabriel-happle:
most of the comments implemented now

gabriel-happle:
What about this?

daren-thomas:
thanks, @gabriel-happle - that was some good work! I have added some (small) changes and have some questions / comments:

- what does "TODO: get names of used functions from globalvars for report" mean?
- how would you like the units to be represented? (spec this out!) - I suggest adding them to the report_variables dictionary (instead of just variable name, use a tuple `(var_name, var_units)`. Maybe keep that optional and test with `isinstance(var, tuple)`?
- splitting outputs into more sheets is easy now and can be done in the `report_variables` dictionary.

PR #195, opened 2016-05-16
I181 schedulemaker

JIMENOFONSECA:
This pull request creates an schedule maker and joins all the databases of archetypes in one excel file. This is better for the user. This request includes fixes to gv.list_uses and improves the speed of the code while reading only the occupancy types used by the user (>0)

!WATCH OUT!. this request modifies de calc_thermal function at the beginning and the reference case. please sync as required.


JIMENOFONSECA:
Dear @daren-thomas, here there is a new one, you will need to also use my pull request of the reference case to test.

daren-thomas:
@JIMENOFONSECA, I am now giving up. I can't get this to run, I think you need to fix it first.

After fixing a bug with `np.vectorize` that only showed up on building 74 (I think), I was able to run this branch all the way to "Building No. 270 completed out of 274".

The problem lies in the call to `calc_loads_electrical`: `Eal_nove`, `Edataf` and `Eprof` are not always set! (case `Af == 0`)

I have checked the history of the change and it seems you introduced `get_internal_loads` three days ago. We used to use `calc_mixed_schedule`, above the `if Af > 0` line, so that meant that `Eal_nove` et al. were set.

Please make sure this pull request runs through, before re-submitting.

JIMENOFONSECA:
@daren-thomas noted. Now it works for all buildings. please check.

PR #197, opened 2016-05-17
I191 solar to radiation

JIMENOFONSECA:
This modification creates a new file called properties_surface.csv. which will be used in the demand file


JIMENOFONSECA:
@daren-thomas this is a subbranch of #195. 
@gabriel-happle gv.get_surfaces() is the file you should call to get properties of freeheight, Shape_Leng etc.. of every facade in the building.

PR #199, opened 2016-05-17
I137 radiation script

daren-thomas:
Updated for use in ArcGIS as a toolbox item.


daren-thomas:
@JIMENOFONSECA: please check this pull request - I have implemented the radiation script as a toolbox for ArcGIS. It uses `fiona` to figure out latitude and longitude from the `building_geometry.shp` file and tries to figure out the timezone too!

JIMENOFONSECA:
The branch works fine in regard to the tool's interface, however the parallel pull request #197 makes extensive changes of the radiation file too. What should we do to avoid a merge conflict? It seems imminent to me.

daren-thomas:
@JIMENOFONSECA this is the next pull request in the pipeline. sadly, due to changes from the last two pull requests, it does not currently work. I am fixing it as we speak. PLEASE WAIT FOR OK TO MERGE!

daren-thomas:
@JIMENOFONSECA try this on your computer. it starts just fine on mine, but arcscene ends up crashing halfway through (day 116)...

daren-thomas:
It runs! I think there is some race condition in `arcpy`... I ran it about 3-4 times and sometimes arcscene crashes, sometimes it doesn't.

@JIMENOFONSECA could you check out this branch, see if it runs for you and then merge the pull request? Then I can get on with the .EPW pull request (which will also be some work as the `radiation.py` file has changed a lot....)

PR #200, opened 2016-05-18
Use of .Epw datafiles

JIMENOFONSECA:
It requires the upgrade to radiation.py and demand.py interfaces in Arcgis. It will increase user functionality and discard weather data as an input


JIMENOFONSECA:
i will create a new one..

PR #216, opened 2016-05-25
I215 buxfix ve schedule

gabriel-happle:
Bugs fixed in ventilation schedule and occupancy calculation


gabriel-happle:
@JIMENOFONSECA please merge

PR #223, opened 2016-05-27
I221 class for calc loads input

gabriel-happle:
@daren-thomas 
like this?
I grouped input arguments to calc thermal loads in a class containing building properties and two dicts containing weather and schedule data.
I made new files directly in the core. Hope that is ok.


daren-thomas:
Pretty much exactly what I was expecting. Some changes I performed:
- moved the class to the demand script (keep related stuff close together!)
- don't bother returning `False` if the dataframe is not set: that just moves the error away from where it happened, making it harder to debug!
- don't set the dataframes to `0`. why even bother? that isn't the type they are going to have - if you have no value, use `None`
- also, you set them as class variables. setting them in `__init__` makes them instance variables
- moved comments above methods to docstrings inside methods. this is how to document python methods.
- renamed `BuildingPropsThermalLoads` to `BuildingProperties`
- we should rename away from `prop_geometry` to either just plain `geometry` or `geometry_df`. Also, inside the calc-thermal-loads functions, we actually use `geometry_row` or something like that... maybe just call the function `get_geometry(self, building_name)`
- there is an issue open to move the `CalcThermalLoads` function into the demand script. Maybe soon is a good time...

Questions:
- do we really need to use a dict of `np.array` for the weather? is it that much of a bottleneck?

daren-thomas:
@JIMENOFONSECA can you test this and merge please?

BTW: I think this is a brilliant start. I also think it is only a start - we will take this quite a bit further, organically "growing" an architecture out of the fertile humus that is Jimeno+...

gabriel-happle:
@daren-thomas 
Thanks for the feedback! I'm learning a lot here.
I'm not sure I understand your question. Do you mean read in the complete .epw file and use that? Yes, we could totally do that. I think it's a good idea.

daren-thomas:
@gabriel-happle, we already read in the complete `.epw` file. You were then splitting it up into `np.array`s. I thought maybe because of performance reasons?

gabriel-happle:
@daren-thomas, I was just following the previous approach. No performance reasons that I'm aware of.

PR #229, opened 2016-06-03
Scenario plots

daren-thomas:
Adds a new tool to the toolbox "Scenario Plots"


JIMENOFONSECA:
Nice stuff!, just two comments:
- In this case, Boxplot diagram works fine only if we compare relative values (e.g,., ghg_kgm2) but not totals ghg. i suggest to take the totals out (e.g., ghg_ton, Ealf_MWhyr)
- The scenario totals for relative values (e.g,., ghg_kgm2) should be a weighted average but not a sum as not every building has the same area.
- Potentially we could have LCA_embodied, LCA_operation, LCA_mobility and LCA_total.

daren-thomas:
These are really good comments! I'll get cracking!

I do not understand the last point (LCA_embodied, LCA_operation, LCA_mobility and LCA_total) - what do you mean?

JIMENOFONSECA:
so we have two pages: one for LCA_embodied and one for LCA_operation, there is also LCA_mobility and LCA_total which sums all the relative values = LCA_embodied + LCA_operation + LCA_mobility
@martin-mosteiro is calculating those

daren-thomas:
Does this look better?

[scenario-plot2.pdf](https://github.com/architecture-building-systems/CEAforArcGIS/files/300322/scenario-plot2.pdf)

PR #230, opened 2016-06-03
I182 benchmark plots

martin-mosteiro:
Benchmark graphs for the 2000 Watt society case for Switzerland or any other benchmark in the future.


JIMENOFONSECA:
mobility.py results are 0, i guess we will have to calc first the main use of the building and then do a merge based on the Code. See emissions.py approach.

JIMENOFONSECA:
Somehow the diagram of operations is not totally displayed so the mean value does not appear in the graph. i made the diagram manually in excel for the sake of comparison.


capture

martin-mosteiro:
Yeah, I redefined the axes on all the graphs to try to make them a bit more understandable (otherwise the outlier you see in your graph makes it hard to compare data). However, I can see now how that would be a problem as in the case you pointed out, so it’s best to take it out.

About the mobility, that’s weird… It worked perfectly when I ran it on my computed. I’ll look into it now.

JIMENOFONSECA:
cool, martin, i also added GFA_m2 value (gross floor area) to the emissions and embodied energy calculations. this will make easier when you compute the total per scenario. I just put it in the master version. you will need to run the demand first

JIMENOFONSECA:
you will need to update from master and solve conflicts

JIMENOFONSECA:
@martin-mosteiro @daren-thomas i just uploaded a new version of the total demand file and all the LCA files. so you guys do not have to run the demand, emissions and operations to continue with your graphs

daren-thomas:
@JIMENOFONSECA, can you give a go-nogo for this by tomorrow morning?

PR #231, opened 2016-06-06
Fix units and results for dhw storage

shanshanhsieh:
In Calc_ThermalLoads, the dhw return temperature in db and unit for mcpww is fixed, documentation for calc_dhw_heating_demand was added.


daren-thomas:
must this be added for the conference?

shanshanhsieh:
yes... otherwise some part of the results will not make sense.
but please wait until I finish testing Jimeno's new commit!

PR #239, opened 2016-06-08
I016 2 simplified new ventilation and thermal loads

gabriel-happle:


JIMENOFONSECA:
@gabriel-happle could you check it once more time the whole simulation? I did it in my machine and it words perfectly fine using test.py. the results are also realistic 

@daren-thomas the default of calc thermal loads was changed to gabriel's model. It makes substantial improvements to the outputs of the CEA tool (in terms of accurracy of the demand model) 

@amraladdin this solves the ventilation problem and includes an easy way to consider leaky or non leaky buildings (changes to n50 parameter in the architecture.shp)

@daren-thomas and @martin-mosteiro  i unfortunately erased the output option of the benchmark and scenario plots. Daren told me about their importance. If needed could you please put them back?

PR #252, opened 2016-07-13
Thermal loads with dataframes (and parallelization)

daren-thomas:
This is the branch I'm doing my refactoring on. I would like this merged to master before I go on vacation so that you can work on this code while I'm gone and we all avoid merge conflicts later on...

As a bonus incentive I have added parallelization of the `calc_thermal_loads_new_ventilation` code.

**Please check output of this branch to output of master branch! It should be identical!!**


JIMENOFONSECA:
Not working....

capture

PR #266, opened 2016-08-09
Port to ArcGIS 10.4

daren-thomas:
Test run compares results of running master on ArcGIS 10.3.1 with the HQ reference-case with this branch on ArcGIS 10.4 - identical results.

Did not run Radiation script...

Changes to installation script will be necessary...


JIMENOFONSECA:
@daren-thomas please check radiation script as this is the one most influenced by the arcpy library. I will check this pull request untill

daren-thomas:
@gabriel-happle: can you try this branch out? You said you would be installing 10.4 this week anyway... There is a branch of the contributors manual with the same name that explains how to install the libraries...

daren-thomas:
I have successfully run the radiation script on this version. It runs through. I would like an independent verification and have asked @gabriel-happle to check it on his machine.

daren-thomas:
@gabriel-happle: I have compared the radiation output from both versions (master vs. this branch) and the output is identical on my machines.

daren-thomas:
@JIMENOFONSECA, according to testing from @gabriel-happle and myself, I think this can be merged to master. I was hoping it would be backwards compatible with 10.3.1 but that does not seem to be the case, so we should all update to 10.4.

Please merge this pull request.

PR #278, opened 2016-08-15
I277 bugfix thermal load

gabriel-happle:
bug fixed.
script works.
but very different results!


daren-thomas:
Are they now more correct or less correct? How could we go about verifying this?

gabriel-happle:
I think more correct. The bugfix is the (hopefully) correct implementation of the ISO13790 calculation procedure. There was something forgotten before.

gabriel-happle:
screen shot 2016-08-15 at 10 32 09

The last sentence is the relevant one.
Thank you @martin-mosteiro for the screenshot. Maybe you can have a look at the code and comment here?

martin-mosteiro:
I agree that this is the correct methodology according to ISO 13790 and SIA 2044. So if the results are different, it must be because the previous results were off.

PR #281, opened 2016-08-16
I279 investigate numba aot to spead up calc tm

daren-thomas:
@JIMENOFONSECA check this out: By compiling `calc_tm`, `calc_ts`, `calc_ta`, `calc_top` and `calc_Im_tot` to native code, we can get a 17% reduction in calculation time for the demand script on the zug case study. This technique can be used in other places that have pure math (no objects) calculations, especially if they are run in loops.

The calculated results are slightly different, but this is to be expected with floating point calculations: Neither result is the "correct" solution anyways.


daren-thomas:
currently, you will need to install numba to run this. with a little refactoring we could remove this dependency. but hey, it is as easy as: `conda install numba`!

JIMENOFONSECA:
@daren-thomas could you check the conflicts? i will go through it tomorrow.

JIMENOFONSECA:
cool stuff!

Issue #293, opened 2016-08-19
Full reports not working with "tsd" as dict


gabriel-happle:
The option for full reporting does not work any more for `tsd` variable as `dict`.

gabriel-happle:
@daren-thomas is it OK that I just assigned you?

daren-thomas:
@gabriel-happle yes. It seems this is blocking a few things right now so I will handle it first thing.

daren-thomas:
Fixed with PR #302...

PR #294, opened 2016-08-19
start with demand calculation before the beginning of the heating season (dynamic warm up phase)

gabriel-happle:
- starts demand calculation at the begin of the heating season
- allows to specifiy a dynamic warm up period
- overwrites warm up with results at the end of the cooling season
- uses current outdoor air temperature as initial guess of building
thermal mass and indoor air temperature
- also changes in ventilation calculation together with @JIMENOFONSECA

--> branch works, results look OK at first glance
--> but full testing can not be done until #293 is solved

### DIFF
--- a/cea/demand/thermal_loads.py
+++ b/cea/demand/thermal_loads.py
@@ -15,6 +15,7 @@ import cea.demand.ventilation_model as ventilation_model
 from cea.demand import occupancy_model
 from  cea.demand import sensible_loads, electrical_loads, hotwater_loads, refrigeration_loads, datacenter_loads
 from cea.technologies import controllers
+from cea.utilities import helpers
 
 
 """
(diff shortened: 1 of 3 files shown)


gabriel-happle:
@daren-thomas, @JIMENOFONSECA, @shanshanhsieh, @martin-mosteiro 
Maybe you see that I replaced `np.zeros(8760)` with `np.empty(8760) * np.nan` in the initialization of `tsd`.
I don't know if that is very efficient, but I do believe that we should not use `0` to distinguish between results and initialized values for calculations that could result in `0`. Another way would be to convert all temperatures to Kelvin, where we could assume that a result is never `0`.

JIMENOFONSECA:
@gabriel-happle could you check if the graphs.py, heatmaps etc. work with these temperatures? The nan might affect the range of values for the heatmaps especially.

PR #298, opened 2016-08-19
I055 floor heating model

martin-mosteiro:
Same as reported for i53 pull request, the results look consistent, but full testing cannot be undertaken until #293  is solved


JIMENOFONSECA:
I CAN GET IT AFTER #293

JIMENOFONSECA:
@martin-mosteiro could you check the plots, and results by running graphs.py and demands.py from the source. We have to update from master in order to do that

martin-mosteiro:
I did. The graphs look pretty consistent with previous demand graphs.

daren-thomas:
PR sent to fix #293 - check out #302...

JIMENOFONSECA:
working nicely..

PR #307, opened 2016-08-23
 i287-problem-demand-graphs

JIMENOFONSECA:
@daren-thomas i blelieve it should be working now, would you mind checking for me? merci


daren-thomas:
This only runs for the first two buildings.

PR #314, opened 2016-08-24
I305 running emissions operation error

### LINKED ISSUE #305: Running Emissions operation error
This might not be a huge problem, however, there's a bug.

When running the emissions operation from the CEA Toolbox, if all the following check boxes are ticked (as seen in the photo), then the tool runs smoothly:

emissions_error00

HOWEVER, if I untick any of the check-boxes, it gives me the following error:

emissions_error03

Strangely, the only case where it runs (other than having all the check-boxes marked) is in the following case:

emissions_error02

So, who should I assign this issue to? @JIMENOFONSECA @daren-thomas

daren-thomas:
Fixes #305 (Running Emissions operation error)

### DIFF
--- a/cea/analysis/emissions.py
+++ b/cea/analysis/emissions.py
@@ -12,6 +12,7 @@ from __future__ import division
 
 import pandas as pd
 from geopandas import GeoDataFrame as gpdf
+import os
 
 import cea.inputlocator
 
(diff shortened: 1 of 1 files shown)


JIMENOFONSECA:
@martin-mosteiro does this work in your computer?

martin-mosteiro:
@JIMENOFONSECA Works fine for me.

Issue #312, opened 2016-08-23
Rename non-renewable primary energy demand in output files


martin-mosteiro:
To be consistent with #301 and #306, the embodied, operational and mobility non-renewable primary energy demand in the corresponding csv output files need to be renamed as E_nre_pen, O_nre_pen and M_nre_pen, respectively.

daren-thomas:
https://github.com/architecture-building-systems/CEAforArcGIS/pull/317/files
In this pull request, the file `operation.py` is not changed. I think we didn't check the files after the merge and forgot to check in a file...

I am re-opening the issue.

martin-mosteiro:
Fixed this on pull request #328 . As soon as that goes through I'll close this issue.

daren-thomas:
please make sure the scenario plots also work with your changes.

daren-thomas:
is this issue done? #328 was merged. Can you check all the changes are now in master and the plots work? Then we can close this task!

martin-mosteiro:
Everything works fine. Closing the issue now.

PR #306, opened 2016-08-23
rename emission.py to operation.py

shanshanhsieh:
1. the emission.py script is renamed as operation.py
2. the non-renewable primary energy demand for operation in total_LCA_operation is named as nre_pen_

Issue #301, opened 2016-08-22
Embodied and Operational Energy


amraladdin:
For the Embodied Energy and Operational Energy, I strongly suggest the following:

1. To change in the documentation the title `Primary energy demand and emissions` to `Operational energy demand and emissions`.

2. Differentiate between the names of the output results to the following:

    * For Embodied Energy `E_ghg_ton,E_ghg_kgm2,E_pen_GJ,E_pen_MJm2`

    * For Operational Energy `O_ghg_ton,O_ghg_kgm2,O_pen_GJ,O_pen_MJm2`

    * For Mobility `M_ghg_ton,M_ghg_kgm2,M_pen_GJ,M_pen_MJm2`

martin-mosteiro:
Separate (but related) note: if the "emissions" script calculates not only the operational greenhouse gas emissions but also the operational primary energy demand... Shouldn't it rather be called "operation"?

Even the output files are called "*_LCA_operation.csv"

amraladdin:
Is this question directed to me or Jimeno?

Issue #28, opened 2016-03-02
Primary energy demand output only accounts for fossil based resources


shanshanhsieh:
Should be stated clear at the output

shanshanhsieh:
@JIMENOFONSECA how about calling it "Non-renewable PEN"? And I think it only applies for primary energy occurred in operation, not the embodied part?
also, the name "emission.py" is a bit confusing, since both ghg emission and primary energy demand are calculated in both "embodied.py" and "emission.py" scripts. idea for the name to replace "emission.py" could be "operational.py"

shanshanhsieh:
@JIMENOFONSECA @gabriel-happle any comment?

shanshanhsieh:
Fixed in pull request #317, therefore, this issue is closed.

PR #317, opened 2016-08-24
I301 embodied and operational energy

martin-mosteiro:
Solves issues #28 . #301 . #306 and #312


JIMENOFONSECA:
@martin-mosteiro there are some new conflicts due to the other pull requests, would you please check, i will do the test then.

martin-mosteiro:
@JIMENOFONSECA should be okay now

Issue #35, opened 2016-03-04
Preserve order of columns when writing out to Excel


daren-thomas:
Go through the codebase and change any `DataFrame.to_excel(... cols={...})` to `DataFrame.to_excel(... cols=[...])`

daren-thomas:
What is the correct order of the columns to use? For which files? This goes hand-in-hand with the documentation effort next week...

daren-thomas:
@JIMENOFONSECA: Do we have a preferred ordering for the columns? Do we still care about this issue?

The function in question would be `demand_main.write_totals_csv`.
Also `thermal_loads.results_to_csv`.

JIMENOFONSECA:
@amraladdin i remember coming up together with this issue, is this still important?, i do not see it important anymore.

daren-thomas:
the fix is really easy, though. someone just has to define the ideal column order.

JIMENOFONSECA:
@amraladdin any comment?

amraladdin:
@JIMENOFONSECA @daren-thomas 
What I had in mind was to have a main fixed number of columns that represent the repeated values in all the tables (e.g. Date, Hour and Building Name). These values should be added in the first few columns in the same order for all the tables.
As for the rest of the columns:
- In the case of the demand calculation, the next row of columns I would put the totals, and after that the breakdown of these values.
- In the case of the occupancy or age of buildings for example... The organization after the basic columns does not matter.
I'll attach an excel file as a sample (Note: The highlighted part in yellow is only for explanation, not to be included).
[Sample of Column Organization.xlsx](https://github.com/architecture-building-systems/CEAforArcGIS/files/422220/Sample.of.Column.Organization.xlsx)

daren-thomas:
@amraladdin, to keep this task as simple as possible: Since we output CSV files, we can only have one header row with column names. These are the same names as the variables used in the calculation. I think keeping it that way has some advantages, at least for people interested in the calculation procedure: Then, the variables in the output correspond directly to those in the calculation code.

If you would like to have a specific column order, could you please look through the output files and define the order for me?

An example:

[...]

PR #325, opened 2016-08-29
I35 preserve order of columns when writing out to excel

daren-thomas:
implements #35


cea/globalvar.py
```diff
@@ -501,6 +501,23 @@ def __init__(self):
...
+                                          'Qww_MWhyr', 'Qhprof_MWhyr', 'Qcsf0_kW', 'Qcsf_MWhyr', 'Qcs_MWhyr',
+                                          'Qcref_MWhyr', 'Qcdataf_MWhyr', 'Ealf0_kW', 'Ealf_MWhyr', 'Eauxf_MWhyr',
+                                          'Ecaf_MWhyr', 'Edataf_MWhyr', 'Eprof_MWhyr', 'Tshs0_C', 'Trhs0_C', 'Tscs0_C',
+                                          'Trcs0_C', 'Tsww0_C', 'Vw_m3yr', 'Vww0_m3', 'Vww_m3yr', 'mcphs0_kWC',
+                                          'mcpww0_kWC', 'mcpcs0_kWC']
+        self.demand_building_csv_columns = ['DATE', 'Name', 'occ_pax', 'QHf_kWh', 'QCf_kWh', 'Ef_kWh', 'Qhsf_kWh',
+                                            'Qhs_kWh', 'Qhprof_kWh', 'Qwwf_kWh', 'Qww_kWh', 'Qww_tankloss_kWh',
+                                            'Qcsf_kWh', 'Qcs_kWh', 'Qcref_kWh', 'Qcdataf_kWh', 'Ealf_kWh', 'Eauxf_kWh',
+                                            'Ecaf_kWh', 'Edataf_kWh', 'Epro_kWh', 'Tshs_C', 'Trhs_C', 'Tscs_C',
+                                            'Trcs_C', 'Tsww_C', 'Trww_C', 'Tww_tank_C', 'Tsref_C', 'Trref_C',
```
JIMENOFONSECA: The value Tww_tank_C is not in the master branch and it is not necessary. I might be missing more.

When including or excluding any variable from the output files (which happens often), should both lists be changed accordingly (one in globalvar.py and one in demand.py)? if so, i believe increasing the complexity of such a task should be left for later (when we have a better product). 

The order of columns seems to be a low priority functionality, and it might instead add up to the chances of making mistakes here and there while coding. I recall the phrase of daren: "early optimization is the root of evil". are we facing one of these cases in here? Maybe i am missing something. @daren-thomas would this be the case?

@amraladdin @martin-mosteiro, would this functionality overcome the drawbacks?

daren-thomas:
@JIMENOFONSECA, actually, only the list in `globalvars.py` will need to be changed. The `results_to_csv` function will eventually be refactored to use the `tsd` variable and just pick out the columns that need to be written out.

I personally think this is a safe change to make. At the same time, I don't really care for the order of the columns in the output. I do think that for the user, it increases the usability of the software!

cea/globalvar.py
martin-mosteiro: I believe in terms of teaching it's much more comfortable to present the results in a practical, legible way, and I suppose that is what @amraladdin was going for when requesting this. Is it absolutely necessary? No. Does it add any functionality? No. Would it be nice to have? Well, yes. But I'd also call it low priority.

JIMENOFONSECA:
@daren-thomas @martin-mosteiro cool, so if the tds repalces the entire list. The change seems ok. Would you take a look at the case of the Tww_tank_C? so i can take a look again?

The issue is that the list in globalvar.py and thermal_loads.py do not match

daren-thomas:
@JIMENOFONSECA: sure. I removed `Tww_tank_C` as requested. What list in `thermal_loads.py`?

JIMENOFONSECA:
Sorry i meant the demand script...

thanks for the change..

JIMENOFONSECA:
@daren-thomas the results plot Tww_C as an empty vector now in the output of demand_main.py

daren-thomas:
I'll look into it!

daren-thomas:
@JIMENOFONSECA: I am not quite sure I understand your comment above. What plot?

- I have checked `outputs\data\demand\Total_demand.csv` and could not find that column at all!
- I have checked `outputs\data\demand\Bau G.csv` and could not find that column at all!

What should I be checking?

JIMENOFONSECA:
weird.. I got it in BAU A.csv

On 5 Sep 2016, at 4:07 pm, Daren Thomas <notifications@github.com<mailto:notifications@github.com>> wrote:


@JIMENOFONSECA<https://github.com/JIMENOFONSECA>: I am not quite sure I understand your comment above. What plot?

  *   I have checked outputs\data\demand\Total_demand.csv and could not find that column at all!
  *   I have checked outputs\data\demand\Bau G.csv and could not find that column at all!

What should I be checking?

daren-thomas:
`Tww_C`? You sure? I have `Trww_C` and `Tsww_C`... both have values in `Bau A.csv`. I do not have a `Tww_c`. I just ran the simulation again from master.

JIMENOFONSECA:
sorry i meant Tww_tank_C, Qww_tankloss_kWh (which should not be there btw).
and Epro_kWh.

all of them appear empty...

daren-thomas:
I have `Qww_tankloss_kWh` and `Epro_kWh` but no `Tww_tank_C`?

I will remove `Qww_tankloss_kWh`.

JIMENOFONSECA:
weird.. i just run the simulation from the branch and got that....

daren-thomas:
I have removed `Qww_tankloss_kWh` and added `Eprof_kWh` to the output (but it is all zeros - is that correct?)

JIMENOFONSECA:
correct..


On 5 Sep 2016, at 5:25 pm, Daren Thomas <notifications@github.com<mailto:notifications@github.com>> wrote:


I have removed Qww_tankloss_kWh and added Eprof_kWh to the output (but it is all zeros - is that correct?)

daren-thomas:
`Tww_tank_C` is definitely not part of the output: `globalvars.py` does not list it in the demand output columns... What file are you looking at?

JIMENOFONSECA:
Bau A.csv


On 5 Sep 2016, at 5:28 pm, Daren Thomas <notifications@github.com<mailto:notifications@github.com>> wrote:


Tww_tank_C is definitely not part of the output: globalvars.py does not list it in the demand output columns... What file are you looking at?

daren-thomas:
Please synchronize again, run the demand script and send me the (new) copy of `Bau A.csv` with the `Tww_tank_C` column in it. I can't find it here.

JIMENOFONSECA:
working now and merging

PR #326, opened 2016-08-30
fixes #250

### LINKED ISSUE #250: Use ISO date format for `gv.date_start`
Currently, this is given in the format "d/m/yyyy". Seriously. That isn't even the US way to do it! The US date format would be "m/d/yyyy" - so using this format is just begging for errors to happen.

Turns out, there is an international convention for this: https://en.wikipedia.org/wiki/ISO_8601
We should use "2016-01-01" ("yyyy-mm-dd") instead.

Updates required:
- `gv.date_start`
- `maker.test_schedule_maker` (test to see if pandas can figure out the ISO date format, I'm betting it can...)
- `demand.demand_calculation`

daren-thomas:
checked output with master - no discrepancies as far as I can tell...

### DIFF
--- a/cea/demand/demand_main.py
+++ b/cea/demand/demand_main.py
@@ -84,7 +84,7 @@ def demand_calculation(locator, weather_path, gv):
     """
     t0 = time.clock()
 
-    time_steps = pd.date_range(gv.date_start, periods=8760, freq='H')
+    date = pd.date_range(gv.date_start, periods=8760, freq='H')
 
     # weather model
     weather_data = epwreader.epw_reader(weather_path)[['drybulb_C', 'relhum_percent', 'windspd_ms']]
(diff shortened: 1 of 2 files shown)


JIMENOFONSECA:
@daren-thomas can we keep "date" for now? other algorithms in the optimization part as well as on the output files of the demand call this vector date. It is easier to understand form our side.

daren-thomas:
done. it is just a really bad name, because, well a date is a singular date. time_steps explains more precisely what this is. `dates` would also work (but is still not as good).

JIMENOFONSECA:
Thanks, let's keep it in mind for the next milestone.....

daren-thomas:
Actually. I take that back. I have thought some more about it and now I get it. `date` is like all the other column names in `tsd` - except it is the index. So. Let's leave it :)

Issue #327, opened 2016-08-30
Scenario Plots no longer work!


daren-thomas:
C:\Users\darthoma\Anaconda2\envs\esri104\python.exe C:/Users/darthoma/Documents/GitHub/CEAforArcGIS/cea/plots/scenario_plots.py
Traceback (most recent call last):
  File "C:/Users/darthoma/Documents/GitHub/CEAforArcGIS/cea/plots/scenario_plots.py", line 169, in <module>
    test_plot_scenarios()
  File "C:/Users/darthoma/Documents/GitHub/CEAforArcGIS/cea/plots/scenario_plots.py", line 165, in test_plot_scenarios
    plot_scenarios(scenarios, output_file)
  File "C:/Users/darthoma/Documents/GitHub/CEAforArcGIS/cea/plots/scenario_plots.py", line 59, in plot_scenarios
    plot_lca_embodied(axes[0], locators, scenario_names, column='pen_MJm2', title='Primary Energy', unit='MJ/m2')
  File "C:/Users/darthoma/Documents/GitHub/CEAforArcGIS/cea/plots/scenario_plots.py", line 113, in plot_lca_embodied
    df[scenario] = scenario_data[column] * 1000 / scenario_data['GFA_m2']
  File "C:\Users\darthoma\Anaconda2\envs\esri104\lib\site-packages\pandas\core\frame.py", line 1914, in __getitem__
    return self._getitem_column(key)
  File "C:\Users\darthoma\Anaconda2\envs\esri104\lib\site-packages\pandas\core\frame.py", line 1921, in _getitem_column
    return self._get_item_cache(key)
  File "C:\Users\darthoma\Anaconda2\envs\esri104\lib\site-packages\pandas\core\generic.py", line 1090, in _get_item_cache
    values = self._data.get(item)
  File "C:\Users\darthoma\Anaconda2\envs\esri104\lib\site-packages\pandas\core\internals.py", line 3102, in get
    loc = self.items.get_loc(item)
  File "C:\Users\darthoma\Anaconda2\envs\esri104\lib\site-packages\pandas\core\index.py", line 1692, in get_loc
    return self._engine.get_loc(_values_from_object(key))
  File "pandas\index.pyx", line 137, in pandas.index.IndexEngine.get_loc (pandas\index.c:3979)
  File "pandas\index.pyx", line 157, in pandas.index.IndexEngine.get_loc (pandas\index.c:3843)
[...]

PR #328, opened 2016-08-30
Rename and column order

martin-mosteiro:
Renamed "pen" to "nre_pen" and fixed column order in output CSV files


martin-mosteiro:
Solves the `pen_`/`nre_pen_` naming issue mentioned by Daren in #327

daren-thomas:
test scenario plots with this as well!

daren-thomas:
it seems to work. i'm merging.

PR #334, opened 2016-09-02
I290 pv generation from new radiation script

shanshanhsieh:
A photovoltaic.py script connecting to new radiation calculation results (from Paul) is added to the contribution.
To run the script, please add the example radiation result files to this path: C:\reference-case-zug\baseline\outputs\data\solar-radiation

Link to the example radiation result files: 
https://drive.google.com/folderview?id=0B0xYHrBRUcQoc1ByOHkzQnBXdEk&usp=sharing

### DIFF
--- a/cea/demand/demand_main.py
+++ b/cea/demand/demand_main.py
@@ -106,11 +106,11 @@ def demand_calculation(locator, weather_path, gv):
     else:
         thermal_loads_all_buildings(building_properties, date, gv, locator, num_buildings, schedules_dict,
                                     weather_data)
-    write_totals_csv(building_properties, locator)
+    write_totals_csv(building_properties, locator, gv)
     gv.log('done - time elapsed: %(time_elapsed).2f seconds', time_elapsed=time.clock() - t0)
 
 
-def write_totals_csv(building_properties, locator):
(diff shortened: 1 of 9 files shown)


JIMENOFONSECA:
PV results in null in hours where generation is 0. you might want to see why it is generating null instead of zero.

we might want to add information about the sensors that were selected so later on Paul can map where the PV energy is being taken form and see i the results also are valid.

shanshanhsieh:
Both issues are fixed now.
The ratio_diffhout in the weather data is null when Isol is zero. It is replaced with zeros now, so when Isol is zero, both I_diffuse and I_direct are zero.
Another file, B0000000_sensors.csv, containing metadata of selected sensors is printed. Users could plot the sensors selected with the information in the same file (Zhongming made a grass hopper script to visualize it, it could be included in the plot if necessary).

JIMENOFONSECA:
Great job!

JIMENOFONSECA:
@shanshanhsieh you might want to add this as a new issue for the next milestone: adding the PV to the core together with solar radiation and geometry reader from paul..

shanshanhsieh:
done!

PR #340, opened 2016-09-15
building_info.py script

daren-thomas:
added a command line utility to export input information on buildings as a csv file


martin-mosteiro:
@daren-thomas this script creates a problem when running the Hochschulquartier case on my computer, could you check? The reason for the problem seems to be that the list of buildings is based on the entire area (1482 buildings), but some of the files only include the zone of study (274 buildings). For example, the properties_surfaces.csv file only contains data for the zone of study.

daren-thomas:
you can actually specify the buildings to consider on the command line:

`python building_info.py --scenario C:\reference-case\masterplan --output my_output_file.csv --buildings B1111 B222 B333`

But of course the error message should be nicer...

daren-thomas:
@martin-mosteiro: please check again - it should now work (it works on my computer with the masterplan scenario)

JIMENOFONSECA:
@daren-thomas @martin-mosteiro how do i test this?

daren-thomas:
You can run the tool from the command line with the argument `--help` to get a summary of how to use it. Notice I needed to set the PYTHONPATH environment variable so that the script could access the cea code - but that goes for all our scripts. If you run it without arguments e.g. in PyCharm, it will use the reference case `C:\reference-case-zug\baseline` and all buildings and output to a file called `building_info.csv` in the current directory. But it might be hard to figure out which directory that is when executing from PyCharm - you can set it when you edit the run configurations.

The `--buildings` argument lets you specify only the buildings you are interested in - handy when the scenario contains a lot of buildings and you don't want 242 columns in your resulting csv file...

```
C:\Users\darthoma\Documents\GitHub\CEAforArcGIS>set PYTHONPATH=%PYTHONPATH%;C:\Users\darthoma\Documents\GitHub\CEAforArcGIS

C:\Users\darthoma\Documents\GitHub\CEAforArcGIS>python cea\utilities\building_info.py --help
usage: building_info.py [-h] [-s SCENARIO] [-o OUTPUT]
                        [-b BUILDINGS [BUILDINGS ...]]

optional arguments:
  -h, --help            show this help message and exit
  -s SCENARIO, --scenario SCENARIO
                        Path to the scenario folder
  -o OUTPUT, --output OUTPUT
                        Path to output file
  -b BUILDINGS [BUILDINGS ...], --buildings BUILDINGS [BUILDINGS ...]
                        list of buildings to print
```

Issue #341, opened 2016-09-15
Demand calculation for HQ Masterplan scenario not working anymore


martin-mosteiro:
After changing the occupancies and deleting the `Shape_Leng` and `Shape_Area` fields in the input files to be consistent with the baseline case, the demand calculation for the masterplan scenario is not working anymore...

```
Executing: DemandTool "C:\Users\User\Documents\GitHub\cea-reference-case\reference-case (HQ)\masterplan" Zurich
Start Time: Thu Sep 15 17:16:34 2016
Running script DemandTool...
Using python: C:\Python27\ArcGIS10.4\python.exe
Path to demand script: C:\Users\User\Documents\GitHub\CEAforArcGIS\cea\demand\demand_main.py
Adding path to PYTHONPATH: C:\Users\User\Documents\GitHub\CEAforArcGIS
Running demand calculation for scenario C:\Users\User\Documents\GitHub\cea-reference-case\reference-case (HQ)\masterplan
Running demand calculation with weather file C:\Users\User\Documents\GitHub\CEAforArcGIS\cea\databases\CH\Weather\Zurich.epw
reading input files
done
calculating thermal properties
done
creating windows
done
Using 2 CPU's
Building No. 1 completed out of 242
Building No. 2 completed out of 242
Building No. 3 completed out of 242
Building No. 4 completed out of 242
Building No. 5 completed out of 242
Building No. 6 completed out of 242
Building No. 7 completed out of 242
Building No. 8 completed out of 242
Building No. 9 completed out of 242
Building No. 10 completed out of 242
Building No. 11 completed out of 242
Building No. 12 completed out of 242
Building No. 13 completed out of 242
Building No. 14 completed out of 242
Building No. 15 completed out of 242
Building No. 16 completed out of 242
Traceback (most recent call last):
 File "C:\Users\User\Documents\GitHub\CEAforArcGIS\cea\demand\demand_main.py", line 182, in <module>
   run_as_script(scenario_path=args.scenario, weather_path=args.weather)
 File "C:\Users\User\Documents\GitHub\CEAforArcGIS\cea\demand\demand_main.py", line 173, in run_as_script
   demand_calculation(locator=locator, weather_path=weather_path, gv=gv)
 File "C:\Users\User\Documents\GitHub\CEAforArcGIS\cea\demand\demand_main.py", line 105, in demand_calculation
   weather_data)
[...]

PR #342, opened 2016-09-19
I341 demand calculation hq masterplan

daren-thomas:
This branch attempts to solve the bug #341 by adding a fallback to the newton solve.


cea/demand/ventilation_model.py
```diff
@@ -409,7 +409,7 @@ def allocate_default_leakage_paths(coeff_lea_zone, area_facade_zone, area_roof_z
 
     coeff_lea_path = np.zeros(5)
     height_lea_path = np.zeros(5)
-    orientation_lea_path = np.zeros(5)
+    orientation_lea_path = [0, 1, 0, 1, 2]
```
JIMENOFONSECA: @gabriel-happle does this affect the model? or can we do this modification?

cea/demand/ventilation_model.py
```diff
@@ -565,7 +565,7 @@ def allocate_default_ventilation_openings(coeff_vent_zone, height_zone):
     # initialize
     coeff_vent_path = np.zeros(4)
     height_vent_path = np.zeros(4)
-    orientation_vent_path = np.zeros(4)
+    orientation_vent_path = [0, 1, 0, 1]
```
JIMENOFONSECA: @gabriel-happle does this affect the model? or can we do this modification?

cea/demand/ventilation_model.py
gabriel-happle: No. Right now it does not have any implications. As windows are not used anyways in any calculation. As far as I understand.
Why would you change that?
And what does this have to do with cooling coils?

cea/demand/ventilation_model.py
gabriel-happle: No. Right now it does not have any implications. As windows are not used anyways in any calculation. As far as I understand.
Why would you change that?
And what does this have to do with cooling coils?

cea/demand/ventilation_model.py
JIMENOFONSECA: @daren-thomas ?

cea/demand/ventilation_model.py
JIMENOFONSECA: @daren-thomas ?

cea/demand/ventilation_model.py
daren-thomas: The reason has to do with data types. `np.zeroes` returns an array of floats, but `orientation_lea_path` is actually used to retrieve an index into another array. Indexes need to be integers. I was getting some errors while debugging because they were floats.

BTW: once you have indexes into arrays to retrieve array indexes into other arrays, that can be considered a code smell and yes, here would be a good place to sprinkle some objects...

cea/demand/ventilation_model.py
daren-thomas: Again. `orientation_vent_path`  is used to retrieve an array index and should be integral. We could also just use `map(int, np.zeros(4))`. But, why not just assign the correct indexes with a list literal? As mentioned above, the code structure still needs some work...

cea/technologies/heating_coils.py
```diff
@@ -8,6 +8,7 @@
 from __future__ import division
 import scipy.optimize as sopt
 import scipy
+import numpy as np
```
daren-thomas: this can probably be removed.

cea/technologies/heating_coils.py
```diff
@@ -83,7 +84,15 @@ def fh(x):
...
-        result = sopt.newton(fh, trc0, maxiter=100, tol=0.01) - 273
+        try:
+            result = sopt.newton(fh, trc0, maxiter=100, tol=0.01) - 273
+        except RuntimeError:
+            print('Newton optimization failed, using slower bisect algorithm...')
+            result = sopt.bisect(fh, 0, 350, xtol=0.01, maxiter=500) - 273
+
+
+        if Ta_sup_cs == Ta_re_cs:
+            print 'Ta_sup_cs == Ta_re_cs:', Ta_sup_cs
```
daren-thomas: this can be removed: i only added it to count how often that case happens (it happens a lot!)

Issue #346, opened 2016-09-29
Radiation data missing from reference-case-zug


daren-thomas:
Steps to reproduce:
1. Download reference-case-zug from repo
2. run data helper
3. run Demand script with weather=Zug
4. ???
5. Boom!!

```
Executing: DemandTool C:\reference-case-zug\baseline Zug
Start Time: Thu Sep 29 16:28:43 2016
Running script DemandTool...
Using python: C:\Python27\ArcGIS10.4\python.exe
Path to demand script: C:\Users\darthoma\Documents\GitHub\CEAforArcGIS\cea\demand\demand_main.py
Adding path to PYTHONPATH: C:\Users\darthoma\Documents\GitHub\CEAforArcGIS
Running demand calculation for scenario C:\reference-case-zug\baseline
Running demand calculation with weather file C:\Users\darthoma\Documents\GitHub\CEAforArcGIS\cea\databases\CH\Weather\Zug.epw
reading input files
Traceback (most recent call last):
  File "C:\Users\darthoma\Documents\GitHub\CEAforArcGIS\cea\demand\demand_main.py", line 187, in <module>
    run_as_script(scenario_path=args.scenario, weather_path=args.weather)
  File "C:\Users\darthoma\Documents\GitHub\CEAforArcGIS\cea\demand\demand_main.py", line 176, in run_as_script
    demand_calculation(locator=locator, weather_path=weather_path, gv=gv)
  File "C:\Users\darthoma\Documents\GitHub\CEAforArcGIS\cea\demand\demand_main.py", line 92, in demand_calculation
    building_properties = BuildingProperties(locator, gv)
  File "C:\Users\darthoma\Documents\GitHub\CEAforArcGIS\cea\demand\thermal_loads.py", line 849, in __init__
    solar = pd.read_csv(locator.get_radiation()).set_index('Name')
  File "C:\Python27\ArcGIS10.4\lib\site-packages\pandas\io\parsers.py", line 474, in parser_f
    return _read(filepath_or_buffer, kwds)
  File "C:\Python27\ArcGIS10.4\lib\site-packages\pandas\io\parsers.py", line 250, in _read
    parser = TextFileReader(filepath_or_buffer, **kwds)
  File "C:\Python27\ArcGIS10.4\lib\site-packages\pandas\io\parsers.py", line 566, in __init__
    self._make_engine(self.engine)
[...]

PR #348, opened 2016-09-30
create the solar radiation folder if it does not yet exist

daren-thomas:
fixes the issue mentioned in #346 about the radiation script not running for the zug case study.


daren-thomas:
@JIMENOFONSECA this is one way to fix the bug. I would prefer this solution (or a similar one) to using `__init__.py` files in the reference cases to keep empty folder structures on GitHub. Specifically, `__init__.py` makes a folder a python package and I'd prefer to keep those to the source repository...

PR #357, opened 2016-10-07
Jenkins - adds continuous integration

daren-thomas:
Run tests on all reference cases with the jenkins server (currently installed on darthoma-pc) to make sure we keep master up-and-running.


daren-thomas:
@JIMENOFONSECA, I think this branch can be merged - then I can reconfigure jenkins to track the master branch (as opposed to this branch)

PR #364, opened 2016-10-13
I297 tabs temperature

martin-mosteiro:
Finally bothered to fix the temperature calculation...


daren-thomas:
who is the best person to review these changes? I have had a quick look, but I think it would be better if we had some domain knowledge here...

@gabriel-happle? @JIMENOFONSECA?

daren-thomas:
Please don't merge yet: I am making some (cosmetic) changes.

daren-thomas:
@martin-mosteiro : could you provide a calculation example (unit test) for the tabs module? we can work on this together if you like... are any buildings in the three reference cases using the tabs module?

martin-mosteiro:
I tested it by creating new scenarios where every building was heated with TABS. Nonetheless, the following buildings from the Zurich baseline scenario use TABS:

- B302019753
- B302019754
- B302019755
- B302020926
- B302030670
- B302030671
- B302030556
**- B302021377
- B302024465**
- B302030047

Out of these, only the ones in bold are within the zone of study.

daren-thomas:
This branch can now be merged. Martin and I will add some unit tests as an example of how we should be testing models in the future, but the branch can be merged before that.

Issue #389, opened 2016-10-31
Numerous issues with the embodied energy script


martin-mosteiro:
I found a number of issues in the embodied energy and emissions calculation.

1. A minor issue: is there any reason to calculate _delta_year_ as 
        `envelope_df['delta_year'] = (envelope_df['envelope'] - year_to_calculate) * -1`
instead of the much simpler and easier to read:
        `envelope_df['delta_year'] = year_to_calculate - envelope_df['envelope']`?

2. A likely couple of typos: the contributions due to HVAC renovation include two mistakes. 
        `HVAC_df['delta_year'] = (`**`HVAC_df['basement']`**` - year_to_calculate) * -1` (uses the year of renovation of the basement instead of the year of renovation of the HVAC **`HVAC_df['HVAC']`**)
        `HVAC_df['confirm'] = HVAC_df.apply(lambda x: calc_if_existing(x['delta_year'], `**`gv.sl_materials`**`),axis=1)` (uses the service life of materials instead of the service life of services **`gv.sl_services`**)

On to the important stuff.

3. @JIMENOFONSECA what is `PFloor` exactly? It is only used for calculating the embodied energy/emissions of **windows (both building and renovation)**, **internal supporting walls** (**only building**, not renovation) and **external walls** (**only renovation**, not building). The fact that for the latter two this factor is only used for either building or renovation seems inconsistent.

Furthermore, for each of the cases described above, the embodied energy/emissions of the corresponding building component is multiplied by `(PFloor - 1)`. Since for every building in the area of study `PFloor` = 1, **the embodied energy and emissions of these components are not included in the calculation**. @JIMENOFONSECA is this meant to be that way?

4. Finally, for the construction as well as for the renovations, `built_df['confirm']`/`envelope_df['confirm']`/etc. are defined as 1 if the calculation year is within 60 years of the year of construction/renovation and 0 otherwise. However, this factor is only used for renovations. This means that:
- **for construction, `embodied energy = (embodied energy / 60)` regardless of the calculation year**
- **for renovation, `embodied energy = (embodied energy / 60)` if within 60 years of renovation
                            `embodied energy = 0` if the calculation year is more than 60 years after renovation.**
@JIMENOFONSECA Is there any reason for this difference? Is it a mistake, as it would seem?

martin-mosteiro:
@amraladdin these are the issues I described in my email to you.

JIMENOFONSECA:
[...]

PR #391, opened 2016-11-04
I389 embodied energy

martin-mosteiro:
Fixed numerous problems with the embodied energy calculation as described in #389


JIMENOFONSECA:
@martin-mosteiro im checking the pull request, what are the other scripts added about?

martin-mosteiro:
@JIMENOFONSECA Hm. Seems like when there was a merge conflict I saved backups by mistake. These additional files should be deleted. Only `embodied.py` matters.

PR #396, opened 2016-11-10
new open case study

JIMENOFONSECA:
As a requirement by prospect users


JIMENOFONSECA:
@daren-thomas would you check if it works, thank you!

PR #402, opened 2016-11-14
Radiation script GUI with longitude and latitude fields

daren-thomas:
this is a small fix for the gui of the radiation tool to allow overriding the latitude and longitude read from the shape files.


daren-thomas:
@JIMENOFONSECA can you give this one a quick look?

PR #405, opened 2016-11-16
Document pv script

shanshanhsieh:
Documentation for photovoltaic.py script.
It was created to communicate the assumptions with Paul for his thesis. The rest of the documentation will be finished when we are fully integrating Paul's radiation script into the master.


daren-thomas:
I'm checking this out right now and will fix the conflicts.

cea/inputlocator.py
```diff
@@ -175,6 +175,42 @@ def get_surface_properties(self):
...
+        """scenario/outputs/data/solar-radiation/radiation.csv"""
+        return os.path.join(self.scenario_path, 'outputs', 'data', 'solar-radiation')
+
+
+    ## POTENTIALS
+    def get_solar_potential_folder(self):
+        """scenario/outputs/data/potentials/solar"""
+        return os.path.join(self.scenario_path, 'outputs', 'data', 'potentials', 'solar')
+
+    def PV_results(self, building_name):
```
daren-thomas: you changed the naming scheme of the functions here - this should only be done with good reason... is it OK if I change them to `get_pv_results` etc?

cea/optimization/ntwOpt/NtwMain.py
```diff
@@ -9,7 +9,7 @@
 
 class ntwFeatures(object):
```
daren-thomas: class names are CamelCase (not camelCase). use `NtwFeatures` instead. In fact, you can use as many characters as you like, so `NetworkFeatures` is even better!

Also, module names are all lower-case and `_` separated words (but try to find single-word options)

`ntwOpt` is also not a good name for a package. Since the package name starts with `cea.optimization`, I think we can just rename it to `network`, right? `cea.optimization.network`.

Unless `ntw` doesn't stand for "network", but then, well, then we have other problems...

I guess this code is inherited from the master students thuy-an and tim, right? We'll just have to clean up the mess now...

cea/utilities/solar_equations.py
```diff
@@ -104,21 +104,23 @@ def pyephem(time, latitude, longitude, altitude=0, pressure=101325,
 def calc_sun_properties(latitude, longitude, weather_data, gv):
 
     date = pd.date_range(gv.date_start, periods=8760, freq='H')
+    hour_date = date.hour
```
daren-thomas: couldn't we just pass the date vector to `get_hour_angle`? then we don't have to worry about `min_date` potentially meaning `min(date)` etc...

cea/utilities/solar_equations.py
```diff
@@ -132,13 +134,13 @@ def calc_sunrise(sunrise, Yearsimul, longitude, latitude):
...
 
-def declination_degree(when, TY):
+def declination_degree(day_date, TY):
     """The declination of the sun is the angle between Earth's equatorial plane and a line
     between the Earth and the sun. It varies between 23.45 degrees and -23.45 degrees,
     hitting zero on the equinoxes and peaking on the solstices.
     Parameters
     ----------
-    when : datetime.datetime
+    day_date : datetime.datetime
```
daren-thomas: maybe we can check the parameter type (i think it might just be an int, the `day`?)

cea/utilities/solar_equations.py
```diff
@@ -150,21 +152,46 @@ def declination_degree(when, TY):
...
     .. [1] http://pysolar.org/
     """
-    return 23.45 * math.sin((2 * math.pi / (TY)) * ((when.utctimetuple().tm_yday) - 81))
 
+    return 23.45 * np.vectorize(math.sin)((2 * math.pi / (TY)) * (day_date - 81))
 
-def get_hour_angle(when, longitude_deg):
-    solar_time = get_solar_time(longitude_deg, when)
+
+def get_hour_angle(longitude_deg, min_date, hour_date, day_date):
```
daren-thomas: refactor parameters to just a date?

cea/utilities/solar_equations.py
```diff
@@ -150,21 +152,46 @@ def declination_degree(when, TY):
...
+    Parameters
+    ----------
+    day_date
+
+    Returns
+    -------
+
+    """
+    B = (day_date-1)*360/365
+    E = 229.2*(0.000075 + 0.001868*math.cos(B) - 0.032077*math.sin(B) - 0.014615*math.cos(2*B) - 0.04089*math.sin(2*B))
```
daren-thomas: we should collectively get in the habit of formatting expressions properly: there is a single space left and right of the operator. pycharm can format this for you and complains if you don't (squiggly lines are a dead give-away)

cea/optimization/ntwOpt/NtwMain.py
shanshanhsieh: This change did not come from me... Maybe it showed up because this is actually some updates in the master branch? The current pull request is for the contribution branch.
But yes, I started working on the network part recently. So I will keep this in mind when modifying the codes in my network branch : )

PR #419, opened 2016-11-22
I415 finish documentation for sensitivity analysis

daren-thomas:
@JIMENOFONSECA, check out this documentation - is it satisfactory? Then we can merge. I have tested this on Euler with METHOD=morris and N=3 and it seems to work (including the bugfix of the naming of the analysis_METHOD_N.xls file)


JIMENOFONSECA:
yes it is :-)

PR #422, opened 2016-11-28
fixes #247 by updating unit tests to reflect changes that had already been made

### LINKED ISSUE #247: Rename `thermal_loads_new_ventilation`
The name only has meaning in the context of "there was an old ventilation model".

Also, package and module names should not include underscores... (but I also don't really see right now how to do it without...)

@gabriel-happle: is there a name we could use that refers to the ventilation model actually used? Is it a model you came up with or an implementation of a standard? Then we could call it `iso1234` or something like that?

@JIMENOFONSECA: should this be moved to core? It is now the default model, right?

daren-thomas:


JIMENOFONSECA:
@daren-thomas there is a conflict there

daren-thomas:
@JIMENOFONSECA conflict solved, waiting for the Jenkins.

PR #435, opened 2016-12-13
222 check for radiation input in demand script

daren-thomas:
How to test: The Jenkins will run the unit test to check the part of the demand script raises the ValueError - but you could run a demand simulation manually (with command line interface) for a scenario without radiation files to check this...

In ArcGIS, run the demand tool for a scenario without radiation files. It should not let you start the simulation, but instead provide an error message ("No radiation data found for scenario. Run radiation script first.")

### DIFF
--- a/cea/GUI/demand_tool.py
+++ b/cea/GUI/demand_tool.py
@@ -1,3 +1,5 @@
+import os
+
 import arcpy
 
 import cea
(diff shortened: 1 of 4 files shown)


JIMENOFONSECA:
It works fine and can be merged without generating conflict with the other pull requests

Issue #441, opened 2016-12-20
Create a call graph of the various scripts


daren-thomas:
Create a graphical representation of how the code flows through the functions in the various scripts, starting with the demand script.

- either with the data from the `functionlogger`
- or with the tool `pycallgraph`
- or some other tracing system

Add the output to the documentation.

daren-thomas:
see #448 for an implementation of this with the following output:

[demand-module-overview.gv.pdf](https://github.com/architecture-building-systems/CEAforArcGIS/files/682398/demand-module-overview.gv.pdf)

[demand-function-graph.gv.pdf](https://github.com/architecture-building-systems/CEAforArcGIS/files/682399/demand-function-graph.gv.pdf)

PR #448, opened 2017-01-03
441 create call graph

daren-thomas:
Creates the call graphs mentioned in #441 at the function and module level.

(The function level is waaay too detailed to show in the documentation but interesting in its own right, the module level graph hints at some refactorings that could be made...)


JIMENOFONSECA:
how do it test it?:-) @daren

JIMENOFONSECA:
One question, does it allow to check also variables and running time? I came across the module pycallgraph which could be useful for this, is it different?

daren-thomas:
@JIMENOFONSECA maybe wait a bit first. I will add more documentation but I'm still working on some interesting aspects of this technology.

You can use the batch files in the `bin` folder to run it.

daren-thomas:
pycallgraph is cool. I used the same technique but don't check running time or variables (waaay too much information to store). Checking running time if you're doing tracing doesn't really work anyway, since the tracing adds a big overhead. This needs to be done with a profiler instead.

JIMENOFONSECA:
@daren-thomas thanks, cool stuff.

daren-thomas:
@JIMENOFONSECA How to test this:

- check out the `bin` folder: There are two `.bat` files that each produce a pdf. 
  - The documentation in those files point out that you need to install GraphViz in order to produce the files - this is an easy-to-install program
- there is an ipython notebook in the `bin` folder that shows a more programmatic use of the `create_function_graph` module to filter out calls to `InputLocator` for documenting input- and output-files to our scripts.
  - output was manually processed to produce the file `docs/module_input_and_output_files.md` that can later be incorporated into the documentation somehow.

Issue #455, opened 2017-01-16
Document every function incl. source in cea/demand folder


gabriel-happle:
Document every function incl. source in cea/demand folder
- short documentation incl. sources
- move constant values out of individual scripts

gabriel-happle:
Documentation work is continuous process. Don't know if issue is needed for that.

Issue #427, opened 2016-12-01
SIA 2044 thermal loads to CEA incl. TABS and Corrections to standard


gabriel-happle:
SIA 2044 methods to CEA incl. TABS and corrections to standard from Korrigenda C1 and Korrigenda C2

gabriel-happle:
@JIMENOFONSECA, @martin-mosteiro, @daren-thomas 
According to my understanding we are using the TABS calculation in a wrong way.
Please check 3.1.8 in SIA 2044. Floor heating (currently implemented in CEA) should be calculated with the "normal" RC model but separated convection and radiation component of the heating power.
TABS are calculated according to 2.2.7.3 ff with an extended RC model.
I want to fix that, so we have a correct TABS calculation and we can include more variety of heating/cooling systems (floor heating). But the merge moves further and further away. Chances are low I can finish by the end of the year. Maybe we can move everything to next year and couple with OO? What do you think?

martin-mosteiro:
As I mentioned during our call, the current version of the calculation does indeed follow 3.1.8 already. However, this uses a hard-coded value of 0.5 for the TABS case. This could be made much simpler/clearer by adding the corresponding variable _f_HC_ mentioned in the standard and set its value depending on whether it's a radiator or a TABS system.

The problem is rather in the temperature controls, which use the wrong temperature (t_s is misused as the surface temperature of the TABS system in the control). The problem is that the way the code is currently structured, the surface temperature calculation is only carried out _after_ the demand calculation, meaning the control would only be established then - that's the reason we discussed 2.2.7.3 on Skype. The way I understood our conversation at that time, this was the main issue that needed to be solved, and it would go along with your substantial restructuring of the code.

Regarding meeting the milestone, as I mentioned back then, I could easily do the fixes we discussed by the milestone if you need help. However, I don't know if it's contradictory to the way you are moving, so it's kind of up to you, I guess.

Also, what is OO?

martin-mosteiro:
Okay, after talking to @gabriel-happle and rereading the above message I get the issue. The current CEA doesn't make a distinction between TABS and floor heating and uses the floor heating calculation method from SIA 2044 3.1.8 with the temperature controls for TABS from the EMPA book. This should be fixed and floor heating and TABS be defined individually.

Issue #392, opened 2016-11-07
move required ventilation rate script


gabriel-happle:
The required ventilation rate according to occupancy is now calculated inside a temperature controller function `cea.controllers.calc_simple_temp_control()`.

The function will be moved to an appropriate location

Issue #379, opened 2016-10-21
Update full report function


gabriel-happle:
A lot of variable can currently not be reported any more in the `utilities.reporting` function, as names of variables have changed.
Names of variables to report are in `globalvar`

`cannot report following variables: q_int, Eal_nove, Edata, vww, vw, Qcdata, Qcrefri, Hve, Htr_is, Htr_ms, Htr_w, Htr_em, Htr_1, Htr_2, Htr_3, QHC_sen, t5, tHset_corr, tCset_corr, Waterconsumption`

`cannot report following variables: Name, Af, Aef, sys_e_heating, sys_e_cooling, Am, Atot, Awall_all, Cm, Ll, Lw, Retrofit, Sh_typ, Year, footprint, nf_ag, nfp, Lcww_dis, Lsww_dis, Lv, Lvww_dis, Tcs_re_0, Tcs_sup_0, Ths_re_0, Ths_sup_0, Tww_re_0, Tww_sup_0, Y, fforma`

daren-thomas:
good point! A lot of these might have been moved to the `tsd` variable...

PR #460, opened 2017-01-18
Updated Demand Calculation

gabriel-happle:
I post this pull request for @JIMENOFONSECA and @daren-thomas to have a look at the new demand calculation procedure. Please check it out while I still continue to do some testing, improvements, clean up and documentation.

This pull request #460 addresses #220 #392 #427 (partially, as there are currently no TABS implemented) #84 and #379   also #455


gabriel-happle:
image

gabriel-happle:
image

gabriel-happle:
@daren-thomas How can I see why checks failed?
I accidentally committed a change to the path of the reference cases folder. maybe that is the reason for the failing checks?

daren-thomas:
@gabriel-happle the easiest way to see why checks fail is to run the `dodo.py` script in the tests folder. I think the reason the tests are failing is because you now are computing different results - you'll need to update the unit tests to reflect the new results. There is a helper script to do that for you. If you can't find out how, we can have a look together :)

gabriel-happle:
@daren-thomas I installed requests and doit. But now I get this authentication error:

`"C:\Program Files (x86)\Anaconda2\envs\esri104\python.exe" C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/tests/dodo.py
.  download_reference_cases
########################################
TaskError - taskid:download_reference_cases
PythonAction Error
  File "C:\Program Files (x86)\Anaconda2\envs\esri104\lib\site-packages\doit\action.py", line 383, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/tests/dodo.py", line 76, in download_reference_cases
    r = requests.get(REPOSITORY_URL % REPOSITORY_NAME, auth=get_github_auth())
  File "C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/tests/dodo.py", line 66, in get_github_auth
    with open(os.path.expanduser('~/github.auth')) as f:
IOError: [Errno 2] No such file or directory: 'C:\\Users\\Gabriel/github.auth'




Process finished with exit code 2`

daren-thomas:
@gabriel-happle I'm terribly sorry: finishing the documentation for the dodo script has been on my list of tasks for waaay too long.

So. You will need to get a github authentication token. On GitHub, go to your profile / settings and in the Developer Settings side-bar go to "Personal Access Tokens". You can create a new access token (read should be enough, I think?) and then save that token to a file called "github.auth" in your user folder on your local drive. The dodo script is set up to use your personal token like this, so that we don't save credentials into the repository.

gabriel-happle:
@daren-thomas 
not sure if I'm doing this right. My token is just one line (looks like hex code). The documentation of the failing function says there should be one line with user name and one line with password.


"C:\Program Files (x86)\Anaconda2\envs\esri104\python.exe" C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/tests/dodo.py
.  download_reference_cases
########################################
TaskError - taskid:download_reference_cases
PythonAction Error
  File "C:\Program Files (x86)\Anaconda2\envs\esri104\lib\site-packages\doit\action.py", line 383, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/tests/dodo.py", line 76, in download_reference_cases
    r = requests.get(REPOSITORY_URL % REPOSITORY_NAME, auth=get_github_auth())
  File "C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/tests/dodo.py", line 67, in get_github_auth
    user, password = map(str.strip, f.readlines())
ValueError: need more than 1 value to unpack

JIMENOFONSECA:
@daren-thomas is it still the case that we need to use the 'tds' dict instead of simple lists? with 8760 points and multiple variables the demand script with 'tds' dict has proven to slow down the code way too much. Besides we are starting to loose track of the input and output variables that enter to every function as they are masked in 'tsd'.

I am thinking in the sensitivity analysis and calibration when we need >50.000 simulations and well every second count. can we come back to lists? I imagine we can easily paralellalize the code with lists too. we can leave the dicts just for reporting.

cea/demand/airconditioning_model.py
```diff
@@ -9,11 +9,12 @@
     [1] Kämpf, Jérôme Henri
```
gabriel-happle: @JIMENOFONSECA 
Full reporting a function used for testing the code and write intermediate results to .xls files. This can be activated/deactivated by setting `testing = False` in the globalvariables (like it used to be before).
Yes. The code is slower. Maybe because the speedy numba functions are not used anymore. I hope we can reestablish the speed. I'm working on that.

cea/demand/airconditioning_model.py
```diff
@@ -9,11 +9,12 @@
...
         On the modelling and optimisation of urban energy fluxes
         http://dx.doi.org/10.5075/epfl-thesis-4548
+
```
gabriel-happle: of course. That's the goal. I think that is easier possible now. As there are smaller functions in general.

cea/demand/thermal_loads.py
```diff
@@ -101,7 +100,7 @@ def calc_thermal_loads(building_name, bpr, weather_data, usage_schedules, date,
     schedules = usage_schedules['schedules']
 
     # get n50 value
-    n50 = bpr.architecture['n50']
+    # n50 = bpr.architecture['n50']
```
gabriel-happle: This is part of bpr. It can be called in every function that has bpr as an input.

cea/demand/thermal_loads.py
```diff
@@ -229,477 +238,107 @@ def calc_thermal_loads(building_name, bpr, weather_data, usage_schedules, date,
...
+    tsd['QHf'] = tsd['Qhsf'] + tsd['Qwwf'] + tsd['Qhprof']
+    tsd['QCf'] = tsd['Qcsf'] + tsd['Qcdataf'] + tsd['Qcref']
+    tsd['Ef'] = tsd['Ealf'] + tsd['Edataf'] + tsd['Eprof'] + tsd['Ecaf'] + tsd['Eauxf'] + tsd['Eref']
+    tsd['QEf'] = tsd['QHf'] + tsd['QCf'] + tsd['Ef']
 
     # write results to csv
     gv.demand_writer.results_to_csv(tsd, bpr, locator, date, building_name)
     # write report
-    gv.report('calc-thermal-loads', locals(), locator.get_demand_results_folder(), building_name)
+    gv.report(tsd, locator.get_demand_results_folder(), building_name)
```
gabriel-happle: see above

cea/demand/ventilation_air_flows_simple.py
```diff
@@ -0,0 +1,196 @@
...
+
+    # get properties
+    n50 = bpr.architecture['n50']
+    area_f = bpr.rc_model['Af']
+
+    # estimation of infiltration air volume flow rate according to Eq. (3) in DIN 1946-6
+    n_inf = 0.5 * n50 * (gv.delta_p_dim/50) ** (2/3)  # [air changes per hour] m3/h.m2
+    infiltration = gv.hf * area_f * n_inf * 0.000277778  # m3/s
+
+    tsd['m_ve_inf_simple'] = infiltration * physics.calc_rho_air(tsd['T_ext'][:])  # (kg/s)
```
gabriel-happle: @daren-thomas @JIMENOFONSECA the question is whether this is faster. I can try. This might also be one of the reasons why the code is slower. I execute some functions many times per time step instead of looking up previously calculated results. What do you think Daren?

daren-thomas:
@JIMENOFONSECA, no. It hasn't proven to slow down the code way too much. The `tds` variable just holds the lists so that we don't have functions that have 7+ parameters. Long lists of parameters are really problematic, especially as the ordering of the parameters can easily get confused. We never stopped using the lists. Having them in the `tds` variable makes it easy for us to ensure that these lists have the same name everywhere. Also, `tds` makes the naming of the variables canonical across the whole software. This was one of the problems I was trying to fix in the first place.

For the end functions at the leaves of the call tree, where only 3-5 parameters are required, you can pass in the lists as taken from the `tds`. But don't fool yourself into thinking this is going to speed it up much. We need to do profiling to find out exactly where the time is being spent. This is the only way to reason about performance. I'm on it. Also, some of the performance loss is due to the fact that the master version uses a compiler to produce native code (machine code) for some core inner functions. I can provide a similar solution after profiling.

Lets get this to work first. Then fast. Then we make changes.

As you pointed out: There are waaay too many things changed in this branch. We should not do it like this anymore! Solve one problem at a time (but I think @gabriel-happle now sees the problems of the big-branch approach)

daren-thomas:
First profiling done: 68% of the time is used in reporting...

daren-thomas:
yes. at the level of `calc_h_ea`, the `tsd` variable is getting in the way (makes it hard to compile to native code)

gabriel-happle:
@daren-thomas 
- if reporting takes long it should not be an issue. just turn it off.
- I will change the code in the `rc_model_SIA` script then. It should not use tsd and bpr, if I get it right. (?)

daren-thomas:
@gabriel-happle please wait. instead, focus on making the code correct. i will do the performance optimization.

cea/demand/airconditioning_model.py
```diff
@@ -27,248 +28,304 @@
...
+    # room supply moisture content:
+    # algorithm for cooling case
+    w3 = calc_w3_cooling_case(t5, w2, t3, w5_prime)
+
+    # State of Supply
+    ts = t3  # minus expected delta T rise in the ducts TODO: check and document value of temp decrease
+    ws = w3
+
+    # actual room condition
+    w5 = (wint + w3 * m_ve_hvac_req) / m_ve_hvac_req
```
gabriel-happle: @JIMENOFONSECA Oops. That should not happen. I haven't encountered that. Do you remember which scenario? I usually test with the Open scenario and the Zug scenario.

JIMENOFONSECA:
Hi, it happens with the open scenario,

Jimeno
On 25 Jan 2017, at 10:50 am, Gabriel Happle <notifications@github.com<mailto:notifications@github.com>> wrote:


@gabriel-happle commented on this pull request.

In cea/demand/airconditioning_model.py<https://github.com/architecture-building-systems/CEAforArcGIS/pull/460>:

> +    # moisture balance accounting for internal moisture load, see also Eq. (4.32) in [1]
+    w5_prime = (wint + w3v * m_ve_hvac_req) / m_ve_hvac_req
+
+    # supply air condition
+    t3 = gv.temp_sup_cool_hvac
+
+    # room supply moisture content:
+    # algorithm for cooling case
+    w3 = calc_w3_cooling_case(t5, w2, t3, w5_prime)
+
+    # State of Supply
+    ts = t3  # minus expected delta T rise in the ducts TODO: check and document value of temp decrease
+    ws = w3
+
+    # actual room condition
+    w5 = (wint + w3 * m_ve_hvac_req) / m_ve_hvac_req


@JIMENOFONSECA<https://github.com/JIMENOFONSECA> Oops. That should not happen. I haven't encountered that. Do you remember which scenario? I usually test with the Open scenario and the Zug scenario.

gabriel-happle:
@JIMENOFONSECA I corrected this. This happened in the case of cooling/heating demand but zero mechanical ventilation, because air-conditioning is now decoupled from ventilation. It will be fixed with the next commit.

daren-thomas:
@gabriel-happle, the unit tests are currently not passing. I can generate new data for the thermal calculation stuff, but what about these tests:

- test_calc_t_em_ls_T0
- test_calc_t_em_ls_raises_ValueError

It seems cea.demand.sensible_loads has changed and the setpoint_correction_for_space_emission_systems function has been removed. Any ideas how to fix this?

gabriel-happle:
@daren-thomas yes. sorry about that. I used the calculation procedure of that respective standard (where Shanshan took the values from) and moved/modified those functions. I hope the test still works if you point to the new location on `space_emission_systems.py`

PR #487, opened 2017-02-10
fixes #486  by adding precision to csv demand output

### LINKED ISSUE #486: Optimization algorithm not working at low load, small heat transfer fluid temperature difference
Optimization Algorithm not working with low load, small heat transfer fluid temperature difference.
Effect was observed in Building 5 in Zug case study by @BhargavaKrishnaSreepathi and me at heating load of 2 W and supply and return fluid temperatures of 20.9°C, resulting in 0°C temperature difference.
Reason 1 is the low precision of the demand .csv file rounding to 3 digits after the comma! Easy fix.
Reason 2 might be the faulty operation strategy of the heating system resulting in very small load calculation.

The issue will be fixed first by increasing the precision of the .csv file output.

gabriel-happle:
- also removes output print statement in space_emission_systems (for
emission losses larger than actual load)


gabriel-happle:
Please merge branch asap. In order for optimization to run.

JIMENOFONSECA:
@gabriel-happle we might need to check why it is failing the jenkings..

gabriel-happle:
@JIMENOFONSECA I think because the results are changed (to include higher precision).

JIMENOFONSECA:
@gabriel-happle @BhargavaKrishnaSreepathi does this solve the optimization problem? I see this affects how the results are displayed including temperature and mass flow rates. It also increases the storage needs of every output file.  Is there any other cause beyond the precision?

BhargavaKrishnaSreepathi:
@JIMENOFONSECA  it does solve the optimization problem. But what we employed here is a quick fix. @gabriel-happle has created a new issue to deal with low heat loads, such as just 0.002 kWh, which is quite small and begs the question of why it is being input in the first place. I do understand that it will consume more memory, but again, it might be a quick fix until we figure out the alternative

gabriel-happle:
@JIMENOFONSECA and @BhargavaKrishnaSreepathi The other option is to adjust the optimization calculations to avoid possible divisions by zero. I usually do this in my functions. Just check if the denominator (in this case a temperature difference) is smaller than some tolerance and then execute another calculation.
The real fix would be by revising the turned off/on status of systems. But if we want to allow for very small loads, we have to increase the precision of the temperature values.

JIMENOFONSECA:
i suggest to go for a more durable fix and see if loads that low should be considered. I said because it increases the storage needs and makes necessary to change the way we visualize the results of numbers in arcgis + again messing with the jenkins dude

On 10 Feb 2017, at 15:26, Gabriel Happle <notifications@github.com<mailto:notifications@github.com>> wrote:


@JIMENOFONSECA<https://github.com/JIMENOFONSECA> and @BhargavaKrishnaSreepathi<https://github.com/BhargavaKrishnaSreepathi> The other option is to adjust the optimization calculations to avoid possible divisions by zero. I usually do this in my functions. Just check if the denominator (in this case a temperature difference) is smaller than some tolerance and then execute another calculation.
The real fix would be by revising the turned off/on status of systems. But if we want to allow for very small loads, we have to increase the precision of the temperature values.

BhargavaKrishnaSreepathi:
sure. may be until we fix that, if some one is using the optimization part, they can increase the decimal points in their branch.

PR #493, opened 2017-02-10
Correcting citation

martin-mosteiro:
One of the references was inconsistently referenced in different
functions.


JIMENOFONSECA:
Accepting as it does not need code verification.

PR #500, opened 2017-02-23
I9 new radiation engine daysim by paul

chenkianwee:
refactored codes for soalr analysis using daysim and pyliburo


JIMENOFONSECA:
@chenkianwee i get the next error when creating the Citygml file. 
image

Is there a release version of pylibburo i need to install? is it conda or pip installable?

JIMENOFONSECA:
@chenkianwee on another hand is it possible that we consider a context of buildings in the solar radiation calculation? Paul did it before it was edited to citygml. the files is district.shp located in the ninecubes example are the current builidings + the context. I uploaded theninecubes  example to the repository in this branch.

chenkianwee:
@JIMENOFONSECA yup I have made some changes to the pyliburo library, I will do a new release so that you can install that on your computer and the script should work. 

For the second question, ya its considered in the solar radiation. I checked the district.shp before, its the same geometry as all the other shp files

JIMENOFONSECA:
@chenkianwee actually i just noticed that i forgot to include in the district.shp the context of buildings. Check it out in the branch. I just uploaded. In total there are 21 buildings including the 9 that are of interest...

chenkianwee:
untitled

@JIMENOFONSECA  there are only 9 buildings when I open the district.shp file

JIMENOFONSECA:
weird, did you downloaded ninecubes.zip again from this branch? maybe try to sync again?

PR #502, opened 2017-02-23
I486 small load bug

### LINKED ISSUE #486: Optimization algorithm not working at low load, small heat transfer fluid temperature difference

gabriel-happle:
Fixes #486 by introducing some tolerance in temperature difference in control of heating and cooling systems


gabriel-happle:
@daren-thomas please have a look. I will include some documentation meanwhile

gabriel-happle:
@daren-thomas can you please take a look? This should help the optimization guys

PR #534, opened 2017-03-08
Singapore weather file 2016

BhargavaKrishnaSreepathi:
Singapore weather file 2016 is created. Even though it is a leap year, to be consistent with other years and to have 8760 hours, 29/02/2016 data is not included in the weather file. The file seems to be integrated well with the other components of the Master. 'demand_main.py' file uses the weather file through 'epw_reader', and also the 'optimizaiton_main.py' runs with the updated demand files


gabriel-happle:
@daren-thomas and @JIMENOFONSECA 
Please merge to test run radiation calculation for EcoCampus

Issue #529, opened 2017-03-07
Hide process to get Bwidth and Blenght in Zone and district file


JIMENOFONSECA:
These values are confusing and make difficult the user experience. they can be calculated in the background.

PR #547, opened 2017-03-21
I529 hideblenght bwidth

JIMENOFONSECA:
solving issue #529 @gabriel-happle would you mind testing this? after we could erase both fields from the input files in all case studies


JIMENOFONSECA:
This pull request includes changes to the ninecubes example. and renames it

PR #549, opened 2017-03-21
I531 new database structure

JIMENOFONSECA:
PR 531 introduces a new structure of databases: 

1. basically the thermal_properties.shp is erased. most of the values contained in this shapefile now will exist in the architecture.shp for the sake of simplicity. @martin-mosteiro please consider this for the HQ case study. If you upload it to the private repository we will make this change for you inmediately.

2. in connection to this, the database envelope_systems.xls will now handle the thermal properties of the Archetypes_properties.xls/THERMAL. With this update, the type of windows, walls and roof are strongly related to a certain U value / Gvalue and etc. This will allow to create new construction material in CEA in agreement with different building certification approaches.

2. It makes all the files in the building-properties folder a type .dbf instead of .shp.  If the case studies have .shp this is not a problem, so no need to change previous case studies. .dbf make better the managment of data and avoids changes in geometry of one file to be reflected in any other shapefile. @amraladdin means that everytime you wanna map a value from the building-properties folder it is necessary to do a join to the zone.shp file to visualize it. Please take into account this in the documentation


JIMENOFONSECA:
@martin-mosteiro could you check if this works for you for the case in examples?

martin-mosteiro:
@JIMENOFONSECA I fixed a minor error (an empty `def` line). Also, this branch requires pysal. After installing it, the demand calculation works fine.

The documentation needs to be updated to include installing pysal.

gabriel-happle:
@JIMENOFONSECA did you completely remove the 'f_cros' values from the building properties?
It's kind of needed for the detailed infiltration calculation. But it should not be a user input/modifiable value.

JIMENOFONSECA:
Yes i did it, should it be back? Or would you hard code it?

Best, Jimeno

On 23 Mar 2017, at 15:38, Gabriel Happle <notifications@github.com<mailto:notifications@github.com>> wrote:


@JIMENOFONSECA<https://github.com/JIMENOFONSECA> did you completely remove the 'f_cros' values from the building properties?
It's kind of needed for the detailed infiltration calculation. But it should not be a user input/modifiable value.

PR #561, opened 2017-03-24
I560 rc model inconsistent inputs

martin-mosteiro:
Fixes error caused by buildings with Hs>0 and type_hs='T0' by setting Hs=0


martin-mosteiro:
@JIMENOFONSECA The same thing was happening with this PR, so I created a new one ( #581 )
This PR is now closed.

PR #563, opened 2017-03-24
556 fix broken build on master

daren-thomas:
tests should now run through - fixed some bugs.

created `cea test` command to help developers in the future to pre-test their pull requests.


daren-thomas:
With this PR, we add a `cea test` command that lets us run the jenkins tests locally before uploading, so that we can see if they run through!

PR #565, opened 2017-03-27
Revert "I9 new radiation engine daysim by paul"

daren-thomas:
Reverts architecture-building-systems/CEAforArcGIS#500


daren-thomas:
to add #500 back, just create a new PR (but first update from master and fix all the errors...)

PR #576, opened 2017-03-30
I568 heating coil inexitent globalvar

martin-mosteiro:
Removing inexistent globalvar in heating_coils.py

PR #579, opened 2017-03-31
Replacing #576 

martin-mosteiro:
The previous pull request was a mess. This one is clean and simple.


martin-mosteiro:
From #576: PR removes inexistent globalvar in heating_coils.py

PR #581, opened 2017-03-31
i560 RC model inconsistent inputes - replaces PR #561 

martin-mosteiro:
If there is no heating system, the building can't be heated. So 'Af' and 'Hs' are set to zero and a warning is printed out.

This pull request replaces PR #561, which was a mess. This new PR is clean and simple.


martin-mosteiro:
This PR fixes the error caused by buildings with Hs>0 and type_hs='T0' by setting Hs=0 and replaces PR #561 .

PR #594, opened 2017-04-10
proposal for fixing credits/licenses

daren-thomas:
This cleans the LICENSE file back to standard MIT and places the copyright with Architecture and Building Systems.

The CREDITS file then lists the contributors per version. The current version is now 2.2.

@JIMENOFONSECA where do Paul Neitzel, Thuy-Ann and Tim go? Also, under each contributor, a few bullet points (3rd level) on what the contribution was could be added.

What do you think?


JIMENOFONSECA:
One quesiton, if there is Ver 2.2 where are the releases from ver. 0.3 on? should not there be a release for every version? I believe we are still on the beta side

JIMENOFONSECA:
I added Zenodo (https://zenodo.org/), which allows to automatically get a list of authors and a DOI identifier for every realease.

I made a release to test it and it works quite cool.

Check the results of the DOI in the Readme.srt file,

In my opinion, with Zenobo, the AUTHORS/CREDITS.md file would not be necessary.

daren-thomas:
@JIMENOFONSECA regarding version numbers: I think they should reflect milestone numbers and we just finished 2.1 (documentation sprint). You're right: This should be beta, but for the `pip install`, I need a base version that is not beta, as without specifying a specific version, that is the one used. I guess we still are beta.

We should consider adding a new file (CREDITS.md) very carefully, as it would create work. At the same time, I'm not entirely comfortable with using zenodo as the only way to document this - what if they stop being a company? How about a dual strategy? DOI with zenobo is sooo cool! But keep CREDITS.md as a changelog thing? We can maybe add a paragraph to explain what happened to versions inbetween?

I personally don't think it matters much to have a version number > 1.0 - it's just a number. And this way we can have milestones and version numbers match. Maybe I can reduce the version number to 2.1b0, what do you think? Also, we should be looking into the next milestone (2.2? 2.3 - if I can't undo the version stuff on pypi...)

JIMENOFONSECA:
Agreed!

PR #597, opened 2017-04-12
I596 occupancy density

### LINKED ISSUE #596: Hiding Occ_m2p and make it part of schedules
The occupancy density variable Occ_m2p should be in the Archetypes_schedules.xls. This makes easier calculating schedules with a mix-use

JIMENOFONSECA:
This PR resolves #596


JIMENOFONSECA:
@daren-thomas with help of @gabriel-happle i set up the unittests of the demand. If it passes should we merge?

JIMENOFONSECA:
merging as needed for the new release and it passes the checks

PR #604, opened 2017-04-18
NSIS setup

daren-thomas:
This PR instroduces an installer for the CEA that creates a self-contained miniconda installation in LOCALAPPDATA (no admin rights required) and sets up everything for the planner. 

See example installer exe below - created using the NSIS installer software.

Setup downloads miniconda, installs it, used that to install geopandas and ephem and then does a pip install cityenergyanalyst (from pypi) for the version of CEA used to create the installer. I'm pretty proud of the last part. Then, the installer runs `cea install-toolbox`. I've tested a bit with a virtual machine. 

@JIMENOFONSECA can you test this on your machine? It should not destroy your python, as it installs a separate one...


daren-thomas:
You can try out the installer here: https://www.dropbox.com/s/19jiriglp6ulz7e/Setup_CityEnergyAnalyst_2.2a4.exe?dl=0

JIMENOFONSECA:
It installs successfully, now how can i test the scripts? is this adding the plug-in to ArcGIS directly?

PR #635, opened 2017-05-15
correct conversion of string to integer if dbf file is defined as double accidentally

gabriel-happle:
correct conversion of string to integer if dbf file is defined as double
accidentally


gabriel-happle:
@JIMENOFONSECA you need this bug fix to read the inputs of EcoCampus.

Issue #655, opened 2017-05-23
Move vent control to ventilation?


JIMENOFONSECA:
I wonder if the tab of ventilation control can be moved to that of ventilation and simplify the inputs. to only type_vent @gabriel-happle what do you think? I would need to check this for the issue #643 . we could sit down and check it also how we can rework it.

![image](https://cloud.githubusercontent.com/assets/8973186/26344269/96928bd8-3fd1-11e7-990c-2a8c2bee000f.png)

![image](https://cloud.githubusercontent.com/assets/8973186/26344282/a15cfcc4-3fd1-11e7-8f3f-edb8ed437014.png)

gabriel-happle:
@JIMENOFONSECA `type_ctrl` is not the control of ventilation. It's the control of the heating and cooling space emission systems used to calculate emission losses.

PR #656, opened 2017-05-24
I655 ctrl vent changes

gabriel-happle:
issue #655 

This should work. Please update from master and test.


gabriel-happle:
@daren-thomas and @JIMENOFONSECA tests are failing because I changed the only function that actually has a unit test ;-) 
How can I run dodo without the `test_sensible_loads.py`?

JIMENOFONSECA:
it looks good, just it needs to pass the unittests

daren-thomas:
@gabriel-happle you can't. but you can update the expected results to be the new ones... huh. I just ran the tests and it seems you changed the signatures of the functions. so. your next step would be to figure out:

- are the tests still the right tests for the code?
- if not, what tests should be there instead?
- update the tests to run through

gabriel-happle:
Exactly @daren-thomas we changed the signatures because we got rid of the hard coded values that these functions were looking up in a dict. They are now just getting something from inside the `bpr` object. Therefore it's hard to check for correct inputs in the function, as the `bpr` can basically contain anything.
I'm sorry I changed the only function with unit tests :(
But I don't know what to check now. The relationship between inputs and outputs is taken care of in the Excel database.
Basically there should be some check during the creation of `bpr` that checks whether all necessary data is present.

gabriel-happle:
C:\Users\Gabriel\Anaconda2\envs\cea\python.exe C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/cea/tests/test_calc_thermal_loads.py
read input files
Building B06 has no heating system, Hs corrected to 0.
done
..
Ran 2 tests in 11.127s

OK

Process finished with exit code 0
```

PR #671, opened 2017-06-09
Cascading config files

daren-thomas:
cascading config files and a new way to specify heating seasons (and cooling seasons)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,4 @@ cea/.doit.db
 setup/cea_version.txt
 .doit.db
 docs/guide_about_files_generated.html
+docs/cascading-configuration-files.html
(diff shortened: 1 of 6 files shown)


JIMENOFONSECA:
@daren-thomas great stuff, would there be an arcgis interface for this?

daren-thomas:
@JIMENOFONSECA this will tie into everything eventually. Yes. Of course an ArcGIS interface for user / project / scenario configuration. But also this is a place to store the last chosen options for each tool :-)

This PR doesn't actually change the functionality of the CEA yet - it is just exploring how to make the config files work and is the basis for slowly removing functionality from `gv` and adding it to these config files. I expect to do that one issue at a time to avoid a huge mega-branch as that has proven to interrupt our flow.

JIMENOFONSECA:
@daren-thomas cool, we merge it then. I am ok with the documentation too.

Issue #612, opened 2017-04-20
Preprocessing doesn't work well for multi-use buildings


martin-mosteiro:
`calc_mainuse` only works if a building has up to two uses: it takes the buildings' uses and generates an array of uses with the largest area (`array_max`) and with the smallest area (`array_min`). If the main use is 'PARKING', the building would have no heated area even if the building was 51% parking and 49%, say, offices. So `calc_comparison` corrects this by defining the building's main use as `array_max` if the use is not 'PARKING', and as `array_min` otherwise. Now what happens if a building is 50% 'PARKING', 49% 'OFFICE' and 1% 'COOLROOM', for example? The building gets defined as 'COOLROOM', even though that use only accounts for 1% of the area (yes, this actually happens for one building in the Hochschulquartier)

Fixing `calc_mainuse` to take the second most important use (instead of the least important) would be an easy fix, but more importantly: does it make sense to apply the archetypes based on a single use?

For some properties (say, construction properties), it might arguably be okay. However, for all properties that depend on the square meters of each type of use (e.g. Occ, Es, Hs, Ve, all internal loads) it certainly doesn't. Why should a building with say for example 34% office, 33% residential and 33% gym require only 10 L/p/d of hot water?

martin-mosteiro:
@daren-thomas You said this needed discussion?

martin-mosteiro:
So here are the effects of this preprocessing on the calculations done on CEA:

1. **demand/electrical_loads.py**

`calc_Eint` calculates the following:
```
tsd['Eaf'] = schedule_applicances_lighting * bpr.internal_loads['Ea_Wm2'] * bpr.rc_model['Aef']
tsd['Elf'] = schedule_applicances_lighting * bpr.internal_loads['El_Wm2'] * bpr.rc_model['Aef']
```
However, **Ea, El and Aef** are only defined for the main use type, so if the building is, say, 51% RESTAURANT and 49% percent LIBRARY, the electricity demand will be overestimated.

Furthermore, if the occupancy types include COOLROOM, SERVERROOM or INDUSTRY, the script calculates the following, respectively:
```
tsd['Eref'] = calc_Eref(schedule_Eref, bpr.internal_loads['Ere_Wm2'], bpr.rc_model['Aef'])  # in W
tsd['Edataf'] = calc_Edataf(schedule_Edata, bpr.internal_loads['Ed_Wm2'], bpr.rc_model['Aef'])  # in W
[...]

Issue #664, opened 2017-05-31
As a researcher, I want a robust and up-to-date multi-zone demand calculation, so I can simulate a higher level of granularity in terms of uses


JIMENOFONSECA:
It refers to a step that can be done in parallel to issue #662. In basic therms the output of the radiation tool will not change.

This issue entails:
- changing occupancy.dbf in reference case open to consider multizonning (Jimeno). Floor by floor. Indicate in the radiation file which surface belongs to which floor.
- making multizonning in CEA (Gabriel) and connecting to radiation. (for this, it is good to create "zones" when some floors have the same function. We should avoid doing the analysis floor by floor.

JIMENOFONSECA:
what happens with funcitons that have an area less than one floor? @martin-mosteiro

JIMENOFONSECA:
![b3abdc60d729f442eeaebc9dcb2a8241](https://cloud.githubusercontent.com/assets/8973186/26722427/7d552e20-47c2-11e7-8e0c-d4555717d698.png)

JIMENOFONSECA:
@gabriel-happle i made the occupancy.dbf with the zoning convention. you can take this issue from here. I did it for the open case for now. 

Please continue the work in the branch i664-multi-zone-radiation

gabriel-happle:
@JIMENOFONSECA I'm working on that in branch i664-multi-zone-radiation but I prefer that @martin-mosteiro 's mixed-use modifications are merged first. I will prepare the database structure of the thermal zones first in theory. But this issue can not be finished in this sprint.

JIMENOFONSECA:
@gabriel-happle @martin-mosteiro i finished my part in here. The part about identfying the floor of each surface could not be completed. let me know if you need help.

JIMENOFONSECA:
@gabriel-happle @martin-mosteiro status?, this is being a bottle neck in the work we are doing with @Khayatian any chance you can put it as a priority?

gabriel-happle:
OK @JIMENOFONSECA . I will make this my priority now.

gabriel-happle:
check #935 and split up into tasks

gabriel-happle:
I converted this to an epic as it contains all the information.
I added all the issues connected to the R-C-model in SIA/ISO to this Epic.
While implementing the multi-zoning, they can be implemented easily.

JIMENOFONSECA:
2 years here, we will not have it anytime soon. Closing it.

PR #675, opened 2017-06-13
I612 preprocessing multiuse buildings

martin-mosteiro:
As discussed in #612, this issue tackles a few different things:
1) Up to now, each building's main use was set the wrong way when most of the building area was of the type 'PARKING'
2) The demands of the building were all set by the building's main use, even in mixed use buildings
3) Schedules did not distinguish the fact that, for example, 10 people working out at a given time do not generate as much humidity gains as 10 people working in an office at the same time

In order to fix these issues, the following was done:
1) Adjust calculation such that if most of the building is used for 'PARKING', the next most common use becomes the main use
2) The archetype values for internal loads, ventilation, heated and electrified are calculated based on an average of the value for each use in the building (rather than being defined for the main use only)
3) Schedules for each type of internal load, ventilation and occupancy are generated for each building based on the schedule database and the mix of uses in the building. these schedules are normalized by the building's archetypal ventilation demands and internal loads such that the building's value can be defined through the corresponding input shapefile while maintaining the demand profile generated by each use type.


JIMENOFONSECA:
@gabriel-happle please review this PR, It would need to pass first than #664

gabriel-happle:
@martin-mosteiro and @JIMENOFONSECA I have started testing the functions in the `properties.py`.
Looks good.

gabriel-happle:
If you replace `properties_df[column][building]=...` with `properties_df.loc[building,column]=...` you can get rid of those warnings:

```
C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/cea/demand/preprocessing/properties.py:253: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  properties_df[column][building] = column_total/people_total
C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/cea/demand/preprocessing/properties.py:254: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  else: properties_df[column][building] = 0
C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/cea/demand/preprocessing/properties.py:261: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  properties_df[column][building] = average

```

gabriel-happle:
I checked the detailed reports, e.g. for occupancy and internal humidity gains for some created mixed use buildings in the open case study. It seems to produce the indented results.
I think we can merge.
@martin-mosteiro I think you have to update the calc_thermal_load tests in order to pass and maybe you want to get rid of those warnings. Especially the `RuntimeWarning: invalid value encountered in true_divide` sounds a bit dangerous.

JIMENOFONSECA:
@martin-mosteiro i believe it works quite well, once it passes the tests we can merge. please connect the PR to an issue. thanks!

martin-mosteiro:
@gabriel-happle, @JIMENOFONSECA thanks for your review, I'll fix these issues today

martin-mosteiro:
The values in the unit test had to be updated since there was an error in the old calculation for water and hot water demand. The thermal loads test now passes on my computer.

PR #680, opened 2017-06-14
566 code the program to call cea from gh

daren-thomas:
adds a file "urban_design/CEA demand.ghuser" that can run the CEA demand, provided it is installed for ArcGIS already.


JIMENOFONSECA:
@shizhongming can you tell us if this works and should be part of CEA?

Issue #642, opened 2017-05-15
ArcGIS interface: Remember last selected options for a scenario (plus last selected scenario)


daren-thomas:
This could be part of the cascading configuration file, but does not necessarily have to be. When a user runs a feature in the ArcGIS interface, make that the selected choice the next time the user starts the interface.

- pre-select the last chosen scenario
- read the last options of that scenario from the scenario-level configuration file?

This depends on the configuration file, I think, but given that, it is pretty straightforward to implement.

PR #686, opened 2017-06-20
642 arcgis remember last options

daren-thomas:
This PR adds a "memory" of options chosen in the ArcGIS interface for most tools. RadiationTool and HeatmapsTool are exceptions (they are complicated and adding the memory would just mess everything up)

Run `cea install-toolbox` to get the newest version of the interfaces and try them out!

This implements #642.


JIMENOFONSECA:
@daren-thomas I tried with the demand tool. it does not work. I believe this feature can be replaced by the existing results window of arcgis. 

In that window you can see the last values incorporated into the tool once you run it . Even if there were errors. I attach some pics

image

image

daren-thomas:
@JIMENOFONSECA how does it not work? what was the error?

this feature goes beyond the arcgis results window in that it stores the settings with the scenario. And also allows to have project / user defaults...

JIMENOFONSECA:
Maybe we can talk on skype. Let me know when you are connected. I do not fully understand what this PR should do.

Best, Jimeno

On 10 Jul 2017, at 18:49, Daren Thomas <notifications@github.com<mailto:notifications@github.com>> wrote:


@JIMENOFONSECA<https://github.com/jimenofonseca> how does it not work? what was the error?

this feature goes beyond the arcgis results window in that it stores the settings with the scenario. And also allows to have project / user defaults...

PR #690, opened 2017-06-21
I673 cea optimization workflow

BhargavaKrishnaSreepathi:
I added the `xml` files used for making the workflow diagram, and also provided a guide to reproduce the results. This is added in the `CEAforArcGIS\docs\optimization_workflow`


BhargavaKrishnaSreepathi:
@JIMENOFONSECA: only the guide to generate a workflow will be visible in developer's documentation as it is the `rst` file. @daren-thomas: any comments on the way it is organized, as most of the files are `xml` files which can be modified further

Issue #699, opened 2017-06-27
Radiation script: Non_overlap does not exist


daren-thomas:
When running the radiation script in an emtpy `Default.gdb` geodatabase, on the reference-case-open, I get the following error: 

> ERROR 000732: Remove Features: Dataset C:\Users\darthoma\Documents\ArcGIS\Default.gdb\Non_overlap does not exist or is not supported

As far as I can tell, this is because the file is only generated, when `rows` in line has a length > 0. This is only the case when the file `DataFactorsBoundaries.csv` in the temp folder has data. After running the reference-case-open, that is not the case. I assume `Non_overlap` is never deleted and therefore this code runs through if another case has been run on that machine where `DataFactorsBoundaries.csv` did have data.

daren-thomas:
Running the reference case for Zurich worked.

daren-thomas:
After running the reference-case-zurich, the reference-case-open runs through. This, in my opinion, proves that the above diagnosis is correct. @JIMENOFONSECA is there a way to move these two lines:

```
arcpy.MakeFeatureLayer_management(Simple_CQ, 'Simple_lyr')
# ...
arcpy.CopyFeatures_management('simple_lyr', NonoverlappingBuildings)
```

Out of the for loop, or, handle the case that they are not executed at all?

PR #702, opened 2017-06-29
fixing special case radiation

JIMENOFONSECA:
it solves #699


daren-thomas:
@JIMENOFONSECA this works! You're my hero!

Issue #672, opened 2017-06-12
Create mini tools converter dbf 2excel and viceversa


JIMENOFONSECA:
I wonder if @Khayatian  this could be your first contribution to CEA.

In essence we are missing two tools that you can call from Arcgis to convert files and make them editable from excel or access.

The result:

1 tool in arcgis called: database to excel converter
1 tool in arcgis called: excel to database converter

The method:

1. order a bit the file we have to do this conversion . cea/utilities/dbfreader.
2. Liase with @daren-thomas an interface for arcgis.

PR #703, opened 2017-06-29
#672 xls2dbf && dbf2xls

Khayatian:
the required script for adding two tools: xls2dbf and dbf2xls


JIMENOFONSECA:
@martin-mosteiro  would you mind being the second pair of eyes in here?

martin-mosteiro:
Can

martin-mosteiro:
Checked the script, seems to work as intended. Really great to have!
Only comment: if the output file has the wrong extension (i.e., neither xls nor dbf), there could be an error message. At the moment it just creates a file that can't be executed.

Khayatian:
tnx for the feedback Martin, will address it.

Issue #383, opened 2016-10-25
Process heating calculation


martin-mosteiro:
In the Universitätsspital, about 20% of the total heating demand is for processes (e.g. sterilization). These need higher supply temperatures than either space heating or hot water. Currently, this demand is supplied by the vapor-based district heating network (192°C as opposed to 90-120°C for the water-based network).

This demand should therefore be calculated separately. In the current version of the CEA, process heating is simply set to zero. Obviously, a calculation method is needed.

martin-mosteiro:
Update: from the Hochschulquartier reference yearly data we can easily calculate the process heating demand for hospitals (sterilization processes, e.g.) - this should be added to the archetype database for building type 'HOSPITAL'.

martin-mosteiro:
The process heat calculation follows the same structure as the electricity demand calculations (schedule * reference value in W/m2 * Af). The same schedule as for process electricity is assumed. The reference value was calculated using a top-down approach as follows:

1. Process heat by sector (in PJ/yr) was taken from Bundesamt für Energie (2011) _Grundlagen für die Energiestrategie des Bundesrates - Aktualisierung der Energieperspektiven 2035_. The process heat from the “services” sector was assumed to come entirely from the hospital sector.

2. Total built area by sector was taken from Bundesamt für Energie (2004) _Zukünftige Entwicklung der Energiebezugsflächen: Perspektiven bis 2035_.

3. The hours per year were calculated based on the CEA process energy schedule as approximately 2000 hours per year. The same value was assumed for hospital processes.

This gives a demand for process heat of about **150 W/m2 for industry** and **16 W/m2 for hospitals**.

In order to test the feasibility of these values, I adapted reference-case-open to include an industry building. Process heat accounted for 73% of the total yearly heating demand in the building, which is comparable to the reference values in BFE (2011) (78% of heat demand in industry is for process heating).

PR #707, opened 2017-06-29
I383 process heat

martin-mosteiro:
This PR incorporates a calculation for process heat, which was up to now set to 0 in all cases.

Furthermore, it also solves an issue by which process electricity was never actually calculated in the CEA.

Reference values for process heat for industry and hospitals were estimated based on national energy statistics (see #383).

The schedule for process heat in hospitals was assumed to be the same as for industry, but seven days a week (i.e., the schedule for sundays is the same as for saturdays).


JIMENOFONSECA:
@martin-mosteiro there are some conflicts with master

martin-mosteiro:
@JIMENOFONSECA done

Issue #583, opened 2017-03-31
Hot water loads create an error when V=0


martin-mosteiro:
hotwater_loads.py includes the following code (lines 172 to 177):
```
try:
    exponential = scipy.exp(-(q * Lsww_dis * t) / (p * cpw * V * (twws - tamb) * 1000))
except ZeroDivisionError:
    gv.log('twws: %(twws).2f, tamb: %(tamb).2f, p: %(p).2f, cpw: %(cpw).2f, V: %(V).2f', twws=twws, tamb=tamb, p=p, cpw=cpw, V=V)
    exponential = scipy.exp(-(q * Lsww_dis * t) / (p * cpw * V * (twws - tamb) * 1000))
```

Basically: try to calculate exponential, unless you get an error, in which case... do the same calculation that caused the error. This looks like a mistake, @JIMENOFONSECA ?

JIMENOFONSECA:
Well, it turns out that sometimes we run the algorithm twice and it converges... it might be an issue with the seed. this is the fastest way to fix it.. if an issue, i would put it priority 3

________________________________

daren-thomas:
how can this converge by running it twice? it's a mathematical function! i don't see any seed in this code. am i blind?

@JIMENOFONSECA can you please look at the code mentioned above again. There must be another reason for doing this twice. Or none at all.

martin-mosteiro:
@daren-thomas that's what I mean. This is just an exponential - if there's a division by zero the first time, why wouldn't there be the same issue when you repeat the calculation with the same exact inputs?

JIMENOFONSECA:
@martin-mosteiro you are right. sorry i confused the issue with another function where we apply bisect and netwton to know the return temperature of the radiator. Yes this does not make sense, please erase..

martin-mosteiro:
@JIMENOFONSECA what should the erased line be replaced by?
ValueError?
Or: if (p * cpw * V * (twws - tamb) * 1000)) == 0, then the exponential tends to 0...?

gabriel-happle:
@martin-mosteiro I modified the milestone as discussed. but I also noticed that nobody is assigned to this issue!

martin-mosteiro:
Thanks @gabriel-happle. I also completely missed this since it was assigned to the previous milestone. I believe this issue was already fixed by an older pull request of mine, but I'll check. I'm assigning myself.

martin-mosteiro:
[...]

PR #716, opened 2017-06-30
Adding solution for case when V=0

martin-mosteiro:
If the denominator is zero and the numerator isn't, the exponential will tend to zero. However if both are zero an error is raised.


martin-mosteiro:
After discussion with @gabriel-happle in #583, it was decided there could be a viable case when this error could appear, so the error is fixed in that case.

Also, I found a typo in one formula, which I corrected.

daren-thomas:
@martin-mosteiro why is it failing the tests?

daren-thomas:
@gabriel-happle @JIMENOFONSECA merge? from my side it seems fine...

martin-mosteiro:
@daren-thomas is it?

Issue #536, opened 2017-03-08
Develop test case for testing the thermal and hydraulic losses


BhargavaKrishnaSreepathi:
Develop a test case to test the thermal_network_matrix.py file. This will be further useful in future to test the hydraulic and thermal losses for network optimization. 

Develop two/three network configurations for the 9 building case study

martin-mosteiro:
Great idea! The network shapefiles are something that needs to be incorporated into the reference cases, anyway, so it obviously makes sense to include them in testing as well.

shanshanhsieh:
@martin-mosteiro  shall I include you in this issue, then?  : ) 
Since it will require two parts:
1. Prepare the shape files of three testing networks for reference-case-open
2. Simulate those three testing networks in simulink

shanshanhsieh:
Note for preparing `network-nodes.shp` and `network-edges.shp` on ArcMap:

1. `network-nodes` add following columns in the attribute table:

- Building Names (join from `zone.shp`)

- Plant: specify which buildings are plants (if yes, value = 1, otherwise 0) 

- Qh: specify which buildings are consumers (if yes, value = 1, otherwise 0)



2. `network-edges` add following columns in the attribute table:

- Shape_Leng: Calculate Geometry, length

**Note**
1. Spatial References of `cea-reference-open`: 

- Projection:

 CH1903_LV03 (automatically include global reference: GCS_CH1903)
2. Make sure the x,y coordinates read from geopandas of edges and nodes are the same. (to check: read x, y coordinates to the attribute table)

shanshanhsieh:
@BhargavaKrishnaSreepathi  This issue is ready to be reviewed!
To test the three networks, please go to branch `i412` and run the reference case in `cea/examples/reference-case-open`, there is a readme guide on how to switch between three network configurations.
To verify the result with simulink model, please access the model via https://github.com/shanshanhsieh/network.

Issue #412, opened 2016-11-18
Plotting potential DHC network paths


shanshanhsieh:
Before the network layout optimization, a grid of potential paths should be prepared. Where buildings and plants are connected should be specified. This include streets, pedestrian pathways, additional that the designer considers could be interesting.

Possible steps:
1. Extract shape file via Shapely: for each node, store information of actual location and distances to other nodes. Create an artificial mesh for network connection.
2. Use NetworkX library to describe the connections and the relationship between nodes.

Issue #409, opened 2016-11-18
Piping database for DHW network


shanshanhsieh:
one excel file with the costs of different types of pipes and the properties of the materials. this should be stored in the database of the cea. Check publication in energy and buildings for these parameters as a starting point.

Issue #408, opened 2016-11-18
Thermal model for DHC network


shanshanhsieh:
Add the thermal model of DHC network into the technologies folder.

For the thermal losses you might want to check:
summarize_network_main.py/calc_piping_thermal_loss. But i suggest to do some research about this

PR #727, opened 2017-07-11
I412 dhc network path

shanshanhsieh:
This pull request is connected to issue #407 #408 #409 #412 #536 

**Changes include:**
1. A new script `...technologies/thermal_network_matrix.py` for thermal network simulation (#407, #408)
- `substation_matrix.py` is updated in order to connect with the new network simulation script
- the modeling procedure is described in the documentation of `thermal_network_main` function 
- a function to extract network edges and nodes from `.shp` files (`get_thermal_network_from_shapefile`) ( #412 )
- piping database is added to `...cea/databases/systems/thermal_networks` 

2. `CEAforArcGIS/validation/thermal_network_matrix` ( #536 )
- Three test networks built from "open case study" for validation in `test_network_open_case`
- Simulink models of the three test networks in `simulink_models`
- The comparison between cea & simulink results are documented in `CEA-model-thermal-hydraulic-network.rst`

3. NOTHING is changed in the `...cea/examples/reference-case-open` folder, feel free to overwrite it.

**To test `thermal_network_matrix.py` :** 
- Zug network:  
1. Change scenario_reference to `reference-case-zug`
2. Run `demand_main.py`
3. Copy `NodesData_DH.csv` and `PipesData_DH.csv` from `...baseline\outputs\data\optimization\network\layout`  to `...baseline/inputs/network/DH` (same for DC) 
4. Run `thermal_network_main` in `thermal_network_matrix.py` by indicating `network_type` (DH or DC) and `source` (.csv)
- Open case study:
1. Change scenario_reference to `reference-case-open`
2. Run `demand_main.py`
4. Copy network `.shp` files from the `test_network_open_case`  to `...baseline/inputs/network/DH` or `DC`
3. Run `thermal_network_main` by indicating `network_type` and `source` (.shp)

**To run the simulink models:**
- The simulink models are compatible with MATLAB 2017a 
- run `run_network1.m` , `run_network2.m`, `run_network3.m` to execute simulations (_note: it takes ~9 hrs to run all 8760 time-steps_)
- the models could be seen in `pipelines_network1.slx`, `pipelines_network2.slx`, `pipelines_network3.slx`


daren-thomas:
@shanshanhsieh @martin-mosteiro how to test this?

shanshanhsieh:
@daren-thomas , to test, one will have to run `thermal_network_matrix.py` for Zug and the open case following the procedure I wrote in the pull request. (let me know if any part is unclear) 
Bhargave had went through the script and the validation procedures with me several times. I think we also have a 'go' from @BhargavaKrishnaSreepathi ? : )

BhargavaKrishnaSreepathi:
@daren-thomas: I have gone through the script, and also through the results generated by it for two examples. @shanshanhsieh performed an analysis comparing the results generated by this script and to a commercially available software. The results are within an acceptable range of approximation. These are summarized in a `rst` file attached in the pull request

JIMENOFONSECA:
@shanshanhsieh coudl you solve the conflicts? for the reference-case-open, make sure you use the one today in master and add your new work in there again.

shanshanhsieh:
@JIMENOFONSECA I changed the `reference-case-open` to the same one from master.

If you want to run the network simulation for `reference-case-open` with the networks I made, simply copy the `.shp` files from `...validation/thermal_network_matrix/test_network_open_case`.

cea/inputlocator.py
```diff
@@ -468,6 +474,33 @@ def get_edge_mass_flow_csv_file(self, network_type):
         return os.path.join(self.get_optimization_network_layout_folder(), 'NominalEdgeMassFlow_' +
                              network_type + '.csv')
 
+    def get_network_layout_edges_shapefile(self, network):
```
JIMENOFONSECA: I suggest to change the input folder name to "thermal-networks". It makes it easier to differentiate with prospect electric networks.

Inside the folder of thermal-networks I suggest to include the shapefiles without any extra subfolders. it would make easier to write documentation. an example for three types of networks.

- type1_nodes.shp
- type1_edges.shp
- type2_nodes.shp
- type2_edges.shp
- type3_nodes.shp
- type3_edges.shp

cea/technologies/thermal_network_matrix.py
```diff
@@ -0,0 +1,1537 @@
...
+    # # prepare data for calculation
+
+    # read building names
+    total_demand = pd.read_csv(locator.get_total_demand())
+    building_names = total_demand['Name']
+
+    # calculate ground temperature
+    weather_file = locator.get_default_weather()
+    T_ambient = epwreader.epw_reader(weather_file)['drybulb_C']
+    T_ground = geothermal.calc_ground_temperature(T_ambient.values, gv)
```
JIMENOFONSECA: If possible, it would be good to avoid the dependency in GV. for this please modify the way ground temperature is calculated.

cea/inputlocator.py
shanshanhsieh: The folder name can be changed to `thermal_networks`
However, the shape files for nodes and edges have to be shipped with multiple dependent files so that geopandas can import all information from the attribute tables. It will be difficult to organize without having sub-folders.

cea/technologies/thermal_network_matrix.py
shanshanhsieh: the function `calc_ground_temperature` is modified

shanshanhsieh:
@JIMENOFONSECA 
In the thermal_networks.xls:

- the format is changed to match with other database

- The "C" is removed from the PIPING CATALOG since it is not used in `thermal_network_matrix.py`, instead, these numbers will be represented as linear functions of pipe sizes in the cost calculation.

- The variable names are updated and the note tab is removed.

Other things:

- Regarding the folder structure in `reference_case_open`: it is not easy to delete the subfolders for the three networks, since there are multiple files that needed to be shipped with `network-edges.shp` and `network-nodes.shp` for geopandas to read the attribute table.

- I can include the .shp files of the three networks to the `reference-case-open` once we agree on the folder structure.

- For the zug case study, we only have the .csv file that contains the network edge/node info. That is the reason why .csv is a type of input for network geometry.

- The standard way of reading network geometry from ArcGIS will be .shp

cea/inputlocator.py
shanshanhsieh: The  `reference-case-open` now includes the .shp files of network nodes/edges under this directory: `...baseline\inputs\thermal-networks\DC or DH`. In `DC`, a network with single plant is set as default, while `DH` has network with two plants. The rest of the network `.shp` files are stored in `validation\thermal_network_matrix\test_networks_open_case`.

shanshanhsieh:
@JIMENOFONSECA , the branch is modified according to your requests. Please take a look and let me know if further modification is needed.

- The  `reference-case-open` now includes the .shp files of network nodes and edges under this directory: `...baseline\inputs\thermal-networks\DC or DH`. In `DC`, a network with single plant is set as default, while `DH` has network with two plants.

- Regarding the input with `.csv` or `.shp`, the major difference is whether the network inputs are prepared by the users or generated by optimization.
If the network is prepared by the users in ArcGIS,  `.shp` files will be used as inputs. On the other hand, we imagine it is easier to produce `.csv` files from network optimization, in this case, the `.csv` files are used as inputs.

- For the zug case study, we only have the .csv file that contains the network edge/node info at the moment.

JIMENOFONSECA:
@shanshanhsieh good work here however, there are some things we should address:

- The network should work at least for a typical minimum spanning tree with branches. See the DH network I included in examples:
- The fields of each network file should address the material and size of the pipe (output of the optimization).

 I made some adjustments in the code to allow for this. please check the function xtract_network_from_shapefile(edge_shapefile_df, node_shapefile_df) for more details. there are a couple of tricks to make the code faster, that it would be good for you to check (e.g., list comprenhensions).

image


image

JIMENOFONSECA:
@shanshanhsieh it runs until half, however it is not clear how this can track also the massflow rates in the 'dummy'nodes as well as the remaining edges. I believe we should take a look at this, as the current network typology is not realistic.

I included one that may consider branches for the heating case.

shanshanhsieh:
@JIMENOFONSECA i am currently updating the code, so it runs with the `.shp` format you created, especially, making sure the columns are in the right orders.
The flow rates on each edge are always tracked, i will include them in the output files.

JIMENOFONSECA:
Hi shanshan, i did big part of this yesterday please sync before hand :-)

Best, Jimeno

On 1 Aug 2017, at 09:30, shanshanhsieh <notifications@github.com<mailto:notifications@github.com>> wrote:


@JIMENOFONSECA<https://github.com/jimenofonseca> i am currently updating the code, so it runs with the .shp format you created, especially, making sure the columns are in the right orders.
The flow rates on each edge are always tracked, i will include them in the output files.

shanshanhsieh:
@JIMENOFONSECA the scripts are updated to read the new .shp file format. 
Both DH and DC in the open case study, as well as the .csv files from the zug case study, are tested.
The pipe properties are included in the output file `DH_AllEdges.csv` and the mass flow on all edges are included in `DH_MassFlow.csv`
I think all the requested changes are addressed now, let me know if there is anything else!

PR #741, opened 2017-07-20
I733 cooling conditioned buildings

gabriel-happle:
fix conditioned area correction for buildings with only cooling systems


gabriel-happle:
@JIMENOFONSECA could you please take a quick look at this and merge. It prevents that the conditioned floor area of buildings that only have cooling systems (and no heating systems, i.e. Singapore) gets corrected to be zero.

PR #763, opened 2017-07-26
I672 xls2dbf

Khayatian:
modified xls2dbf, ready for testing


daren-thomas:
@JIMENOFONSECA this should be a working implementation of @Khayatian 's tool. I did some testing, but four eyes are better than two... to test:
 
- `cea install-toolbox`
- City Energy Analyst.pyt -> Utilities -> Convert DBF to Excel
  - choose dbf file path, choose excel output file path, run!
  - open in excel...
- City Energy Analyst.pyt -> Utilities -> Convert Excel to DBF
  - choose xls file path, choose dbf output file path, run!
  - ? maybe just convert back to excel and compare??

PR #769, opened 2017-07-28
705 update planning installer

daren-thomas:
@JIMENOFONSECA this updates the planning installer and also concludes the move to 64 bit python.

BTW: The planning installer is a great way to install CEA as it basically automates the "conda install..." steps from the installation guide. Maybe with a little bit of effort, this could be updated to include a link to the command line and also to git clone the repository: voilà an installer for developers / researchers!


JIMENOFONSECA:
@daren-thomas one question, how can I test it? is there a place where the .exe file is located?

daren-thomas:
@JIMENOFONSECA: great question. the docs here are incomplete: http://city-energy-analyst.readthedocs.io/en/latest/how-to-create-a-new-release.html#creating-the-installer-for-the-planner-s-edition

this is the installer: https://github.com/architecture-building-systems/CityEnergyAnalyst/releases/download/v2.4a1/Setup_CityEnergyAnalyst_2.4a1.exe

when we finish the milestone, we can create an installer for 2.4 (without the alpha status)

Issue #774, opened 2017-08-02
Ensure `cea test` can be run without access to private reference cases


daren-thomas:
When run without specifying one of the private reference cases, just extract the example reference case and run the tests on that - without checking for `cea_github.auth` in the home folder.

To test this, just delete `~/cea_github.auth` and run `cea test` - currently it breaks with this error:

```
TaskError - taskid:download_reference_cases
PythonAction Error
Traceback (most recent call last):
  File "E:\cea\lib\site-packages\doit-0.29.0-py2.7.egg\doit\action.py", line 383, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "E:\cea\lib\site-packages\cityenergyanalyst-2.3-py2.7.egg\cea\tests\dodo.py", line 110, in download_reference_cases
    r = requests.get(REPOSITORY_URL % REPOSITORY_NAME, auth=get_github_auth())
  File "E:\cea\lib\site-packages\cityenergyanalyst-2.3-py2.7.egg\cea\tests\dodo.py", line 85, in get_github_auth
    with open(os.path.expanduser(r'~\cea_github.auth')) as f:
IOError: [Errno 2] No such file or directory: 'C:\\Users\\darthoma\\cea_github.auth'
```

Issue #768, opened 2017-07-28
Add an interface to extract reference case to the ArcGIS toolbox


daren-thomas:
This is a really small task, but could be quite useful to new users: Extract reference case. Allow the user to choose destination folder. Extract the built-in reference case. Tutorials could then start with this step.

I'd put it in the "Utilities" category?

daren-thomas:
pushing priority as this needs to get on the student's version

PR #775, opened 2017-08-02
768 add an interface to extract reference case to the arcgis toolbox

daren-thomas:
implements #768 and #774.


daren-thomas:
@JIMENOFONSECA: I need this for the student's installation at Hönggerberg. Basically, it let's users extract the sample reference case to a folder from ArcGIS (without doing `cea extract-reference-case` on the command-line) and also adds a tool to run `cea test` from ArcGIS: this let's the planner's version be tested - and the student version too!

Is there a way to get this merged quickly? To test: `cea install-toolbox` and then check the "Utilities" category for the two new items "Test CEA" and "Extract reference case".

PR #779, opened 2017-08-08
updates to the documentation - sphinx now builds without warnings

daren-thomas:
this fixes failing sphinx builds... temporarily: i'll have to do something similar every now and then just to make sure everything is a-ok.

### DIFF
--- a/cea/analysis/lca/mobility.py
+++ b/cea/analysis/lca/mobility.py
@@ -46,6 +46,10 @@ def lca_mobility(locator):
     - total_LCA_mobility (.csv)
       csv file of yearly non-renewable primary energy demand and emissions due to mobility for each building
 
+    Additional references:
+
+    - [SIA_Effizienzpfad_2011]_
+
     :param locator: an InputLocator instance set to the scenario to work on
     :type locator: InputLocator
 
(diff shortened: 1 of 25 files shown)


daren-thomas:
@JIMENOFONSECA the best was to test this:

- `cd docs`
- `make html`

The output should not include warnings and errors.

For this to work, you need the whole sphinx stuff installed. The easiest way to do that is to do this in the CityEnergyAnalyst project root folder:

- `pip install -e .[dev]`

(after some testing and learning, it seems the `pip install .` route is better than the `python setup.py install` as it will use packages from PiPy in a smarter way)

daren-thomas:
Also note, that recent merges have broken the documentation build again. I've left them for the moment so you can see firsthand what malformed documentation does to the documentation build and what such errors would look like. When you're done, let me know: I'd be happy to go over them and fix them up :-) Also, if you have time (I'm sure you don't) you can try your hand at fixing up some of the warnings yourself? Maybe I can do a small tutorial as my presentation next week for the other members of the team...

JIMENOFONSECA:
So i installed the planners version and somehow my anaconda prompt got corrupted. It changed the path to the installation of anaconda. now i will have to recreate an environmnet and install all cea again. is there a way this does not happen? 
image

daren-thomas:
oh no! that is sooo annoying. I'm sorry, @JIMENOFONSECA...

I don't think I can fix that easily. Instead, I suggest you use the new installation guide to the CEA - I made it a lot easier! Check it out on readthedocs...

Issue #684, opened 2017-06-16
Launch tool: solar isolation


JIMENOFONSECA:
- [   ] Create interface to the Daysim Radiation Tool.
- [   ] Create Tutorial How does the Urban Solar Radiation tool work?

daren-thomas:
@JIMENOFONSECA sure, I'd like to do that. Also, we should move the `settings.py` stuff to the cascading configuration files ;-)

JIMENOFONSECA:
Sure, although as every tool works independently in CEA (and it should) i wonder if every tool should have its own config or settings file rather than a mega file for all tools.

Best, Jimeno

On 16 Jun 2017, at 15:33, Daren Thomas <notifications@github.com<mailto:notifications@github.com>> wrote:


@JIMENOFONSECA<https://github.com/jimenofonseca> sure, I'd like to do that. Also, we should move the settings.py stuff to the cascading configuration files ;-)

daren-thomas:
it's all about tradeoff... the config file proposal i have made is divided into sections, one for each tool. i suggest we see how well that plays out and then split the files up if they end up becoming too complicated.

JIMENOFONSECA:
One question, do you expect this config file to be edited from the GUI of each tool in arc-gis? Or what interface will be used?



On 16 Jun 2017, at 16:28, Daren Thomas <notifications@github.com<mailto:notifications@github.com>> wrote:


it's all about tradeoff... the config file proposal i have made is divided into sections, one for each tool. i suggest we see how well that plays out and then split the files up if they end up becoming too complicated.

daren-thomas:
I'm still figuring that out, Jimeno. what do you think: could editing by hand work?

JIMENOFONSECA:
by hand it would be difficult, especially for new users.

I imagine including these parameters inside the ArcGIS GUI of each tool and creating a separate file storing default values in the main root folder of each tool. Example of the settings.py file in cea/radiation/radiation_daysim

This might make it accessible for 1. the user of arcgis. and 2. the developers (who only see cea and the rootfolder from pycharm).

In essence I see the CEA toolbox working as independent tools, each one with some important variables to consider and worth manipulating for research purposes.

what do you think?


________________________________

JIMENOFONSECA:
@daren-thomas The Solar Radiation tool should point to the file: cea/resources/radiation_daysim/radiation_main. 

It should require as inputs:
- the scenario
- the weather file
[...]

PR #791, opened 2017-09-05
684 launch tool urban solar radiation

daren-thomas:
implements #684 (Launch Tool: Urban Solar radiation)


JIMENOFONSECA:
@shanshanhsieh please kindly check if this works as intended

PR #798, opened 2017-09-26
I724 daylight saving

JIMENOFONSECA:
This PR cleans up the code so daylight saving hours are taking into account


cea/resources/radiation_arcgis/radiation.py
```diff
@@ -5,6 +5,8 @@
 
 import datetime
 import os
+from datetime import timedelta
+from datetime import datetime as datetime2
```
daren-thomas: this seems ugly... why not just use `datetime.timedelta` and `datetime.datetime`? a little more typing, but tacking on a `2` at the end of a variable feels weird to me...

cea/resources/radiation_arcgis/calculate_sunny_hours_of_year.py
```diff
@@ -54,16 +54,11 @@ def calculate_sunny_hours_of_day(day, sunrise, temporary_folder):
...
-    if 90 <= day < 300:
-        D = 1
-    else:
-        D = 0
     # Calculation of Sunrise time
     Sunrise_time = sunrise[day - 1]
     # Calculation of table
     for x in range(values):
-        Hour = int(Sunrise_time) + int(D) + int(x)
+        Hour = int(Sunrise_time) + int(x)
```
daren-thomas: While not really a problem, it is slightly faster to not do the casting (`int(Sunrise_time)`):
```
In [4]: s = 3

In [5]: %%timeit
   ...: for x in range(1000):
   ...:    hour = s + x
   ...:
10000 loops, best of 3: 43.7 µs per loop

In [6]: %%timeit
   ...: for x in range(1000):
   ...:    hour = int(s) + int(x)
   ...:
1000 loops, best of 3: 246 µs per loop
```

cea/resources/radiation_arcgis/radiation.py
```diff
@@ -366,6 +368,10 @@ def calculate_sunrise(year_to_simulate, longitude, latitude):
     print('complete calculating sunrise')
     return sunrise
 
+def is_dst(zonename):
```
daren-thomas: this function is not used anywhere...

daren-thomas: also, it doesn't have a comment? i'm just going to assume it returns a boolean, `True`, if `zonename` is wait... what?

cea/resources/radiation_arcgis/calculate_sunny_hours_of_year.py
JIMENOFONSECA: it is needed, so better we keep it.

cea/resources/radiation_arcgis/radiation.py
JIMENOFONSECA: done

cea/resources/radiation_arcgis/radiation.py
JIMENOFONSECA: done

JIMENOFONSECA:
@daren-thomas can we merge this?

PR #834, opened 2017-10-04
I832 minisplit

JIMENOFONSECA:


daren-thomas:
Build finished.

BhargavaKrishnaSreepathi:
Hey @JIMENOFONSECA, I did `cea test` in anaconda prompt and all the tests passed. 
Following this, I ran the script `cea\demand\preprocessing\properties.py`. Following this I ran `demand_main.py` file. All these ran without any error. 
I also repeated the above step for reference-case-zug. It ran without any error.
If there are no other tests to be done apart from the above, it is ready to be merged.

PR #842, opened 2017-10-06
I840 dx minisplit

JIMENOFONSECA:
This adds the funcitonality of direct expansion and cooling tower insingle buildigns


JIMENOFONSECA:
@BhargavaKrishnaSreepathi please check if this passes your computer, so we can merge

BhargavaKrishnaSreepathi:
@JIMENOFONSECA: the `demand_main.py` file runs seamlessly for both `reference-case-open` and `reference-case-ecocampus`. But for some reason `cea test` was taking a really long time with no progress.

Issue #833, opened 2017-10-04
Enable swtich of databases Singapore and zurich


JIMENOFONSECA:
In essence we are finding great difficulty on easily running simulations considering the databases of Singapore, instead those of Zurich. Until now we have to change the input locator and global variables all the time. which is inefficient and a gigantic source of error.

This issue will require in essence to:
1. add to the data helper interface in arcgis a new input field pointing to  which database to use. two options for now: European or Asian.
2. add to the dynamic demand main in arcigs with the exact same possibility. 
3. add to the configuration file the option of selecting one or the other databases and attach this to the input locator.
4. Make possible that we make the change from pycharm easily without need to go to comand line or arcgis.

@daren-thomas this is kind of an urgent issue which is blocking #713 . could we talk on the phone anytime soon to fix details?

JIMENOFONSECA:
I created the database structure in the branch i833-switch-databases.I leave it there for you to use

These are the pointers of the input locator to account for.
    def get_archetypes_properties(self):
    def get_archetypes_schedules(self):
    def get_supply_systems_cost(self):
    def get_life_cycle_inventory_supply_systems(self):
    def get_life_cycle_inventory_building_systems(self):
    def get_data_benchmark(self):
    def get_supply_systems_database(self):

Nice continuation! thanks @daren-thomas let me know if there is any doubt.

PR #856, opened 2017-10-10
836 multiprocessing not working in demand main

daren-thomas:
multiprocessing should now work in demand_main as defined in the configuration script.


daren-thomas:
@JIMENOFONSECA can you check this PR out please? It solves the issue of multiprocessing not working when running the demand_main.py script from PyCharm.

To configure your system, add a file called `cea.config` to your home directory (`%USERPROFILE%`, normally something like `C:\Users\USERNAME` but IT might have changed that for you).

The important content of that file is this section:

```
[general]
default-scenario: C:\reference-case-open\baseline
weather: Zug
multiprocessing = true
```

(but feel free to copy anything you want from the `cea\default.config` file and overwrite the default values as you please)

The `weather` setting can either be the name of an `*.epw` file in the cea database or a full path to a weather file to use.

After the first run, you will find a `scenario.config` file in the scenario - you can use that to change the values local to the scenario.

JIMENOFONSECA:
@daren-thomas it does not pass the first iteration from PyCharm

C:\Users\Jimeno\AppData\Local\conda\conda\envs\cea\python.exe C:/Users/Jimeno/Documents/CityEnergyAnalyst/cea/demand/demand_main.py
failed to import from rc_model_sia_cc.pyd, falling back to pure python functions
failed to import from storagetank_cc.pyd, falling back to pure python functions
  File "C:/Users/Jimeno/Documents/CityEnergyAnalyst/cea/demand/demand_main.py", line 164, in <module>
    multiprocessing=config.multiprocessing)
  File "C:/Users/Jimeno/Documents/CityEnergyAnalyst/cea/demand/demand_main.py", line 136, in run_as_script
    assert os.path.exists(scenario_path), 'Scenario not found: %s' % scenario_path
AssertionError: Scenario not found: c:\users\jimeno\appdata\local\temp\reference-case-open\baseline
failed to save configuration file to default scenario.

Process finished with exit code 1


@daren-thomas could you also add to the docs a guide on how to work with the config file in this branch? I see this as the first step in interacting with CEA from now on. It is important to avoid confusion with all the dev team after this PR is merged

daren-thomas:
@JIMENOFONSECA sure! it needs documentation!

It looks like you have not configured the `default-scenario` option in your user-config. I'm going to push a change that copies the default configuration to your home folder - that way it will be easier to edit.

daren-thomas:
@JIMENOFONSECA check out the added bits to the installation-guide. Is that enough? I think we should also create a whole document on the configuration file...

JIMENOFONSECA:
@daren-thomas it is a good starting.  But as you said, there should be a whole document on the configuration file to avoid confusion. Also, we need to clear out how to add new settings variables to the config file on our own, otherwise we will keep creating these settings files all the time.

Would you mind adding it to accept this PR? thanks?

daren-thomas:
@JIMENOFONSECA this PR was a hotfix because you said you were stuck because of a bug. Issues #858 and #833 address a wider scope and will change any documentation about the configuration files.

Did you see the change to the installation guide in this PR? Is that enough? Otherwise, it will take about two weeks before I'm ready for a PR with the documentation.

PR #859, opened 2017-10-12
I838 network ecocampus

JIMENOFONSECA:


shanshanhsieh:
@JIMENOFONSECA 
I could run `substations_location.py` successfully.
There are two other errors I encountered:
1. Couldn't run `connectivity potential.py` (btw, is there a reason why there is a space but "_" between the names?)
image

2. `nodes_buildings.shp` , `nodes.shp`, and `streets.shp` do not have the same coordinate system.
This leads to the error in `minimum_spanning_tree.py` since the nodes in `nodes_buildings.shp` and `nodes.shp` are not matching.

JIMENOFONSECA:
So an update: the results for Ecocampus are stored in the private folder. I also cleaned up the results of the simulation to erase nodes from the minimum spanning tree. Project wise. we canuse those results to run the network simulation

shanshanhsieh:
@JIMENOFONSECA This is the error i got from running `thermal_network_matrix.py` with the shape files you provided.
It got stuck at the `gpd.read_file()` function at line 1329. 
image

JIMENOFONSECA:
@shanshanhsieh no idea about the error, which branch are you working on?

shanshanhsieh:
i am running `thermal_network_matrix.py` from the master
maybe it has something to do with the ISO 88591 character set, which was added to the `node.cpg`, 'edges.cpg`, `nodes_building.cpg` in the latest commit?

JIMENOFONSECA:
I checked for a solution online,

The best is to erase those two .cpg files created once we modify the nodes and edges in arcgis, now it works!. the next problem is that the index of the shape file needs to be re indexed once you read it int a dataframe

PR #870, opened 2017-10-19
I839 simulate cooling network for ecocampus

shanshanhsieh:
update `thermal_network_matrix.py`
- adding check functions to make sure the shape files has no duplicated nodes/edges, and no nodes at dead ends
- adjust precision of coordinates to match building substations with nodes

the simulated results are updated in `cea-reference-case/reference-case-ecocampus/.../output/optimization/network`

To test:
1. run `thermal_network_matrix.py` for reference-case-open for 8760 time-steps
2. run `thermal_network_matrix.py` for ecocampus (maybe with reduced timestep e.g. 240, it can be done by changing the ranges at line 157 and 501)


BhargavaKrishnaSreepathi:
@shanshanhsieh : done with the testing. And both of the tests specified ran through.

PR #877, opened 2017-10-25
bufgix calculation of cooling heat pump electricity

gabriel-happle:
- if statement in thermal_loads.py was looking up values of supply
systems instead of hvac systems
- logic of if statement (hard coded 'T'-values) moved to
control_heating_cooling_systems.py
- documentation


gabriel-happle:
Please review this small bugfix @JIMENOFONSECA .
It is blocking the work of @BhargavaKrishnaSreepathi on the EcoCampus case study.

PR #882, opened 2017-11-06
changes to incorporate geothermal potential calculations

BhargavaKrishnaSreepathi:
This PR fixes the geothermal calculations, to make it dependent on the supply_systems.dbf files instead of geothermal.csv files


cea/optimization/preprocessing/decentralized_buildings.py
```diff
@@ -31,7 +34,10 @@ def decentralized_main(locator, building_names, gv):
     :rtype: Nonetype
     """
     t0 = time.clock()
-    geothermal_potential = pd.read_csv(locator.get_geothermal_potential(), index_col="Name")
+    prop_geometry = Gdf.from_file(locator.get_zone_geometry())
+    geothermal_potential_data = dbfreader.dbf_to_dataframe(locator.get_building_supply())
+    geothermal_potential = geothermal_potential_data[['Name', 'restr_geo']]
+    geothermal_potential['Area_geo'] = geothermal_potential['restr_geo'] * prop_geometry.area
```
JIMENOFONSECA: This will be a source of error, better use merge on=Name. Otherwise it might happen that the order of the cells be shifted.

cea/optimization/preprocessing/decentralized_buildings.py
```diff
@@ -246,10 +252,9 @@ def calc_new_load(mdot, TsupDH, Tret, gv):
         # Check the GHP area constraint
         for i in range(10):
             QGHP = (1-i/10) * Qnom
-            areaAvail = geothermal_potential.ix[building_name,"Area_geo"]
+            areaAvail = geothermal_potential.ix[geothermal_potential['Name'] == building_name, 'Area_geo']
```
JIMENOFONSECA: I suggest setting the index of the geothermal_potential to "Name" and then use the formulation geothermal_potnetial.ix[building_name, "Area_geo].

BhargavaKrishnaSreepathi:
The reviewer comments were addressed in the new commit. @JIMENOFONSECA, you can take a look at it

cea/optimization/preprocessing/decentralized_buildings.py
```diff
@@ -246,9 +253,9 @@ def calc_new_load(mdot, TsupDH, Tret, gv):
         # Check the GHP area constraint
         for i in range(10):
             QGHP = (1-i/10) * Qnom
-            areaAvail = geothermal_potential.ix[building_name,"Area_geo"]
+            geothermal_potential = geothermal_potential_data.set_index('Name')
```
JIMENOFONSECA: great work at fixing this. However, better do this line out of the loop. It is necessary only once.

BhargavaKrishnaSreepathi: Aaah..i am an idiot. Fixed it in the new commit. Thanks

PR #892, opened 2017-11-14
Calculate radiation total from insolation files

BhargavaKrishnaSreepathi:
This PR calculates the total radiation of a building and categorizes them into the `radiation_roof` and `radiation_walls`. For both cases, the surfaces generated by day_sim corresponding to `roof` and `wall` are aggregated and the radiation in Whperm2 is calculated and saved. This can further be used to visualize the radiation in an area.

Also, few corrections corresponding to the new config file have been incorporated.

The way to test this is, to run `radiation_total.py` and then compare the results with the solar insolation files of the corresponding building. This can be done by using `cea\resources\radiation_daysim\plot_points.py`, and by looking at the plot


JIMENOFONSECA:
I wonder if we need to get the wighted average of each building fro Wh/m2 instead of the sum. The sum will be big for big buildings which is rather trivial. I guess it is better to get the weighted average. Wh/m2 per m2 of surface of the building

daren-thomas:
Build finished.

BhargavaKrishnaSreepathi:
@JIMENOFONSECA: a small confirmation, the `B001_insolation_Whm2` files provide the surface wise radiation. What are the units of this? is it Wh per m2 of the surface?

JIMENOFONSECA:
w/m2, BUT every single surface has different w/m2. AND every single surface has a different area. 
hence you need to do a weighted average.

BhargavaKrishnaSreepathi:
ok ok. So basically convert into W/m2 of each surface to W (multiplying by corresponding area) and then add it up based on Surface type and then divide by the Surface area. Got it.

BhargavaKrishnaSreepathi:
@JIMENOFONSECA : You can take a look at the PR now. Weighted average calculations corresponding to roof and walls have been implemented in the new commit

JIMENOFONSECA:
cool, do you have a map to show the results  per building so it easy to evaluate? thanks!
On 14 Nov 2017, at 3:45 pm, Bhargava Krishna Sreepathi <notifications@github.com<mailto:notifications@github.com>> wrote:


@JIMENOFONSECA<https://github.com/jimenofonseca> : You can take a look at the PR now. Weighted average calculations corresponding to roof and walls have been implemented in the new commit

daren-thomas:
Build finished.

BhargavaKrishnaSreepathi:
Added the pictures to `cea\plots\4D_plots_ArcGIS`. Take a look at let me know.

daren-thomas:
Build finished.

JIMENOFONSECA:
so the results for the roof does not look ok to me, any reasons why the roofs are not similar? is the code doing the wighted average right" (sum from i to n of (isolation[i] * surface area[i]) / (sum from i to n of (surface area[i]))

BhargavaKrishnaSreepathi:
@JIMENOFONSECA: Yes the code does that calculation. But for roof scenario, most of the buildings have only one surface with `SURFACE TYPE` as roof. So effectively it is showing the values generated from day_sim. I just manually counter checked, the values from the `building_insolation` files are very different for the same time.

PR #924, opened 2017-11-30
Create the release v2.6

daren-thomas:


JIMENOFONSECA:
@daren-thomas this is good to go. is there anything else we have to add to the release?

daren-thomas:
@JIMENOFONSECA yes, I believe we should fix the ArcGIS interface first, before creating a release. I am working on it and expect results today!

PR #925, opened 2017-12-01
20171201 bugfix in master veolia

daren-thomas:
creating an installation for veolia and found some bugs that needed fixing!

(this is URGENT)

Also, it fixes a bug in master (the tests weren't running through because of a change to the zurich reference-case (difference in buildings listed in scenario and test radiation data)


JIMENOFONSECA:
@daren-thomas this branch is bleeding form commits in other branches and it is difficult to review + it includes commits from other authors. would you mind creating a "clean PR" with only the commits of this issue? thanks!.

daren-thomas:
ah. wait. i see the problem: this isn't based off master... hm... OK. I will create a new branch and pull request.

Issue #930, opened 2017-12-06
`MonthlyDemandWriter` broken for sensitivity analysis


daren-thomas:
The constructor (`__init__`) of `MonthlyDemandWriter` was updated in commit [09ea089](https://github.com/architecture-building-systems/CityEnergyAnalyst/commit/09ea0893fe46b440ad881e5a6c7082f074bdcd93) without checking for usage in the rest of the project. The sensitivity analysis now generates the following error:

```
Running demand simulation for sample 0
Traceback (most recent call last):
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\20171206cea\Scripts\cea-script.py", line 11, in <module>
    load_entry_point('cityenergyanalyst==2.6a1', 'console_scripts', 'cea')()
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\20171206cea\lib\site-packages\cea\interfaces\cli\cli.py", line 34, in main
    script_module.main(config)
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\20171206cea\lib\site-packages\cea\analysis\sensitivity\sensitivity_demand_simulate.py", line 212, in main
    output_parameters=config.sensitivity_demand.output_parameters)
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\20171206cea\lib\site-packages\cea\analysis\sensitivity\sensitivity_demand_simulate.py", line 173, in simulate_demand_batch
    totals, time_series = simulate_demand_sample(locator, weather, output_parameters)
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\20171206cea\lib\site-packages\cea\analysis\sensitivity\sensitivity_demand_simulate.py", line 113, in simulate_demand_sample
    gv.demand_writer = cea.demand.demand_writers.MonthlyDemandWriter(gv)
TypeError: __init__() takes exactly 4 arguments (2 given)
```

The other `DemandWriter` implementations should also be checked.

PR #932, opened 2017-12-06
20171214 bugfixes for master

daren-thomas:
Candidate for a release. Please check all scripts. Sensitivity Analysis doesn't work as per issue #930.

@JIMENOFONSECA Could you please have a look as I need a stable version to sent to veolia preferably tomorrow.


cea/default.config
```diff
@@ -150,47 +152,58 @@ year.help = The year to base the calculations on. WARNING!, make sure you are us
...
 format-output.help = format of output files. csv is the default, HDF5 is used when data needs to be compressed.
+format-output.category = Advanced
 
 use-dynamic-infiltration-calculation = false
 use-dynamic-infiltration-calculation.type = BooleanParameter
 use-dynamic-infiltration-calculation.help = Use the dynamic infiltration calculation (slower!)
+use-dynamic-infiltration-calculation.category = Advanced
 
-use-daysim-radiation = true
+use-daysim-radiation = false
```
JIMENOFONSECA: this should be true

environment.yml
```diff
@@ -23,11 +23,12 @@ dependencies:
...
 - freexl=1.0.2=vc9_2
 - functools32=3.2.3.2=py27_1
-- gdal
+- geos
+- gdal=2.1.3=py27_vc9_7
 - geopandas>=0.2.1
-- geos>=3.5.1
 - hdf4=4.2.12=vc9_0
 - hdf5>=1.8.18
+- h5py
```
JIMENOFONSECA: I am not sure if we need this. I am using pandas.to_hdf and pandas.read_hdf to use hdf5

cea/default.config
daren-thomas: you're right... i was setting this to false for veolia - i'll just have to also explain how to install daysim. no problem.

environment.yml
daren-thomas: We don't need it specifically... it seems to be included in one of our dependencies. I'm removing it from the list.

daren-thomas:
@JIMENOFONSECA i have made the changes you requested and also run the tests. I think we're good to go. Next step: Integrate the sensitivity tool stuff...

PR #946, opened 2017-12-19
945 modify dynamic supply systems interface

### LINKED ISSUE #945: Modify dynamic supply systems interface
The interfaces of `Photovoltaic Panels`, `Solar Collector Panels`, and `PVT Panels` are running smoothly.
But there are some redundant parameters in the individual panes.
Modifications required are shown in the pictures below:
1. For Photovoltaic Panels
picture1
2. For Solar Collector Panels
picture2
3. For PVT panels
picture3

daren-thomas:
fixes #945

### DIFF
--- a/cea/default.config
+++ b/cea/default.config
@@ -224,9 +224,11 @@ solar-window-solstice.help = Desired hours of solar window on the solstice
 
 t-in-sc = 75
 t-in-sc.type = RealParameter
+t-in-sc.help = Inlet temperature of solar collectors [C]
 
 t-in-pvt = 35
 t-in-pvt.type = RealParameter
+t-in-pvt.help = Inlet temperature of PVT panels [C]
 
 dpl = 200
 dpl.type = RealParameter
--- a/cea/default.config
+++ b/cea/default.config
@@ -222,33 +222,31 @@ solar-window-solstice = 4
 solar-window-solstice.type = IntegerParameter
 solar-window-solstice.help = Desired hours of solar window on the solstice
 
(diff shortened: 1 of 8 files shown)


cea/default.config
```diff
@@ -194,62 +194,61 @@ year.help = the year to calculate radiation for
...
+type-pvpanel.choices = PV1 PV2 PV3
+type-pvpanel.help = Type of PV panel
 
-type-SCpanel = SC1
-type-SCpanel.type = ChoiceParameter
-type-SCpanel.choices = SC1 SC2
-type-SCpanel.help = type of solar collector, SC1: flat plat collectors, SC2: evacuated tubes
+type-scpanel = SC1
+type-scpanel.type = ChoiceParameter
+type-scpanel.choices = SC1 SC2
```
shanshanhsieh: @daren-thomas  I have a question on how can users add their customized technologies? For example, adding a `SC3` in the `region/economics/supply_systems.xls` database. 
Is there a way to go check on what is in the database, and then form the dropdown choices?

shanshanhsieh:
@daren-thomas  thanks for the changes!
This is how my interface looks like after running `install-toolbox`. Seems like there's an error on the DemandTool.
image

cea/default.config
daren-thomas: Currently, no, we don't have a way to do that... we could think up a way, though by making a class called "ExcelLookupChoiceParameter" or something. Not quite sure if we need it. But go ahead and create an issue for this :-)

PR #947, opened 2017-12-19
fixes #353

### LINKED ISSUE #353: Update installation guide for new radiation script

daren-thomas:


JIMENOFONSECA:
@daren-thomas so it means that we do not need anymore the LFS?

daren-thomas:
i think so... @JIMENOFONSECA i'm not sure. how would we test this? i'll check - in a separate issue.

JIMENOFONSECA:
@daren-thomas I am confused, why does this PR then erases the line in the documentation that states that LFS is needed

daren-thomas:
@JIMENOFONSECA I just checked: we don't need git lfs anymore. We used to have radiation data > 100MB stored in the repo, but we no longer do that. LFS is not needed to clone the repository and it is not needed to run the CEA (as tested yesterday with the ubuntu branch)

PR #953, opened 2018-01-08
steps to avoid Nan's and 0K temperature values

BhargavaKrishnaSreepathi:
While running optimization for a higher number of generations, there is a chance where the COP becomes negative and which in turn leads to negative costs. This was being caused because of the default values of temperatures in networks (in Kelvin) being '0'. Now I changed it to '273' (which is the default 0 in degree C). This fixed the issue of having negative costs and emissions. Also, while I was at it, fixed few loops where the denominator can become '0' and thus raising Nan's. These observations were not found prior because they occur once or twice in 1000 samples.

### DIFF
--- a/cea/optimization/master/summarize_network.py
+++ b/cea/optimization/master/summarize_network.py
@@ -202,7 +202,7 @@ def calc_temp_withlosses(t0, Q, m, cp, case):
         else:
             t1 = t0 - Q / (m * cp)
     else:
-        t1 = 0
+        t1 = 273
     return t1
 
 def calc_return_temp(sum_t_m, sum_m):
@@ -219,7 +219,7 @@ def calc_return_temp(sum_t_m, sum_m):
     if sum_m > 0:
         tr = sum_t_m / sum_m
     else:
-        tr = 0
+        tr = 273
     return tr
 
 
@@ -246,7 +246,7 @@ def calc_supply_temp(tr, Q, m, cp, case):
         else:
             ts = tr - Q / (m * cp)
     else:
-        ts = 0
+        ts = 273
     return ts
 
 #============================
--- a/cea/optimization/master/summarize_network.py
+++ b/cea/optimization/master/summarize_network.py
@@ -179,31 +179,31 @@ def network_main(locator, total_demand, building_names, config, gv, key):
 # Supply and return temperatures
 # ============================
 
-def calc_temp_withlosses(t0, Q, m, cp, case):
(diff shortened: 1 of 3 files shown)


cea/optimization/master/summarize_network.py
```diff
@@ -202,7 +202,7 @@ def calc_temp_withlosses(t0, Q, m, cp, case):
         else:
             t1 = t0 - Q / (m * cp)
     else:
-        t1 = 0
+        t1 = 273
```
daren-thomas: can we use a constant here? (also, wasn't it more like 273.15?) `ZERO_DEGREES_CELSIUS_IN_KELVIN`?

cea/optimization/slave/least_cost.py
```diff
@@ -190,7 +190,7 @@ def source_activator(Q_therm_req_W, hour, context):
                     E_coldsource_HPSew_W = float(Q_HPSew_cold_primary_W)
 
                 if (
-                MS_Var.GHP_on) == 1 and hour >= MS_Var.GHP_SEASON_ON and hour <= MS_Var.GHP_SEASON_OFF and Q_therm_req_W > 0:
+                MS_Var.GHP_on) == 1 and hour >= MS_Var.GHP_SEASON_ON and hour <= MS_Var.GHP_SEASON_OFF and Q_therm_req_W > 0 and tdhsup_K != tdhret_req_K:
```
daren-thomas: when comparing floating point numbers, it is often unwise to use `!=` and `==`, as they could be very close but still not the same due to how floating point numbers are represented by the computer. You could use `not np.isclose(tdhsup_K, tdhret_req_K)`

cea/optimization/slave/least_cost.py
```diff
@@ -219,7 +219,7 @@ def source_activator(Q_therm_req_W, hour, context):
                     E_GHP_req_W = E_GHP_req_W
                     E_coldsource_GHP_W = Q_GHP_cold_primary_W
 
-                if (MS_Var.HP_Lake_on) == 1 and Q_therm_req_W > 0 and HPLake_allowed == 1:  # run Heat Pump Lake
+                if (MS_Var.HP_Lake_on) == 1 and Q_therm_req_W > 0 and HPLake_allowed == 1 and tdhsup_K != tdhret_req_K:  # run Heat Pump Lake
```
daren-thomas: see comments above on comparing floating point numbers.

PR #954, opened 2018-01-08
904 config file editor

daren-thomas:
Implements a first stab at a config file editor for the user configuration file.

How to test:

- (you might need to `conda env update` if you get errors about packages missing)
- also do `pip install -e .[dev]`
- run `cea config-editor`

You should see a GUI that let's you edit the options of each section in the config file. Clicking the `Save` button saves the current section to disk.


JIMENOFONSECA:
@daren-thomas I followed the guideline. It seems the module htmlpy is missing

image

daren-thomas:
@JIMENOFONSECA sorry. Fixed.

JIMENOFONSECA:
@daren-thomas it is awesome. I wonder if there should be bottom that says "restore defaults" in every tab.  I got somehow lost in the naming of every tab and sometimes I am not sure to which tool the tab refers to. What is your opinion? Should the tab have the same name as the tool in ArcGIS?

JIMENOFONSECA:
@daren-thomas I have one quesiton is it possible to run it from pycharm?

daren-thomas:
@JIMENOFONSECA good idea ("the restore defaults button")! I added an issue: #957

Regarding starting from PyCharm: Sure. Just open the file (cea/interfaces/config_editor/config_editor.py) - BUT: once this is merged, it is also possible to create an icon in the task bar (or windows menu) that launches the tool... I think we might also be using this technology to wrap the timeseries graphs and other visualizations (see your issue on the dashboard) so we might end up with a collection of apps to run...

Regarding tab names - these are the names of the sections as per the config files. Some scripts share a section. There is no one-to-one mapping. I'm open for suggestions.

Issue #951, opened 2017-12-21
chisqprob is deprecated


daren-thomas:
According to this page: https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.stats.chisqprob.html, the package `chisqprob` is deprecated and we should use `stats.distributions.chi2.sf` instead - I haven't been able to find the documentation online for this yet, but it stands to reason that we need to update our code.

We ran into this problem when upgrading to Ubuntu.

```
$ cea test
Traceback (most recent call last):
  File "/home/jose/anaconda2/envs/cea/lib/python2.7/site-packages/doit/doit_cmd.py", line 168, in run
    return command.parse_execute(args)
  File "/home/jose/anaconda2/envs/cea/lib/python2.7/site-packages/doit/cmd_base.py", line 122, in parse_execute
    return self.execute(params, args)
  File "/home/jose/anaconda2/envs/cea/lib/python2.7/site-packages/doit/cmd_base.py", line 380, in execute
    self, params, args)
  File "/home/jose/anaconda2/envs/cea/lib/python2.7/site-packages/doit/cmd_base.py", line 263, in load_tasks
    return self._load_from(cmd, self.mod_dict, self.cmd_names)
  File "/home/jose/anaconda2/envs/cea/lib/python2.7/site-packages/doit/cmd_base.py", line 247, in _load_from
    task_list = loader.load_tasks(members, cmd_list, cmd.execute_tasks)
  File "/home/jose/anaconda2/envs/cea/lib/python2.7/site-packages/doit/loader.py", line 148, in load_tasks
    _process_gen()
  File "/home/jose/anaconda2/envs/cea/lib/python2.7/site-packages/doit/loader.py", line 140, in _process_gen
    task_list.extend(generate_tasks(name, ref(), ref.__doc__))
  File "/home/jose/anaconda2/envs/cea/lib/python2.7/site-packages/doit/loader.py", line 314, in generate_tasks
    for task_dict, x_doc in flat_generator(gen_result, gen_doc):
  File "/home/jose/anaconda2/envs/cea/lib/python2.7/site-packages/doit/loader.py", line 28, in flat_generator
    for item in gen:
  File "/home/jose/Documents/CityEnergyAnalyst/cea/tests/dodo.py", line 136, in task_run_data_helper
    import cea.demand.preprocessing.data_helper
[...]

PR #959, opened 2018-01-10
951 chisqprob is deprecated

daren-thomas:
Solves the issue #951 - CEA can be installed on Ubuntu 16 by the steps in the docs/installation-guide-ubuntu.rst file.


BhargavaKrishnaSreepathi:
I was able to install CEA in Ubuntu by following the instructions. Only comment is that it takes really long time to get the installation done, it took around 1.5 hours to get it installed. But the bright side is, it ran through without any errors.

PR #981, opened 2018-01-24
I 975 detailed heat balance

gabriel-happle:
- heat balance graph for buildings
- auxiliary calculations for heat balance
- refactoring
- moving stuff out of globalvar


gabriel-happle:
@JIMENOFONSECA probably we have to update the tests because of the solar radiation balance fix
I will try to run the tests now

JIMENOFONSECA:
@gabriel-happle the last PR included a new case study with all the solar calculations, so please use this one for the tests

gabriel-happle:
@JIMENOFONSECA 
based on only comparing the results of the solar heat gain changes,
changes in heating and cooling energy are very significant!
up to + 30% of heating
up to - 15% of cooling
I think it makes sense, since we counted the re-radiation losses as gains.

however, The dodo tells me that cooling is higher than before. So I don't know what numbers are used there.
should I generate new test data?

PR #984, opened 2018-01-25
I968 opt graphs

JIMENOFONSECA:
@BhargavaKrishnaSreepathi you need to have the optimization ready to run with this PR. Please run plot.optimization.dashboard_district

There should be three graphs created in 

/baseline/outputs/plots/timeseries.


BhargavaKrishnaSreepathi:
Few comments:

1. IN the 'District_Pareto_curve_performance', the fit does not make much sense, as it is making a 2D fit on a 3D curve. 
2. In the 'District_Pareto_curve_over_generations', the axis need to be fixed. Few points are right on the axis, it would be better if there is a gap between the axis and the points. This can be done by increasing the axis range
3. In the 'District_Pareto_curve_capacity_installed', the x-axis is not going from ind 0 to ind 2, but is going like this ind 1, ind 0, ind 2. I assume this might because of the descending order of Power Capacity. I suggest this be as per the individuals order instead of the power capacity

JIMENOFONSECA:
@BhargavaKrishnaSreepathi thanks for the comments.

I made the changes for 1 and 2. but I kept the fit line to show how is the trend in x and Y, I find this useful when there are hundreds of individuals. I changed the name to "fit X vs Y" so it isclear that it is 2D.

As for comment 3. it thought we wanted to organize information from maximum to minimum, which can be helpful when we have several individuals. Should we keep it like that?

BhargavaKrishnaSreepathi:
So for the maximum to minimum organization, the logic looks good, but I was thinking that we  uncheck few technologies then the individuals change, but I just checked and it does not change. So the system is good, no change required.

Apart from this, I have no comments and thanks for doing a good work 👍

JIMENOFONSECA:
@BhargavaKrishnaSreepathi cool thanks, could you click in 'add review' on t he top of the page and aprove so it can be merged? thanks

PR #990, opened 2018-01-25
colors added as a class

JIMENOFONSECA:
@shanshanhsieh this adds the colors to new variables, would you mind adding a review?


shanshanhsieh:
@JIMENOFONSECA  i tried to run several plots on demand and solar potential.
The scatter plots work fine, but the colors for bar charts are all black.
Is it the same if you run on your computer?
image

PR #991, opened 2018-01-25
966 try plotly dash

daren-thomas:
this shows two plotly dash plots (load_curve and load_duration_curve).

Try it out: `pip install dash dash-renderer dash-html-components dash-core-components`

Also: `pip install plotly --upgrade` (maybe not necessary)

And then run the `cea/interfaces/dashboard/dashboard.py` file in PyCharm. Click on the link in the output to view the charts in the browser.


JIMENOFONSECA:
It also needs cefpython3 to run

daren-thomas:
@JIMENOFONSECA Maybe you got the wrong branch: there are two 966 branches - the one this pull request is about is the 966-try-plotly-dash branch. The other branch that went the cefpython3 route turned out to be a dead end (cefpython3 has some bugs in it and i don't think the effort to fix them is worth it)

JIMENOFONSECA:
Yep we did the wrong one.

JIMENOFONSECA:
@daren-thomas so should we let this open? and see who do we get to do it further?

daren-thomas:
@JIMENOFONSECA I think so? We can also close the PR... The branch can stay open indefinitively...

PR #993, opened 2018-01-26
I987 cea workflow

JIMENOFONSECA:
@daren-thomas I am having some errors while compiling the documentation. I can not solve them, would you mind giving me a hand?


JIMENOFONSECA:
@BhargavaKrishnaSreepathi to check the workflow:
- go to cea/docs from the command line
- do make html
- open cea/docs/build/getting-started.html

daren-thomas:
@JIMENOFONSECA sure I can help. First step: 

```
WARNING: [...]\docs\installation-guide-ubuntu.rst:9: (WARNING/2) Title underline too short.
```

That refers to the line underneath a title - it has to be exactly the same length as the text above it. I found three cases where there was either a character too much or one too few. I posted a commit that solves this.

daren-thomas:
The next error "WARNING: Enumerated list ends without a blank line; unexpected unindent." refers to a badly formatted enumerated list. Basically, if a list item spans multiple lines, the successive lines need to be indented.

I'm going to go error by error, posting the resolution here as a sort of mini-tutorial - you can check the specific commits to see how the error was resolved.

daren-thomas:
The next error "WARNING: Unexpected indentation" is just that. In one case, a field list was indented on the next line. This is just not sound reStructuredText. And in the other case, the citation needs to be it's own paragraph (have a blank line above and below).

daren-thomas:
"ImportError: cannot import name optimal_angle_and_tilt": This looks like a bug in the code - PyCharm can't resolve those names either. Maybe something was refactored away from there? Is this file still being used? I suggest we just remove it from the codebase.

daren-thomas:
"WARNING: Field list ends without a blank line; unexpected unindent." Yes. The `:param xyz: ...`, `:type xyz: ...` and `:return: ...` lines are syntactically "field lists". See here: http://docutils.sourceforge.net/docs/user/rst/quickref.html#field-lists

The problem was, that the second line of the `:return: ...` field list has to be indented. Then, the whole field list (all such lines) need to be separated from the rest of the text by a blank line, if anything is following. That's what the error message refers to.

I have fixed the ones I could fix - but I'm not sure what to do about the `photovoltaic_arcgis.py` file. Are we going to keep it? Then we need to fix it. Currently, it is just dead code.

PR #996, opened 2018-01-29
fixing errors in electricity load

### LINKED ISSUE #988: reference case zug not working
This is the error i got from running `data_helper`:
image
it might have something to do with the updates in `cea-reference-case` two days ago by @JIMENOFONSECA , would you mind taking a look at it?
Thanks!

JIMENOFONSECA:
This fixes #988 @shanshanhsieh would you mind testing if the zug case runs now? you need to use the new one from the master branch of the reference case repository.


shanshanhsieh:
i ran `radiation` and `demand` for the new `reference-case-zug`
this is the error i encounter when running `demand`
image

JIMENOFONSECA:
@shanshanhsieh did you tried with this branch? I solved that problem here..

PR #1002, opened 2018-01-30
I1001 economic plots

JIMENOFONSECA:
@roglen   would you mind testing this? thanks!

- run cea>demand>demand_main.py
- run cea>analysis>operation_costs.py
- run cea>plots>life_cycle>district_dashboard.py

there should be a new plot in casestudy.../baseline/outputs/plots


roglen:
In the dashboard_district file I get the following error:

IOError: File /home/energy/Documents/CityEnergyAnalyst/cea/plots/life_cycle/C:/reference-case-open/baseline/outputs/data/demand/Total_demand.csv does not exist

most likely because on my ubuntu macbook I don't have a C: partition so it tries to create one and search for the demand file in there. My scenario path from the config file is: 

/home/energy/Documents/reference-case-open/baseline

hence my demand output is in 

/home/energy/Documents/reference-case-open/baseline/outputs/data/demand/

PR #1004, opened 2018-01-30
i997_thermal_network_matrix_does_not_run_for_DH

roglen:
switched keyword from "is" to "=="


daren-thomas:
Good call! Using `is` for string comparison is an anti-pattern!

daren-thomas:
Before we merge: Do we have a test case for this code path? Can we write a unit test for this?

shanshanhsieh:
i am running the script with the open case.
and yes! i was just thinking about that! @daren-thomas, please give instructions on how to create unit test for this.

PR #1008, opened 2018-01-30
I1007 radiation prerequisite

martin-mosteiro:
Up until now, the demand script looked up ArcGIS radiation files even when Daysim radiation was being used. This PR fixes that.


martin-mosteiro:
Sorry, I found an issue with this pull request. I will create a new one when appropriate.

PR #1009, opened 2018-01-30
I1007 radiation prerequisite

martin-mosteiro:
This pull request solves the issue of `demand_main` looking for ArcGIS radiation files even when `use_daysim_radiation=True`.

The step where the existance of the appropriate files was also changed from an `if not os.path.exists()` statement at the beginning of `demand_calculation(...)` to an assert statement in `main(config)`.


martin-mosteiro:
@JIMENOFONSECA indeed. on it!

PR #1028, opened 2018-02-01
Use appropriate InputLocator to list building names

martin-mosteiro:
Building names in building_properties.py were being listed by using `envelope.index` instead of the appropriate input locator (`locator.get_zone_building_names()`). This created problems in certain cases, so it needed to be fixed.

### DIFF
--- a/cea/demand/building_properties.py
+++ b/cea/demand/building_properties.py
@@ -312,7 +312,7 @@ class BuildingProperties(object):
         envelope['Aroof'] = np.nan
 
         # call all building geometry files in a loop
-        for building_name in envelope.index:
+        for building_name in locator.get_zone_building_names():
             geometry_data = pd.read_csv(locator.get_radiation_metadata(building_name))
             geometry_data_sum = geometry_data.groupby(by='TYPE').sum()
             # do this in case the daysim radiation file did not included window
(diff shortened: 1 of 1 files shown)


cea/demand/building_properties.py
```diff
@@ -312,7 +312,7 @@ def geometry_reader_radiation_daysim(self, locator, envelope, occupancy, geometr
         envelope['Aroof'] = np.nan
 
         # call all building geometry files in a loop
-        for building_name in envelope.index:
+        for building_name in locator.get_zone_building_names():
```
JIMENOFONSECA: Here the best is to create a variable with the names of the buildings once. The thing is that now it is reading the csv file and extracting the names again and again. This adds overhead...

Also, what are the kind of 'weird' results that triggered this PR?

daren-thomas: it looks like in this codepath (creating the `BuildingProperties` object) it only gets called once. chances are, it doesn't get called a lot for the execution of the whole demand script (or similar scripts) - I think optimizing for readability instead of introducing more parameters to pass around the whole script beats a (small) performance penalty. I admit I haven't tested this though. Here is a preliminary test with the reference-case-open:

```
In [3]: %timeit -r 100 -n 10 locator.get_zone_building_names()
10 loops, best of 100: 3.25 ms per loop
```
It doesn't really look like overhead to me.

martin-mosteiro: @JIMENOFONSECA no weird results, but the list of building names should consistently be generated from the same source, in this case the geometry. It could be that you have more entries in the envelope file than necessary, but not in the geometry file. So it makes sense to use a consistent source for the list of building names (which is why the input locator was created).

However, one alternative would be to use `prop_geometry.index` instead of `prop_envelope.index` in order to be consistent. Would this be an appropriate solution? Also in terms of readability, @daren-thomas.

PR #1050, opened 2018-02-13
I1049 life cycle

JIMENOFONSECA:
@martin-mosteiro would you mind checking this PR for me?

How to run?
1. erase config file in userprofile.
2. run Dashboard_district.py located in cea/plots/lca/...

Expected results:
5 graphs should be saved in  outputs/plots

### DIFF
--- a/cea/default.config
+++ b/cea/default.config
@@ -633,14 +633,6 @@ maxtime.type = IntegerParameter
 maxtime.help = maximum allowed time in seconds, 604800 is equivalent of 7 days
 
 [dashboard]
-scenario = C:/reference-case-open/baseline
-scenario.type = PathParameter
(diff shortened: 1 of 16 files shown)


martin-mosteiro:
@JIMENOFONSECA The script work as expected and the graphs look good. Maybe it's worth renaming the labels, though? See for example `District_emissions.html`:
screen shot 2018-02-14 at 11 15 10
The graph would be a lot clearer if the legend said `Mobility`, `Operation` and `Embodied` instead of `M_ghg_ton`, `O_ghg_ton` and `E_ghg_ton`.

JIMENOFONSECA:
@martin-mosteiro good point I will create a separate issue for that. if it is ik would you mind accepting a review so I can merge? thanks!

PR #1051, opened 2018-02-13
fix according to calc_temperature_ground

JIMENOFONSECA:
this fixes #914


gabriel-happle:
Cool!
Looks good.
With this fix I believe we can remove the `seasonhours` parameter from `gv`. Could you please test that and remove it, if the checks pass.

JIMENOFONSECA:
@gabriel-happle thanks :-). unfortunately seson hours is still used by the neural network...

JIMENOFONSECA:
@gabriel-happle if it is ok, would you add a review so we can merge?

JIMENOFONSECA:
@gabriel-happle if it is ok, would you add a review so we can merge?

PR #1064, opened 2018-02-20
1035 excel to shapefile polylines

daren-thomas:
@shizhongming can you check this out? it contains the polyline fix. use the `polygon=False` option to `excel-to-shapefile` when creating a shapefile with polyline information.


shizhongming:
image
I tried but it seems that I am not able to change the polygon from =True to =False.
Grasshopper doesn't accept the argument from the panel, and it also seems to keep resetting the config file to default each time it runs....

daren-thomas:
OK, instructions for testing: We need to do a `cea install-grasshopper` for any changes to the "cea" highlevel api (parameters for scripts changed with the introduction of the "polygon" parameter)

PR #1065, opened 2018-02-20
first stab at creating a test document

daren-thomas:


JIMENOFONSECA:
@daren-thomas good about this. I wonder if we can simplify it a bit in terms of steps to avoid overhead.

I looked at other repository which I found easy to read on how to test it.

https://github.com/thu-ml/zhusuan/blob/master/CONTRIBUTING.md

daren-thomas:
@JIMENOFONSECA I didn't mean to create a pull request just yet - we're still not ready. That was a mistake (just wanted to publish the branch so @amraladdin can start working on it too).

I checked the the repository you mentioned and that is talking about something totally different. Basically, their `python -m unittest discover -v` is equivalent to our `cea test`.

What this pull request is about is not automated unit and integration testing. This is about acceptance testing and the fine polish we want to have for the executive course. This is about dividing up the workload of testing each script manually looking for bugs we haven't found yet. This is about going through the documentation looking for spelling / grammar problems. Checking the labels on the interfaces. Trying out different combinations that might break the scripts ("what happens if this option is used in combination with that option?").

I believe this is an effort we need to make. And I believe we should do it in a documented way, hence this PR, so that next time we create a big release, we can reproduce the testing effort. Also, by documenting it, we might be able to automate some of the manual stuff. There will still always be things that a human needs to look at.

JIMENOFONSECA:
@daren-thomas So I would suggest to make a CONTRIBUTING.md file in CEA which basically states the steps we need to take to make sure the tool runs. is that right? Or would you like to have a thorough version in the docs of CEA?

daren-thomas:
@JIMENOFONSECA that is a separate issue in my mind. Great idea, though! We should do that!

What I'm getting at with this pull request is something different: This is the planning of the tests for the superfinal milestone. So we're all sure of the capabilities for the executive course. I expect everyone needs to chip in some time to do some testing.

JIMENOFONSECA:
@daren-thomas I see. I think then a channel is slack / discussion in an  issue / google doc shared with the parties interested will do better than making this part of the docs. what do you think?

daren-thomas:
@JIMENOFONSECA that could also work.

JIMENOFONSECA:
@daren-thomas so I am closing this as there will be a document somewhere else to run all tests for a major release.

PR #1080, opened 2018-02-27
Fixing excel shapefile

JIMENOFONSECA:


JIMENOFONSECA:
@shizhongming so let me know if this works? we tested it yesterday but it would be good you see if it works. The other branch of Darren was merged

daren-thomas:
@JIMENOFONSECA @shizhongming I just saw that the #1064 pull request was merged into the branch `i665-Export-UBG-to-CEA-files` so... can we maybe merge that one first to master? I'm creating a pull request right now.

shizhongming:
We have another solution working that solves the problem of spatial reference system. We set aside this branch for the time being, and may come back in the future if necessary.

Issue #1100, opened 2018-03-06
KeyError: 'lon_0' when using photovoltaic script with reference-case-open


daren-thomas:
Steps to reproduce:

- extract new reference-case-open (`cea extract-reference-case`)
- run demand script (`cea demand`)
- run photovoltaic script (`cea photovoltaic`)

I expected this to just run through, instead, I get this error:

```
λ cea photovoltaic
Running photovoltaic with scenario = C:\reference-case-open\baseline
Running photovoltaic with date-start = 2016-01-01 00:00:00
Running photovoltaic with dpl = 200.0
Running photovoltaic with eff-pumping = 0.6
Running photovoltaic with fcr = 1.3
Running photovoltaic with k-msc-max = 0.217
Running photovoltaic with min-radiation = 0.75
Running photovoltaic with panel-on-roof = True
Running photovoltaic with panel-on-wall = True
Running photovoltaic with ro = 1000.0
Running photovoltaic with solar-window-solstice = 4
Running photovoltaic with t-in-pvt = 35.0
Running photovoltaic with t-in-sc = 75.0
Running photovoltaic with type-pvpanel = PV1
Running photovoltaic with type-scpanel = SC1
Traceback (most recent call last):
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\cea\Scripts\cea-script.py", line 11, in <module>
    load_entry_point('cityenergyanalyst', 'console_scripts', 'cea')()
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\interfaces\cli\cli.py", line 33, in main
    script_module.main(config)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\technologies\solar\photovoltaic.py", line 791, in main
    longitude = shp.crs['lon_0']
KeyError: 'lon_0'
```

daren-thomas:
@JIMENOFONSECA any ideas? is this resolved with your PR #1096 ?

daren-thomas:
Here is an IPython session showing the problem a bit more clearly:

```
In [1]: import fiona

In [2]: shp = fiona.open(locator.get_zone_geometry())

In [3]: shp.crs
Out[3]: {'init': u'epsg:32632'}

In [4]: locator.get_zone_geometry()
Out[4]: 'C:\\reference-case-open\\baseline\\inputs\\building-geometry\\zone.shp'
```

In `photovoltaic.py`, the following code is used to get latitude and longitude:

```
[...]

PR #1122, opened 2018-03-12
I1100 longitude problem

JIMENOFONSECA:
this solves #1100 @daren-thomas would you mind taking a look at this PR? thanks!


BhargavaKrishnaSreepathi:
before running need to do `pip install utm`. Even i faced the mentioned error, so used the piece of code in this PR and it solved the error

BhargavaKrishnaSreepathi:
@JIMENOFONSECA: Similar error will arise when we run `cea\technologies\solar\photovoltaic_thermal.py` and `solar_collector.py`. I used the same changes done in `photovoltaic.py` and the error is solved. I suggest that you implement the changes in these files too in this PR

JIMENOFONSECA:
oh,oh that seems to be a problem let me check. Probably Because I changed the digital elevation model granularity, unbelievable how sensitive this can be.

JIMENOFONSECA:
@daren-thomas this is ready to test again

Issue #1125, opened 2018-03-13
a bug corresponding to the change of individual where NG boiler is being changed to BG Boiler


BhargavaKrishnaSreepathi:
There is a bug in the master, where during optimization, the NG Boiler is changed to BG Boiler and vice versa. This creates a mismatch in the individual, as the corresponding numbers are 1 and 2.

So the exact individual will not always be created.

Learnings till now include:

1. Both the individuals, that is individuals with 1 or 2, both have no issues in running. It is just that the individual is being changed.

2. The only fix we made till now is in the cea/plots/optimization/dashboard_district, where the testing is made as `len(columns_of_saved_files) - 1` instead of just `len(columns_of_saved_files)`, to handle this bug

PR #1126, opened 2018-03-13
I1058 heat load balance PR apart from optimization folder structure changes

BhargavaKrishnaSreepathi:
This PR mainly does two things:

1. Check the load balances between heating, cooling, electricity for all the individuals that are being generated in CEA. The differences were made evident in the dashboard. This was mainly because of the columns that were being imported and being compared. Now, this is fixed. So from now, the `cea\plots\optimization\dashboard_district.py`, when run, should produce balanced graphs.

2. The optimization folder structure is changed in this PR. Before all the saved files corresponding to activation patterns of all population was stored in `outputs/data/optimization/slave`. So it was quite messy in identifying the individuals of a particular generation. Now the files are saved in their corresponding generation folders in `outputs/data/optimization/slave/gen_number`. Apart from this, a csv file is present in each of these folders with all the individuals which were calculated in that generation. Also, the files are now saved as `ind_number_Heating_Activation_Pattern.csv`, instead of the big `configuration` attached to it. This is linked to the above mentioned csv, which also has individual number and generation number. Finally, a csv containing all the individuals along with their individual number and generation number is saved in `outputs/data/optimization/slave` for the convenience of referencing in dashboard. So the plots will directly call this csv and then find the address of the corresponding individual

3. Apart from the main things above, small changes in `cea\plots\optimization\dashboard_district.py` were done to comply with the updated structure and the matching of the individual in the new csv structure.

**NOTE**: A bug has been identified, which was already existing on Master, where during optimization, the NG Boiler is changed to BG Boiler and vice versa. This creates a mismatch in the individual, as the corresponding numbers are 1 and 2. This is being dealt in a separate issue #1125 

The way to test this PR is by following the steps in the guide http://city-energy-analyst.readthedocs.io/en/latest/cea-workflow-guide.html and then running the dashboard_district.py in cea/plots/optimization and then seeing the plots


JIMENOFONSECA:
@BhargavaKrishnaSreepathi somehow the branch has some conflicts to be solved once it is intended to be updated from master

JIMENOFONSECA:
@BhargavaKrishnaSreepathi sewage heat exchanger is not running in the PR, would you mind checking this out?

BhargavaKrishnaSreepathi:
@JIMENOFONSECA: fixed it, it was because of removing the constants from global variables. This will be comprehensively addressed in another PR in coming days. I made a hot fix for this PR for now

JIMENOFONSECA:
I run the network, but it seems that the next file name changed so the optimization cannot find it.
could you check by eliminating the previous files and running the optimization?

 IOError: File C:\reference-case-open\baseline\outputs\data\optimization\network\layout\DH__AllEdges.csv does not exist

BhargavaKrishnaSreepathi:
@JIMENOFONSECA: did you run the `cea\technologies\thermal_network\thermal_network_matrix.py`. This will take a long time, but will generate the corresponding network files

JIMENOFONSECA:
@BhargavaKrishnaSreepathi right, I just needed to wait until the whole script finished

PR #1130, opened 2018-03-14
I1128 fix demand output

gabriel-happle:
The variables `mcpcsf_kWperC` and `mcphsf_kWperC` reintroduced to demand output `.csv`.

Valid choices updated in the `default.config` file.

### DIFF
--- a/cea/default.config
+++ b/cea/default.config
@@ -140,7 +140,7 @@ buildings.category = Advanced
 
 loads-output =
 loads-output.type = MultiChoiceParameter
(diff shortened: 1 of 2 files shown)


cea/default.config
```diff
@@ -140,7 +140,7 @@ buildings.category = Advanced
...
 loads-output.help = a list of loads to run the simulation for - leave blank to simulate all loads in demand_writers
```
JIMENOFONSECA: @gabriel-happle actually, there is one input variable for temperatures and massflows called temperature-output, massflows-output. Please use those to indicate which temperatures and massflows to output. 

No need to include Name, Af_m2, Aroof_m2, GFA_m2, people0 as tehy are always by default part of the output.

PR #1134, opened 2018-03-15
I928 subset simulation

Khayatian:


JIMENOFONSECA:
@Khayatian would you let me know how to test this?

Khayatian:
this branch had some issue, therefore, there is no need for merginig. We can delete it and use the branch "i1076-revert-to-last-working" instead

Issue #1010, opened 2018-01-31
dodo test for thermal network


shanshanhsieh:
we have to add the thermal network to the dodo test.
the test should check if the `thermal_network_matrix.py` is running for reference-case-open.
@daren-thomas  could you please give me a hand on that? : )

daren-thomas:
sure! what are the steps to take to ensure it runs? how long will computation take?

shanshanhsieh:
Thanks, @daren-thomas !

Here are the steps:
After obtaining the demand results for reference-case-open, we have to run `thermal_network_matrix.py` for `DH` and `DC` (change 'network-type' in 'cea.config')

It takes 25 mins to run reference-case-open with 'DH', and a bit less for 'DC'
However, it is more important to make sure 'DH' runs.
If we want to save time, it is good enough to run the script only for 'DH'

daren-thomas:
@shanshanhsieh this will increase the time `cea test` takes - both locally (when reviewing pull requests) and on the jenkins server (run for each update to each pull request) and might decrease the value of the dodo test... so i'm wondering... i think we definitively should implement that on the "merge-to-master" part of the jenkins!

BUT: can you think of a way to test the important parts without running the whole script? can it be split up into functions / code-paths that can be tested once?

shanshanhsieh:
yea, i was thinking about the same thing... 
It is quite important to make sure the outputs from parts of the script doesn't break other parts, therefore, i would like to run it from the begining to the end.
Instead of running the script for the whole year, we could do it for 2000 timesteps, which will only take 5 minutes. 

To do this, we have to change two lines of code before running dodo. 
And the question is where to implement the changes?
1. As this modification is only required for the dodo test, modify it everytime via the `dodo.py` script.
2. Implement a `timestep` variable in `cea.config`

daren-thomas:
OR: split the code for a single timestep into a separate function. so you end up with this structure:

```
def main(config):
    pre_simulation(config)
    for t in xrange(HOURS_OF_YEAR):
        simulate_timestep(t, config)
    post_simulation(config)
```

Then, in dodo, you can just run the `pre_simulation`, then 2000 times the `simulation_timestep` and then the `post_simulation`.

Of course maybe you'd end up with some variables to communicate between these steps - those are added parameters.

daren-thomas:
[...]

PR #1140, opened 2018-03-20
I1090 multiprocessing for thermal network

daren-thomas:
This adds multiprocessing for the thermal_network_matrix module - and a TON of refactorings to make that possible. Many thanks to @roglen for his help here! Also, this includes a fix to #1010 - adding a jenkins test for the thermal_network_matrix and while we're at it, also makes it possible to run a single test from the command line (`cea test --tasks run_thermal_network_matrix`)


roglen:
Tested reference case zug (DH), looped network N5 (DH) and mixed network N7 (DC)

Ran thermal_network_matrix.py and compared output values of:
-  Plant heat requirement
- Total pressure loss (P_DeltaP_Pa)

for the master compared to this branch. Also compared outputs with or without multiprocessing. 

Results show very small variations of output to master (<1%), this is most likely due to changes in the rounding of the edge mass flows. 

This test covers the files: thermal_network_main and substation_matrix , these two are ready to merge.

PR #1149, opened 2018-03-22
Thermal network plots alternate

roglen:
Same Pull request as before but cleaned up. 

Tested and running.


roglen:
@shanshanhsieh no changes to the last update of the previous request, are we ready to merge?

Issue #1132, opened 2018-03-15
Reverts and posterior PR did not fix the energy balance diagram


JIMENOFONSECA:
@gabriel-happle so the energy balance was not fixed after the last PR. somehow there are some quantities missing to run the diagram. 

It seems we might have forgotten some part of the script to be passed when we reverted the PR.

PR #1151, opened 2018-03-22
refactoring and re-adding final heating and cooling demand calc

gabriel-happle:
a small addition and refactoring in the demand calculation that unfortunately somehow got lost in the reverted PR of issue #1132


gabriel-happle:
@shanshanhsieh I created the PR.

Issue #1121, opened 2018-03-09
i1076-NN-Recurrent-Conv multi-processing


Khayatian:
There is an issue which I cannot understand.
branch:      i1076-NN-Recurrent-Conv
file:            ~/demand/metamodel/nn_generator/input_prepare.py

The file is called from another script "scaler_sampler.py" and contains two modules:

lines 40-68 which performs a task with multi-processing
lines 72-86 which performs the same task as the one above without multi processing
**(you should untag the lines 72-86 and tag (#) lines 40-68 to check the single processing, the latter sequence of script (lines 72-86) is kept as a backup for situations like this when the multi processing fails)**

while the latter (72-86) without the multi processing works, the former (40-68) with multi-processing fails.

the error reads:
...
  File "C:\CEAforArcGIS\cea\demand\metamodel\nn_generator\input_prepare.py", line 53, in input_prepare_main
    NN_input_ready , NN_target_ready=job.get(240)
  File "C:\Anaconda2\envs\cea\lib\multiprocessing\pool.py", line 567, in get
    raise self._value
zipfile.BadZipfile: File is not a zip file
...

Could you please review this and provide me with some feedback?

JIMENOFONSECA:
@daren-thomas I got the same error. would you mind giving us a hand with this multi-processing problem?

PR #1160, opened 2018-03-23
I1076 revert to last working

Khayatian:
all scripts of the neural network surrogate is correctly working


Khayatian:
This branch update the neural network scripts, all modules have been tested. However the multi-processing is not yet implement as I will need some help to solve the issue. Please see issue #1121

PR #1167, opened 2018-03-28
Two versions of CEA logo and one MUSES

shizhongming:
A folder containing these three logos in the png formate has been created.


JIMENOFONSECA:
@shizhongming would you try to update your branch from master? that might fix the error

daren-thomas:
@JIMENOFONSECA @shizhongming is there a reason we don't also include the vector formats? png is really bad when you need to scale the image...

PR #1168, opened 2018-03-28
I1152 linking thermal pressure losses to dhn

BhargavaKrishnaSreepathi:
This PR links the pressure loss calculations to the thermal_network_matrix and the piping cost to the overall network laid. The piping database is updated with values from China.


JIMENOFONSECA:
@BhargavaKrishnaSreepathi maybe if you update from master it could work?

JIMENOFONSECA:
@BhargavaKrishnaSreepathi would you mind checking the conflicts? This needs special care as it might be possible that someone also improved those files before. Make sure it is the latest version

PR #1174, opened 2018-03-29
I1173 atot calculation

martin-mosteiro:
This PR solves an error in the RC-model in CEA up to now.

bpr.rc_model['Atot'] and bpr.rc_model['Am'] were not calculated consistently and therefore generated impossible results (e.g. 160% of internal gains going into the thermal mass node).

The simplified calculation for Atot from ISO 13790 is now used, once the new standard is implemented both bpr.rc_model['Atot'] and bpr.rc_model['Am'] will be replaced by detailed calculations.


gabriel-happle:
Ready to merge if conflicting files are resolved

Issue #1177, opened 2018-03-29
Cannot run CEA with a using a modified database


martin-mosteiro:
Currently, when you run the demand script for a scenario the corresponding regional databases are copied into the scenario folder. This means that if you edit the databases in your scenario, they'll get overwritten when running the demand.

Based on feedback from @daren-thomas, that is true *unless* you specify the region as "custom", in which case the demand script uses the databases in the scenario folder. Indeed, if you run the demand with **region = anything_else**, you get the following:
```
AssertionError: Invalid parameter, choose from: ['CH', 'SIN', 'custom']
```

However if you run it with **region = custom**...
```
IOError: [Errno 2] No such file or directory: 'C:\\Users\\User\\Documents\\GitHub\\CityEnergyAnalyst\\cea\\databases\\custom\\archetypes\\system_controls.xlsx'
```

PR #1178, opened 2018-03-29
Fix _get_region_specific_db_file

martin-mosteiro:
This PR solves issue #1177 by fixing a bug in `_get_region_specific_db_file` so that if the region is specified as "custom" the database is imported from the scenario folder (and not the CEA database folder).


martin-mosteiro:
Not quite. Before the statement was “if the region is custom and the files aren’t there, raise error; otherwise get the files from the CEA database”. Since there is no “custom” region in the CEA database, that created the error I copied.

The version in the PR does “if the region is custom, if the files are not there raise error (but don’t do anything if the region is custom and the files *are* there); otherwise get the files from the CEA database”.

Or try it and you’ll see this one works as intended ;)



On 29 Mar 2018, at 20:47, Daren Thomas <notifications@github.com<mailto:notifications@github.com>> wrote:


@daren-thomas requested changes on this pull request.

I'm pretty sure these two forms are equivalent.

martin-mosteiro:
In other words, the current code can't correctly handle the case `if region == 'custom' and os.path.exists(result_file)`

martin-mosteiro:
@daren-thomas Did you have a chance to check what I said above? Otherwise could you please suggest what should be done differently?

PR #1188, opened 2018-04-06
functionality to check for successful merge of emission_systems and e…

gabriel-happle:
…nvelope_systems database

This PR prevents that buildings are dropped from the demand calculation due to faulty specifications of architecture or technical systems. Instead it breaks the code and gives some info why.

This functionality can be tested by adding *funky* keys (e.g. 'T55') to the inputs `technical_systems.dbf` or `architecture_systems.dbf` OR by changing/deleting keys in `cea\databases\systems\emission_systems.xls` or `cea\databases\systems\envelope_systems.xls`


gabriel-happle:
@daren-thomas this could maybe also be useful for the `datahelper` script.
I wrote this function because the merge option `how="left"` just fills non-existing values/rows in the result dataframe with `nan`. I think that doesn't help us.

gabriel-happle:
Hi @daren-thomas  I think the database is not really fixed. For some systems, yes, different parts of code are called, so it is kind of fixed. But for other systems it's just a look-up table.
With this check it just makes sure that every row is merged with something.

You mean we should check if the building input properties only contain keys that are also in the database?
Maybe that is a better option.
However, in the concrete case of Zhongmings case study failing, that would be a bit strange, because I think it's actually the database that is at fault...

I can write such checks. You just would have to wait, because I discovered some higher priority mess up today.

JIMENOFONSECA:
@daren-thomas @gabriel-happle I would like to close modifications to the core by monday. Is this good to go for the executive course release?

gabriel-happle:
what does @daren-thomas think?

daren-thomas:
we should feature freeze as quickly as possible. but i still can't get the plots script to run through...

gabriel-happle:
I am not actively working on this.
So if @daren-thomas thinks the code as it is now is not worth to be implemented, we can just not merge this feature and make sure manually that all our case studies run when we change building properties, or when participants change building properties.

JIMENOFONSECA:
@gabriel-happle @daren-thomas closing until ready to merge

PR #1191, opened 2018-04-08
9 Scenarios for WTP case studies

shizhongming:
These nine scenarios include different land use mixes and different
density (low, medium and high for each land use combinations). I ran all
of them through data-helper, radiation-main and demand already. Since
the file is so big, I excluded the raidiation results in this commit. If
you run it, it will take 10 mintues or so each.


daren-thomas:
@shizhongming - I think you chose the wrong repository for the scenarios. please check in with @JIMENOFONSECA - there is a repository called `cea-reference-cases` that would be a better fit (and is private, as opposed to the cea repository which is public and scenarios might include sensitive data)

JIMENOFONSECA:
@shizhongming @daren-thomas right, this is not the repository. It should be in the reference cases repository

amraladdin:
So @shizhongming, should I wait till you guys are done moving the scenarios to the correct repository?

PR #1206, opened 2018-04-11
I 209 psychroplotting for cea

gabriel-happle:
code cleanup and documentation needs to be done

creates a graph like this:
image


gabriel-happle:
@JIMENOFONSECA I want to finish this today.
What kind of changes to the visual appearance of the plot should be implemented?
What kind of additional analysis should be presented in the table?
Inputs from @martin-mosteiro and @shanshanhsieh welcome.

image

PR #1228, opened 2018-04-13
I1169 integrating thermal network matrix with the new demand outputs

roglen:
Updates the thermal network matrix script to incorporate the new cooling demands. Tested with reference case open DC and DH mode. This PR eliminates the use of the aggregated heating and cooling loads. HEX at the substations are split into a separate HEX for each demand type, e.g. ahu, aru, shu, etc.

To check:
- [ ] run a DC network
- [ ] run a DH network
- [ ] enable and disable substation systems (lines 2576, 2577) and evaluate changes (e.g. with breakpoints)


shanshanhsieh:
hey @roglen , the new script has not quite resolved this issue yet.
because the `substation_HEX_specs` should be re-done as well

here are the required changes: 
1. resize all HEXs: in `substation_HEX_sizing`, please replace the HEX for Qcf with separated HEXs for Qcsf_aru, Qcsf_ahu, Qcsf_scu, Qcdataf 
2. in the same function, Qhsf should be replaced with 3 HEX (Qhsf_ahu, Qhsf_aru, Qhsf_shu)
3. `calc_substation_return_DC` should follow the same format for substation_return_DH, which is to simulate the return temperatures for all 4 HEXs for cooling, and the `T_DC_return_K` is the mixing temperatures of all HEXs
4. `calc_substation_return_DH` should be further modified as well, hence there are more HEXs 
5. go through the line 390-395 in `thermal_network_matrix.py` carefully and make sure all the calculations are coherent

roglen:
@shanshanhsieh thanks for your feedback and sorry for my initial confusion! I have made the discussed changes and run the code in DH and DC mode, with different lists of substation systems. Please double check if it now covers the changes the way you imagined it :)

PR #1250, opened 2018-04-19
I1248 update thermal network plots

roglen:
PR updates network plots:
- adds bar chart of losses for each edge
- adds new network plots
- fixes a bug in Node placement
- loss duration curve now for pressure loss only
- updates network plot labels and colors
- fixes bug in new demand outputs for looped networks

To test:
- [ ] run `thermal_network_matrix.py` and `plots/thermal_networks/dashboard_district.py` for DC and DH mode 
- [ ] check for improvement potential in new plots


roglen:
@shanshanhsieh the losses_curve plots are currently using Qshf and Qcshf and are not checking which demands are connected to the network. Should I leave it this way for now? 
The main reason I have not yet changed it is the fact that the list containing the connected network demands (e.g. Qcsf_aru,... ) is currently a list only locally available in the `thermal_network-matrix.py', and as far as I know we are not yet sure where we will move it.

roglen:
Branch is cleaned up, requested changes are made. I've tested for the open case for DH and DC.
Please have a look again @shanshanhsieh  and thank you for your previous feedback! Looks much better now.

PR #1251, opened 2018-04-19
Integrating cooling technologies into CEA

BhargavaKrishnaSreepathi:
Based on the new demand files which split the load into AHU, ARU and SHU/SCU components, the downstream files had to be adapted. This PR does the same along with few major changes in the optimization.

In this PR, the individual generated by the optimization will have the following:
1. heating block corresponding to heating technologies and their corresponding share in the Peak load
2. indices corresponding to the temperature supplied in the DHN and the number of units among 'AHU, ARU, SHU' to which the DHN supplies
3. buildings connected to DHN
4. cooling block corresponding to cooling technologies and their corresponding share in the Peak Load
5. indices corresponding to the temperature supplied in the DCN and the number of units among 'AHU, ARU, SCU' to which the DCN supplies
6. buildings connected to DCN

Thus the length of the individual has changed significantly.

To match with this, decentralized building calculations were also changed. Now there are two files corresponding to heating and cooling. This PR works exclusively for the cooling case, as we will be needing it for the executive course. In future, the same should be included for the heating case.

In this PR, the cooling technologies have also been added, namely VCC, VCC backup, Storage tank for cooling.

Substation.py file has also been changed to appropriately include the loads corresponding to data center cooling, refrigeration and space cooling.

The best way to test the PR is to run everything from scratch using the reference-case-open and setting the Region as SIN and any singapore based weather files. Good luck testing and let us know if there are any comments. While this is going through the review process, I will add detailed docs


JIMENOFONSECA:
@shanshanhsieh @BhargavaKrishnaSreepathi 

First couple of errors while running the new case study scenario 1. I can pass them to you in an USB so you do not need to run the thermal network, demand, solar etc..

image

image

PR #1273, opened 2018-05-08
1116 Automated setup of looped networks

roglen:
This PR adds code functionality to create looped networks starting from the Steiner minimum spanning tree. If two nodes are neighbors in the street layout but not in the Steiner tree, they are connected. As a result, loops are generated for all cases where nodes can be directly connected to form a loop.

A new line in the config file is added in the `network-layout` section, called `allow-looped-networks`. This flag is default set to False and toggles the function above on (true) or off (false).

The code was tested with WTP cases 1-3. 

To review this PR please:
- [x] generate a new config file, set `allow_looped_networks' to `true`
- [x] run `cea/technologies/thermal_network/network_layout/main.py` (preferably in DH mode as DC has been tested)
- [x] review edges in ArcGis for possible duplicates or empty attributes (length, type, etc.)
- [x] run `thermal_network/thermal_network_matrix.py`
- [x] run `plots/thermal_netowrk/main.py`


roglen:
@JIMENOFONSECA  this may also be of interest to you.

JIMENOFONSECA:
@roglen So the minimum mass flows passed and the diagrams look quite good. what about this?

roglen:
Shanshan is still reviewing now that it is updated from master. Should be ready soon!

shanshanhsieh:
@roglen , it is almost ready beside the weight column in `edge.shp`, please let me know when that is modified.

PR #1299, opened 2018-05-14
I1288 substation calling only cooling for design

roglen:
Small PR to fix substation HEX calling function. 
Renamed variables and adapted call of functions. 
Outputs verified to be in correct order of size wth HISAKA online calculation tool. 
(https://www.hisaka-ex.com/simulator_english/)


roglen:
@shanshanhsieh I made the requested changes, please have a look

PR #1304, opened 2018-05-14
Fixes issue in mobility.py

martin-mosteiro:
The error was caused by the list of occupancy types being defined by the mobility database instead of the occupancy types in the actual reference case. mobility.py now runs as expected.


JIMENOFONSECA:
@roglen Hi Lenny, does this fix the problem?

roglen:
@JIMENOFONSECA This solves the issue for me. Both mobility.py and plots for LCA run. Ready to merge.

PR #1306, opened 2018-05-15
fixing guide

JIMENOFONSECA:


JIMENOFONSECA:
@roglen would you mind checking if this addresses the issue?

roglen:
@JIMENOFONSECA I added one line on the datahelper and made some small changes (typos, spelling).
Can you quickly check that what I wrote about the datahelper is accurate?
Other than that this fixes the issue!

JIMENOFONSECA:
@roglen actually there is a substep including the data helper in Step 1. you need to go to create new project. Do you think this is too hidden?

roglen:
@JIMENOFONSECA ah now i see, it's in the guide for new projects. 
Well I think it's ok as long as we don't need the data helper to run any of our reference cases, and only if we create our own setup. 
I see this guide as the step by step I need to follow to run any of the provided CEA case studies successfully. So if I use files from the 'cea-reference-case' repository I wouldn't have checked the 'new-project-guide'.

PR #1334, opened 2018-05-16
I1305 bug in master

BhargavaKrishnaSreepathi:
This PR should fix the bug in the master. The bug arised because the substation file is not being run in the dencentralized_buildings_heating scenario. This PR fixes it. Also when I am at it did some minor refactoring in storage files. 

To test it, run the decentralized_main, followed by optimization. Follow this with running the plots

### DIFF
--- a/cea/optimization/preprocessing/disconnected_buildings_heating.py
+++ b/cea/optimization/preprocessing/disconnected_buildings_heating.py
@@ -18,6 +18,8 @@ from cea.resources.geothermal import calc_ground_temperature
 from cea.utilities import dbf
 from geopandas import GeoDataFrame as Gdf
 from cea.utilities import epwreader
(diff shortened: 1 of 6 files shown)


JIMENOFONSECA:
@shanshanhsieh would you mind checking if this runs for you? @BhargavaKrishnaSreepathi what about changes in the optimization.main.py which deletes the dataframe of cooling? is that intended?

BhargavaKrishnaSreepathi:
@JIMENOFONSECA: yes it was intentional, as the centralized network-type is 'DH' meaning there will be no cooling supplied by the network, so the calculations will not be present. And only when network-type is 'DC' then the cooling need to be concentrated.

shanshanhsieh:
@JIMENOFONSECA , i can test it

PR #1348, opened 2018-05-18
Solves call from ArcGis

roglen:
Solves issue of thermal netowrk plot not being called correctly if only thermal network plots are run.


roglen:
@JIMENOFONSECA  ready to merge

PR #1356, opened 2018-05-18
Fix mobility issue again

martin-mosteiro:
PR #1304 was reverted because there appeared to be an issue with master. This PR solves the mobility issue again.


JIMENOFONSECA:
@martin-mosteiro could you check why it breaks?

martin-mosteiro:
That's strange, I don't see why that one line of code should affect the tests. I'll look into it by Wednesday.

martin-mosteiro:
@JIMENOFONSECA Good catch - I should've checked the rest of the code but just hadn't had time. The code looks good to me now and the branch seems to pass all tests.

PR #1371, opened 2018-05-22
I1279 update pressure loss calculation in thermal network matrix

roglen:
This PR adds pressure losses at substations, modeled by a Valve and HEX for each demand type (e.g. ahu, aru, etc.) 
HEX losses are modeled based on a linear fit equation of mass flows to pressure losses from datasheet values. 

Edge pressure losses in plots and outputs have also been updated to reflect both supply and return network instead up supply network only. 

The substation pressure losses are plotted in a new bar chart and added to the existing network plots. Variable naming has been unified. 

To verify this PR:
- [ ] review linear fit equation for HEX pressure losses (datasheets and source excel file uploaded in cea-validation repository). Double check with ecocampus measured data to see if scope of results is correct. Especially for DC, substation pressure losses largely exceed edge pressure losses. 
- [ ] run thermal_network_matrix and plots for both DH and DC.


JIMENOFONSECA:
@roglen to solve conflicts i recommend to compare visually your files and those of master (including the last PR of Bhargava) and just push the one that looks better. Unfortunately that is the only way to fix conflicts in excel files

roglen:
@JIMENOFONSECA already working on it :+1: 
This branch still needs review anyways because we need to check with the ecocampus data if the pressure losses are in the correct order of magnitude.

PR #1397, opened 2018-05-24
1263 api build debug

Jack-Hawthorne:
I've made some changes to the make.bat to generate fresh rst files every time a make clean or html is run This is to ensure new modules are accounted for and existing ones are updated.
I've also changed the docs config to include some special methods for which documentation was written but not created.


daren-thomas:

Jack-Hawthorne:
the specific build issues will be address one by one as i go through the documentation.

docs/conf.py
```diff
@@ -195,3 +196,18 @@ def __getattr__(cls, name):
...
 
+
+# ## Add documentation for python special methods (with exclusions)
+
+def autodoc_skip_member(app, what, name, obj, skip, options):
+        exclusions = ('__weakref__',  # special-members
+                  '__doc__', '__module__', '__dict__',  # undoc-members
+                  )
+        exclude = name in exclusions
+        if name == name in exclusions:
```
daren-thomas: I don't understand this line.

docs/make.bat
```diff
@@ -4,6 +4,9 @@ pushd %~dp0
 
 REM Command file for Sphinx documentation
 
+sphinx-apidoc -f -M -T -o modules ../cea
+print(Sphinx has updated rst files)
```
daren-thomas: did you check this? i'm not sure it can work.

daren-thomas: yep. batch doesn't have a `print(...)` statement. replace with:
```
echo Sphinx has updated rst files
```

Jack-Hawthorne: no worries ill have a look at these now

Jack-Hawthorne:
@daren-thomas please have a look at the latest commit with your advised changes

cea/optimization/preprocessing/preprocessing_main.py
```diff
@@ -1,6 +1,6 @@
 """
 ============================
-pre-processing algorithm
+Pre-processing algorithm
 ============================
```
JIMENOFONSECA: @Jack-Hawthorne it is not clear why some of the lines  "===============" are erased and some others are not across different scripts. I think they should be erased from everywhere except where they are needed to be read by sphynxk. Would you mind to check for consistency?

daren-thomas:
@Jack-Hawthorne you still have not addressed the question i had about `docs/conf.py` - come talk to me when you get to the office: I'd like to show you why I don't think this does what you think it does:

```
        if name == name in exclusions:
            return skip
```

daren-thomas:
Also, this needs to get merged as soon as possible. I already had to do a lengthy merge to get it back on track with master.

PR #1414, opened 2018-05-29
fixes #1386 - ArcGIS interface: Plotting tool only works if at least one scenario for comparison is selected

### LINKED ISSUE #1386: ArcGIS interface: Plotting tool only works if at least one scenario for comparison is selected
image

This is not a critical bug.

daren-thomas:
To test, run `cea install-toolbox` and restart ArcScene. Run the plotting tool with no scenarios for comparison selected and make sure the list of categories of plots to create includes "scenario_comparisons". Running this, should produce the following output:

```
Executing: PlotsTool SIN C:\reference-case-open\baseline # # scenarios_comparisons 1;2;3 ind1 DH #
Start Time: Tue May 29 13:04:50 2018
Running script PlotsTool...
Executing: C:\Users\darthoma\AppData\Local\conda\conda\envs\cea\python.exe -u -m cea.interfaces.cli.cli plots --buildings  --network-names  --scenario C:\reference-case-open\baseline --region SIN --network-type DH --scenarios  --individual ind1 --generations 1, 2, 3 --categories scenarios_comparisons
Running `cea plots` with the following parameters:
- general:region = SIN
- general:scenario = C:\reference-case-open\baseline
- plots:scenarios = []
- plots:buildings = []
- plots:categories = ['scenarios_comparisons']
- plots:generations = ['1', '2', '3']
- plots:individual = ind1
- plots:network-type = DH
- plots:network-names = []
Running plots for the next categories = ['scenarios_comparisons']
ERROR: Comparison plots require at least two scenarios to compare. See config.plots.scenarios.
Tool did not run successfully: Check parameters
Completed script PlotsTool...
Failed to execute (PlotsTool).
Failed at Tue May 29 13:04:53 2018 (Elapsed Time: 3.46 seconds)
```

Note the "ERROR:..." and the "Tool did not run successfully: Check parameters" output. This is highlighted in the tool output window in blue and red. No stacktrace is shown as this error is expected, by checking in the beginning of `cea.plots.comparison.main.plot_main`.

BONUS: Other tools that check their parameters can raise `cea.ConfigError(msg)` to provide the same behavior. This removes some of the "scary" from error messages that result from just testing if the input is sane and should be done everywhere in the code. We can add more diverse exception classes to `cea.__init__.py` to handle other classes of errors that we test for.


martin-mosteiro:
capture
I still get this issue when running with no value for 'DC' or 'DH' selected - but it's solved if there's a default value for these before you select it.

daren-thomas:
@martin-mosteiro can you try again please?

martin-mosteiro:
The error message says DH or DC should be selected, even if you are not running thermal networks plots. But as I hinted at in my previous message, since one of 'DH' or 'DC' is already given by default, the user would have to go and delete the default value and run any kind of plots. So it's kind of a bug, but not a major one.

JIMENOFONSECA:
@martin-mosteiro I will merge it as the bug is not a major one

PR #1427, opened 2018-05-31
1238 glossary terms and databases

Jack-Hawthorne:
The glossary function is finished, however the metadatabase.csv still needs to be updated to reflect any changes since the last tutorial documents were created. Thomas (a new assistant) and I will be working on this over the next couple of days.
@JIMENOFONSECA was this what you intended? Feedback is welcome.


JIMENOFONSECA:
@Jack-Hawthorne this is a good start. However I see it complicated to implement new inputs in this format. plus the idea is that this file replaces plots/naming.csv  which is used as a repository of the variables plotted in CEA. Maybe a combination of what you have and this is what we are looking for.

daren-thomas:

Jack-Hawthorne:
@JIMENOFONSECA that is a good idea to include plot variables, ill have a look into it and update you.

@daren-thomas i agree this could be done in a much better way, but as you've said, this can be done later.

daren-thomas:
@JIMENOFONSECA is there a reason for not merging this?

JIMENOFONSECA:
We could, but the next PR  #1454 changes the subscripts of all the variables in the next fashion:

hsf  ——> hs_sys
csf  ——> cs_sys
wwf  ——>ww_sys

I recommend Jack to use the tool of replace text in csv. and replace all with hsf to hs_sys etc… that is an easy way to do it.

The name of the variables will be

“system heating demand” instead of “final heating demand”
“system cooling demand” instead of “final cooling demand”
etc...


On 7 Jun 2018, 3:55 PM +0800, Daren Thomas <notifications@github.com>, wrote:


@JIMENOFONSECA<https://github.com/JIMENOFONSECA> is there a reason for not merging this?

daren-thomas:
so... why not merge this and create a new issue?

JIMENOFONSECA:
Sure, please go ahead with merging and the new issue.
On 7 Jun 2018, 16:11 +0800, Daren Thomas <notifications@github.com>, wrote:

so... why not merge this and create a new issue?

PR #1432, opened 2018-06-01
IMPORTANT SOLAR RADIATION FIX!

gabriel-happle:
When creating the JSON of radiation results we were using the results for building 2 for ALL buildings except builiding 1 due to not counting up the index.

### DIFF
--- a/cea/resources/radiation_daysim/daysim_main.py
+++ b/cea/resources/radiation_daysim/daysim_main.py
@@ -173,5 +173,5 @@ def isolation_daysim(chunk_n, rad, geometry_3D_zone, locator, weather_path, sett
         items_sensor_name_and_result = dict(zip(sensor_code_building, selection_of_results))
         with open(locator.get_radiation_building(building_name), 'w') as outfile:
             json.dump(items_sensor_name_and_result, outfile)
-        index = sensors_number_building
+        index = index + sensors_number_building
 


gabriel-happle:
I will update the test results asap!
unfortunately this changes every simulation result we have ever done with daysim :/

on the plus side, I can now visualize Daysim and Radiance geometry with Matlab (if anyone is interested).

gabriel-happle:
FYI:
image

gabriel-happle:
This means we also have to update the radiation files in the case studies...
Can someone help me @JIMENOFONSECA ?

gabriel-happle:
@daren-thomas how can I run `dodo` with my own, new radiation files?

martin-mosteiro:
Wow, brilliant stuff!
Do you have a quick summary/writeup about the visualization part?

daren-thomas:
@gabriel-happle check cea/tests/radiation_data - you need to update these.

gabriel-happle:
@JIMENOFONSECA @martin-mosteiro @amraladdin @shanshanhsieh 
We did some more testing today.
We discovered some behavior that we can not explain - maybe daysim internal stuff.
If we set `n-buildings-in-chunk=1` vs. `n-buildings-in-chunk=X` only the first building in every chunk produces an identical `json` radiation file.
E.g., `n-buildings-in-chunk=5` vs. `n-buildings-in-chunk=1` `B001, B006, B011,...` will have identical results, while all other buildings have small differences in the radiation output of daysim.

Because of this behavior we suggest to use:
`n-buildings-in-chunk=1` as default simulation. The run time is not that different from larger chunks.

Also we observed that if `consider-windows=false AND consider-floors=false` there will be no sensors placed on the walls.
We therefore suggest to remove the `consider-windows` option from the `config` and set the value to always `true`.

For higher accuracy of the results the `consider-floors` should also be `true`.
I create an issue to change the default config and recalculate all the radiation for the test case and case studies.

PR #1439, opened 2018-06-04
update ceajenkins.py to use ngrok.io instead of localtunnel.me

daren-thomas:
The Jenkins is not working correctly. This update switches the tunneling solution used to use ngrok.io (paid) instead of localtunnel.me (free).


daren-thomas:
NOTE: Still need to write documentation before merging.

JIMENOFONSECA:
@daren-thomas I assume it is ready and can put jenkins online again

PR #1440, opened 2018-06-04
1401 - Add shapefile-to-excel tool to ArcGIS

### LINKED ISSUE #1401: Add Shapefile-to-excel tools back to ArcGIS interface
Somehow these got removed...

@JIMENOFONSECA  before i put them back, quick check: was this on purpose or an accident?

daren-thomas:
Fixes #1401 with an assumed default CRS value of {"init": "epsg:32632"} (taken from the builtin reference-case).

### DIFF
--- a/cea/config.py
+++ b/cea/config.py
@@ -98,6 +98,8 @@ class Configuration(object):
                             parameter.replace_references(
                                 command_line_args[parameter.name])))
                 except:
+                    import traceback
+                    traceback.print_exc()
                     raise ValueError('ERROR setting %s:%s to %s' % (
                         section.name, parameter.name, command_line_args[parameter.name]))
                 del command_line_args[parameter.name]
--- a/cea/default.config
+++ b/cea/default.config
@@ -506,7 +506,7 @@ shapefile.type = FileParameter
 shapefile.extensions = shp
 shapefile.help = Path to the ESRI Shapefile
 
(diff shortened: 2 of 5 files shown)


JIMENOFONSECA:
@daren-thomas the coordinate system official for cea is contained inside the functions

1. shapefile_to_WSG_and_UTM(shapefile_path):

and 

2. get_projected_coordinate_system(lat, lon):

inside the file cea/utilities/standarize_coordinates.py

the official projected coordinate system of CEA is UTM.

daren-thomas:
@JIMENOFONSECA just looking into the function `shapefile_to_WSG_and_UTM` and noticed: the `ensure_cpg_file` function is problematic at best! It's actually a bug, because it just overwrites the code page used. What you want to do instead is use the cpg defined to read the data and then re-encode the data in the code page you want (`ISO-8859-1`). I'm creating an issue for this (#1444) - we can fix it after the executive course.

daren-thomas:
@JIMENOFONSECA as far as I can tell we have a problem with this functionality: Converting the shapefile to Excel will encode the geometry based on the projection. And to my understanding so far, the projection is dependent on longitude and latitude of the project. This direction is not the problem. The other is:

Given an Excel file with geometry - how to get the proper projection parameters (lat/long)? We could supply them by hand, but this will just create errors, right? Maybe save them as columns in the Excel file?

I think we should just ignore this feature - the roundtrip to access the data can be done with the `dbf<-->xls` tools anyway. And the dashboard will provide an interface for editing these structures too. 

What do you think?

PR #1452, opened 2018-06-06
I1364 change final to sys

JIMENOFONSECA:
This solves a wide number of issues

@BhargavaKrishnaSreepathi woudl you mind reviewing it?

You will need to do the next for you scenario in the cooling case?

1. run data helper.
2. run radiation,
3. run demand,
4. run lca/main
5. run costs/operation_costs
5. run distributed optimization,
6. run optimization main
7. run the plots.


daren-thomas:
@JIMENOFONSECA Running `cea test` will reveal that the configuration section `embodied-energy` was removed - the embodied-energy script does not run anymore. That is why it is currently failing.

gabriel-happle:
@JIMENOFONSECA most of the demand plots are working.
I made a commit fixing some of the naming.
However, the `Total_demand.csv` contains empty columns about fuels, I guess that's the reason why some peak "fuel" plots don't work (no data to plot).

Empty columns are:
`OIL_hs0_kW`, `WOOD_hs_MWhyr`, `GRID0_kW`, `GRID_MWhyr`, ...

Let me know when I should test again.

PR #1456, opened 2018-06-08
Limit number of processes of multiprocessing

roglen:
Introduces limiting function to the total amount of processes used in multiprocessing, default is set to leave on CPU open. The maximum number of processes is now a user input into the config file.

### DIFF
--- a/cea/default.config
+++ b/cea/default.config
@@ -16,6 +16,11 @@ multiprocessing = true
 multiprocessing.type = BooleanParameter
 multiprocessing.help = Multiprocessing for quicker calculation, if available.
 
+max-processes = -1
+max-processes.type = IntegerParameter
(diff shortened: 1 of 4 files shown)


JIMENOFONSECA:
@roglen I believe this should only apply to the thermal network. We do not have problems in other scripts using multiprocessing and we risk to reduce performance overall.

roglen:
@JIMENOFONSECA It was actually a suggestion by @daren-thomas in the connected issue (i1450), main reason being consistency. 
For the demand it might be a relevant function once we have a large amount of buildings (e.g. 30) and a large amount of CPUs (e.g. the SEC server). In that case we might run out of RAM when trying to write the 30 files at once, so it might be useful to have an option of limiting the amount of CPUs

PR #1505, opened 2018-06-14
cea-workflow

daren-thomas:
This PR adds a file `bin/cea-workflow.bat` that can be used to run all the cea scripts in the workflow as used in the executive course in sequence, creating a file `%CEA-SCENARIO%/cea-workflow.log`.

In order to run this script, set the environment variables:

- CEA-SCENARIO
- CEA-REGION
- CEA-WEATHER

Example run (from command prompt):
```
set CEA-SCENARIO=c:\reference-case-WTP-reduced\WTP_MIX_m
set CEA-REGION=SIN
set CEA-WEATHER=Singapore-2016
call bin/cea-workflow.bat
```


JIMENOFONSECA:
@daren-thomas cool stuff, I see however the optimization part is there twice.

daren-thomas:
@JIMENOFONSECA that is by design. once with the smaller test (indivitualind/ngen=2) and once with it set to 5. should i just leave it at 2?

daren-thomas:
@JIMENOFONSECA ok. do you want to do the honors?

PR #1517, opened 2018-06-19
Setting random seed for optimization

BhargavaKrishnaSreepathi:
This PR sets up a seed for optimization. The default value is 64. This will allow us to reproduce the error encountered by anyone while running optimization (which was a problem before as the random generator gives rise to different options).

The PR passes local tests, runs for reference-case-open (ran three times and got the exact same individuals, as desired), runs for WTP-reduced/WTP_CBD_m (ran three times and got the exact same individuals).

This is the way to test it too, run the reference case open, and then compare the All_individuals.csv present in `scenario\outputs\data\optimization\slave`


BhargavaKrishnaSreepathi:
hey @daren-thomas, can you make this a priority? It will be very useful to get it done before the testing process (cea case studies, 30 pop and 25 gen )

cea/default.config
```diff
@@ -700,6 +700,10 @@ recoverycheckpoint = 0
 recoverycheckpoint.type = IntegerParameter
 recoverycheckpoint.help = in case the optimization stops, it can be resumed from this checkpoint.
 
+random-seed = 64
```
daren-thomas: I believe the default should be 0 which should be interpreted as not setting the random seed (or setting it to a number derived from "now" - and we print it out when we run a simulation so that if it fails, that can be used in the issue)

cea/optimization/master/generation.py
```diff
@@ -108,7 +107,10 @@ def generate_main(nBuildings, config):
 
         # Allocation of Shares
         def cuts(ind, nPlants, irank):
-            cuts = sorted(random_sample(nPlants - 1) * 0.99 + 0.009)
+            cuts = []
```
daren-thomas: no. let's keep using numpy for this. check out https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.random.seed.html - then you don't have to change your code. Call this in the main function of the optimization with the config variable.

cea/optimization/master/generation.py
```diff
@@ -170,7 +172,10 @@ def cuts(ind, nPlants, irank):
...
+            for i in range(nPlants - 1):
```
daren-thomas: see above.

cea/optimization/master/master_main.py
```diff
@@ -83,6 +84,7 @@ def evolutionary_algo_main(locator, building_names, extra_costs, extra_CO2, extr
     function_evals = 0
     euclidean_distance = 0
     spread = 0
+    random.seed(64)
```
daren-thomas: We introduced a configuration variable for this. And now we just hardcode it?

daren-thomas:
@BhargavaKrishnaSreepathi please address the points mentioned in the review before we can merge.

daren-thomas:
OK. I'll accept this as soon as I have uploaded a small change (set the default value for the random-seed to an empty string)

BhargavaKrishnaSreepathi:
@daren-thomas: a small doubt. Is it not better to have an explicit 0 instead of a null, where we are not sure on the value? Or am I interpreting it differently?

daren-thomas:
this passes local tests. i'm merging.

PR #1530, opened 2018-06-22
Fix to decentralized cooling, where the same name has been referenced for different technologies

BhargavaKrishnaSreepathi:
A quick fix to solve the bug present in decentralized cooling calculations


BhargavaKrishnaSreepathi:
It passed local tests

PR #1533, opened 2018-06-23
Add output of HEX sizing to Nodes.csv

roglen:
Adds size of substation HEXs at each building in [kW] to the _Nodes.csv output file.

### DIFF
--- a/cea/technologies/thermal_network/substation_matrix.py
+++ b/cea/technologies/thermal_network/substation_matrix.py
@@ -57,16 +57,21 @@ def substation_HEX_design_main(buildings_demands, substation_systems, thermal_ne
     t0 = time.clock()
 
     # Calculate disconnected buildings_demands files and substation operation.
-    substations_HEX_specs = pd.DataFrame(columns=['HEX_areas', 'HEX_UA', 'HEX_Cost'])
+    substations_HEX_specs = pd.DataFrame(columns=['HEX_areas', 'HEX_UA', 'HEX_Cost', 'HEX_Q'])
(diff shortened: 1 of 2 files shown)


JIMENOFONSECA:
@shanshanhsieh would you mind checking this? I imagine this should allow us to run the thermal network reduced and then run the network plots based on the results.

shanshanhsieh:
@JIMENOFONSECA, okay, i am on it!

shanshanhsieh:
@roglen how can i test this PR?

roglen:
@JIMENOFONSECA  @shanshanhsieh  
This PR adds additional columns to the nodes.csv, which contain information on the substation HEX sizing (in kW). 
This does not make plots possible for the reduced timesteps option. I don't think that it makes sense to allow plots with the representative_week set as true since:
1. Timesteps are cut together so any plots with dates/times on an axis do not make sense
2. Output files in this mode are made only to give information for the network optimization. So the number of outputs is reduced to reduce optimization runtime. 

In general, the representative week is meant to be run only with the optimization, in which case we don't want plots for each individual. 

Sorry for the confusion!

@JIMENOFONSECA The way I understood it, the only change should be the new output, not the possibility of plotting with reduced timesteps?

shanshanhsieh:
@roglen , i just tested the branch, it works fine.
could you also add the hex size of the plant in the same file?

roglen:
I am adding plant pressure losses and the output of plant HEX sizing to the nodes csv now. Having some internet problems so I hope the upload will work.

PR #1543, opened 2018-06-27
I1542 qcre takes 3 arguments

martin-mosteiro:
This PR solves four different minor issues:
1) since COOLROOM building types have Hs=0, the electricity demand for refrigeration was not calculated for this building type (which is the only type that has refrigeration by default)
2) the wrong number of arguments was passed to the function `refrigeration_loads.calc_Qcre_sys`
3) the wrong number of arguments was passed to the function `refrigeration_loads.calc_Qref`
4) the wrong label was used for a column in the DataFrame `energy_source` in `refrigeration_loads.calc_Qref`


martin-mosteiro:
This PR can wait until after the executive course since we don't have COOLROOM type buildings in the WTP case study, but it is necessary in order to run the other case studies correctly.

Issue #1570, opened 2018-07-05
Typo in Scenarios_occupancy_types legend


martin-mosteiro:
It currently says "Regrigeration rooms".

Issue #1569, opened 2018-07-05
Primary energy/emissions for embodied and commuting are off by a factor of 10


martin-mosteiro:
The operational primary energy and emissions are defined by GFA
The mobility and embodied energy are defined by 0.1*GFA

PR #1571, opened 2018-07-05
I1569 lca comparison plots

martin-mosteiro:
This PR solves an issue with the comparison plots, where the primary energy and emission intensity (i.e., per square meter) were calculated incorrectly.


martin-mosteiro:
This PR solves issues #1569 and #1570

PR #1592, opened 2018-07-26
I1588 bugfix aru model

gabriel-happle:
Fixes cooling system calculations for low outdoor temperature situations.


gabriel-happle:
This also seems to fix #1593

gabriel-happle:
@martin-mosteiro the annual cooling demand in the open case is affected a lot (more than 10% lower now). See the changes to `test_calc_thermal_loads.config`.
Is there a way you could check if that is OK?

gabriel-happle:
@daren-thomas any idea why this is failing the tests? I can successfully run `dodo` from pycharm and `cea test` from Anaconda Prompt.

martin-mosteiro:
Also worth pointing out: this control strategy causes the cooling systems to start and stop a lot moe often.
Master branch:
screen shot 2018-07-26 at 12 58 40
i1588:
screen shot 2018-07-26 at 12 59 19

martin-mosteiro:
With the following modification to the control strategy, the Hochschulquartier case study runs and the results for reference-case-open are the same as before:
```
def cooling_system_is_active(bpr, tsd, t):
    """
    Checks whether the cooling system is active according to rules for a specific hour of the year
    i.e., is there a set point temperature

    :param tsd: a dictionary of time step data mapping variable names to ndarrays for each hour of the year.
    :type tsd: dict
    :param t: hour of the year, simulation time step [0...8760]
    :type t: int
    :return: True or False
    :rtype: bool
    """

    if not np.isnan(tsd['ta_cs_set'][t]) and not tsd['T_int'][t - 1] <= np.max(bpr.hvac['Tc_sup_air_ahu_C'],
                                                                               bpr.hvac['Tc_sup_air_aru_C']):
        # system has set point according to schedule of operation & internal temperature is not below the set point
        return True
    else:
        return False
```

gabriel-happle:
Thanks for the testing @martin-mosteiro.
But it still doesn't pass the tests in jenkins...

JIMENOFONSECA:
@gabriel-happle do the test pass locally? maybe jenkins is down.

martin-mosteiro:
@JIMENOFONSECA @gabriel-happle The tests don't pass locally on my computer, but it seems to be an issue with my installation of HDF5
```
Warning! ***HDF5 library version mismatched error***
The HDF5 header files used to compile this application do not match
the version used by the HDF5 library to which this application is linked.
Data corruption or segmentation faults may occur if the application continues.
This can happen when an application was compiled by one version of HDF5 but
linked with a different version of static or shared HDF5 library.
You should recompile the application or check your shared library related
settings such as 'LD_LIBRARY_PATH'.
You can, at your own risk, disable this warning by setting the environment
variable 'HDF5_DISABLE_VERSION_CHECK' to a value of '1'.
Setting it to 2 or higher will suppress the warning messages totally.
Headers are 1.10.2, library is 1.10.1
```

JIMENOFONSECA:
I see, would you mind reverting to an older version and setting up that version in the setup.yml file of cea?

Best, Jimeno
On 2 Aug 2018, 17:48 +0800, Martín Mosteiro-Romero <notifications@github.com>, wrote:

@JIMENOFONSECA<https://github.com/JIMENOFONSECA> @gabriel-happle<https://github.com/gabriel-happle> The tests don't pass locally on my computer, but it seems to be an issue with my installation of HDF5

Warning! ***HDF5 library version mismatched error***
The HDF5 header files used to compile this application do not match
the version used by the HDF5 library to which this application is linked.
Data corruption or segmentation faults may occur if the application continues.
This can happen when an application was compiled by one version of HDF5 but
linked with a different version of static or shared HDF5 library.
You should recompile the application or check your shared library related
settings such as 'LD_LIBRARY_PATH'.
You can, at your own risk, disable this warning by setting the environment
variable 'HDF5_DISABLE_VERSION_CHECK' to a value of '1'.
Setting it to 2 or higher will suppress the warning messages totally.
Headers are 1.10.2, library is 1.10.1

BhargavaKrishnaSreepathi:
@JIMENOFONSECA @martin-mosteiro: i ran the tests locally on my computer for this PR. It runs successfully without any errors.

Issue #1597, opened 2018-08-02
pyliburo is replaced by py4design, thus the installation of the environment needs to take this change into account


BhargavaKrishnaSreepathi:
pyliburo is renamed as py4design, so the change needs to be made. Also, the folder structure in py4design is different than what it used to be for pyliburo. This impacts the radiation scripts in CEA. 

The problem is faced when we are creating the CEA environment afresh

JIMENOFONSECA:
@daren-thomas the best for this is to keep the old version of pyliburo. We do not know what else has changed, and it would be too much work to connect to the new version of the library. I think the problem is that a fresh installation of CEA is looking for pyliburo and it does not exist?

daren-thomas:
@JIMENOFONSECA technical debt - we've known about this for a while. yes, the old version suits our needs. and as I understand it, the installation should still be pulling it from our fork. I'll have a look and see what can be fixed.

daren-thomas:
actually, I think keeping the old pyliburo is harder than it seemed at first... unless... we can publish the fork as pyliburo, but i think we should ask @chenkianwee first before doing this.

@JIMENOFONSECA is @chenkianwee around in singapore? Would you mind asking him in person if we can publish our fork (in his name!) to PyPI?

PR #1599, opened 2018-08-03
Network optimization

roglen:
Very big PR with a bunch of new changes and added features for the optimization of DC networks. 

Most importantly adds:
- network optimization as described in my thesis & presentation & code comments
- changes to config file with a new section of thermal network optimization
- large new script with main optimization functions
- new constants related to optimization
- VCC COP calculations
- HEX cost calculations
- edits to steiner tree function, which allow user inputs to decide on which buildings to disconnect, suitable plant locations, etc. 
- edits thermal network matrix to be run from network optimization

To test: 
run a variety of cases in the different modes of optimization, e.g.
- with / without loop optimization
- with / without load optimization
- with / without connected building optimization
- with / without optimization of number of buildings
- with / without optimization of plant locations
- with / without rule based approach

A combined case running all decision variables at once has neot yet been tested successfully due to memory errors. Hence, a logical first step would be to implement checkpoints or think over the current data stroage structure of the optimization which uses 3 different dataframes, some of which may contain redundant values. Sadly I didn't find the time to do this.


daren-thomas:
This pull request passes the tests on my computer. The reason for not passing the jenkins tests is #1597 and not related to this pull request.

But I do get this when running `cea thermal-network-matrix`:

```
AttributeError: Parameter not configured to work with this script: {thermal-network-optimization:use-representative-week-per-month}
```

shanshanhsieh:
hey @roglen ,

i was only able to run `thermal_network_optimization.py` with the default config, and the other modes all failed for different reasons (this was tested in different computers).

Here are the errors i encountered:

1. `optimize-loop-branch = true`: 
this error showed up after the 4th generation
"""
  File "<string>", line 1, in <module>
  File "C:\Users\Admin\Anaconda2\envs\cea\lib\multiprocessing\forking.py", line 380, in main
    prepare(preparation_data)
  File "C:\Users\Admin\Anaconda2\envs\cea\lib\multiprocessing\forking.py", line 509, in prepare
    '__parents_main__', file, path_name, etc
  File "C:\Users\Admin\Documents\GitHub\CityEnergyAnalyst\cea\technologies\thermal_network\thermal_network_optimization.py", line 10, in <module>
    from cea.technologies.thermal_network import thermal_network_matrix as thermal_network_matrix
  File "C:\Users\Admin\Documents\GitHub\CityEnergyAnalyst\cea\technologies\thermal_network\thermal_network_matrix.py", line 16, in <module>
    import geopandas as gpd
  File "C:\Users\Admin\Anaconda2\envs\cea\lib\site-packages\geopandas\__init__.py", line 4, in <module>
    from geopandas.io.file import read_file
  File "C:\Users\Admin\Anaconda2\envs\cea\lib\site-packages\geopandas\io\file.py", line 3, in <module>
    import fiona
  File "C:\Users\Admin\Anaconda2\envs\cea\lib\site-packages\fiona\__init__.py", line 69, in <module>
    from fiona.collection import Collection, BytesCollection, vsi_path
  File "C:\Users\Admin\Anaconda2\envs\cea\lib\site-packages\fiona\collection.py", line 9, in <module>
    from fiona.ogrext import Iterator, ItemsIterator, KeysIterator
ImportError: DLL load failed: The paging file is too small for this operation to complete.
"""
2. `optimize-network-loads = true`
after generation 8, first the Importerror and then the memory error popped up
"""
  File "<string>", line 1, in <module>
  File "C:\Users\Shanshan\Miniconda2\envs\cea\lib\multiprocessing\forking.py", line 380, in main
    prepare(preparation_data)
  File "C:\Users\Shanshan\Miniconda2\envs\cea\lib\multiprocessing\forking.py", line 509, in prepare
    '__parents_main__', file, path_name, etc
  File "C:\Users\Shanshan\Documents\CityEnergyAnalyst\cea\technologies\thermal_network\thermal_network_optimization.py", line 10, in <module>
    from cea.technologies.thermal_network import thermal_network_matrix as thermal_network_matrix
  File "C:\Users\Shanshan\Documents\CityEnergyAnalyst\cea\technologies\thermal_network\thermal_network_matrix.py", line 16, in <module>
    import geopandas as gpd
  File "C:\Users\Shanshan\Miniconda2\envs\cea\lib\site-packages\geopandas\__init__.py", line 4, in <module>
    from geopandas.io.file import read_file
  File "C:\Users\Shanshan\Miniconda2\envs\cea\lib\site-packages\geopandas\io\file.py", line 3, in <module>
    import fiona
  File "C:\Users\Shanshan\Miniconda2\envs\cea\lib\site-packages\fiona\__init__.py", line 69, in <module>
    from fiona.collection import Collection, BytesCollection, vsi_path
  File "C:\Users\Shanshan\Miniconda2\envs\cea\lib\site-packages\fiona\collection.py", line 9, in <module>
    from fiona.ogrext import Iterator, ItemsIterator, KeysIterator
ImportError: DLL load failed: The paging file is too small for this operation to complete.
calculating thermal hydraulic properties of DC network  ...  time step 7
calculating thermal hydraulic properties of DC network  ...  time step 8
  File "C:/Users/Shanshan/Documents/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_optimization.py", line 941, in <module>
    main(cea.config.Configuration())
  File "C:/Users/Shanshan/Documents/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_optimization.py", line 865, in main
    sortedPop = network_cost_calculation(newMutadedGen, network_info)
  File "C:/Users/Shanshan/Documents/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_optimization.py", line 90, in network_cost_calculation
    fitness_func(network_info)
  File "C:/Users/Shanshan/Documents/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_optimization.py", line 319, in fitness_func
    thermal_network_matrix.main(network_info.config)
  File "C:\Users\Shanshan\Documents\CityEnergyAnalyst\cea\technologies\thermal_network\thermal_network_matrix.py", line 3402, in main
    thermal_network_main(locator, network_type, network_name, file_type, set_diameter, config, substation_systems)
  File "C:\Users\Shanshan\Documents\CityEnergyAnalyst\cea\technologies\thermal_network\thermal_network_matrix.py", line 533, in thermal_network_main
    repeat(region, times=(stop_t - start_t))))
  File "C:\Users\Shanshan\Miniconda2\envs\cea\lib\multiprocessing\pool.py", line 251, in map
    return self.map_async(func, iterable, chunksize).get()
  File "C:\Users\Shanshan\Miniconda2\envs\cea\lib\multiprocessing\pool.py", line 567, in get
    raise self._value
MemoryError: out of memory
"""
3. `optimize-building-connections = true`
here we have another arror
"""
  File "C:/Users/Shanshan/Documents/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_optimization.py", line 941, in <module>
    main(cea.config.Configuration())
  File "C:/Users/Shanshan/Documents/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_optimization.py", line 874, in main
    newMutadedGen = mutateGeneration(newGen, network_info)
  File "C:/Users/Shanshan/Documents/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_optimization.py", line 784, in mutateGeneration
    mutated_individual = list(mutateConnections(mutated_individual, optimal_network))
  File "C:/Users/Shanshan/Documents/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_optimization.py", line 611, in mutateConnections
    random_index = index[0]
IndexError: list index out of range
"""

please test the runs mentioned above and fix the problems, thanks!

roglen:
hey @shanshanhsieh !
Errors number 1 and 2 both seem to be connected to memory errors as far as I can tell. I believe that they will both be solved by implementing checkpoints - and making sure that memory isn't overloaded as much as it currently is. One way I could imagine of doing that is by saving individuals to file, or at least each generation? I left this as one of the big "To Dos" because I ran out of time.

For error number 3 I will have a look in the next days/week (currently on holiday in England), but it seems like a smaller bug. I think I've seen this before. I'll let you know when I fix it!

shanshanhsieh:
@roglen thanks for your fast response!
However, i am not convinced that this branch can be merged if the memory errors persist.
The minimum requirement would be that one can perform all the tests you mentioned in this PR (not with the combination) and use them to reproduce the results from your report.

Maybe saving the results of each individual could be an easy solution. From my understanding, it is currently saving after each generation, right? if needed, we could set up a meeting with @BhargavaKrishnaSreepathi regarding what is the best way to organize the data generated from optimization.

Here are the test results from my side:
1. with / without loop optimization - unsuccessful
2. with / without load optimization - unsuccessful
3. with / without connected building optimization - unsuccessful
4. with / without optimization of number of buildings - unsuccessful
5. with / without optimization of plant locations (1 & 2 plants) - successful, but it seems like there's no option to do the "without"
6. with / without rule-based approach: on-going

cea/technologies/thermal_network/thermal_network_costs.py
```diff
@@ -136,30 +135,41 @@ def calc_Ctot_cooling_plants(network_info):
...
-            disconnected_demand = {}
-            for building in network_info.building_names:
-                disconnected_demand[building] = pd.read_csv(
-                    network_info.locator.get_demand_results_file(building))
-            for t in range(HOURS_IN_YEAR):
-                # calculate COP of plant operation in this hour based on supplied loads
-                # calculate plant COP according to the cold water supply temperature in SG context
-                supplied_systems = find_non_zero_demand_systems(network_info, t, disconnected_demand)
+            if network_info.config.yearly_cost_calculations:
+                supplied_systems = find_systems_string(network_info.full_cooling_systems)
```
shanshanhsieh: why not using line 158-160 to have the exact supplied_systems?

cea/technologies/thermal_network/thermal_network_costs.py
```diff
@@ -0,0 +1,563 @@
...
+        Opex_fixed_chiller = 0
+        Capex_CT = 0
+        Opex_fixed_CT = 0
+        if plant_heat_peak_kW > 0:  # we have non 0 demand
+            peak_demand_W = plant_heat_peak_kW * 1000  # convert to W
+            print 'Calculating cost of heat production at plant number: ', plant_number
+            if network_info.config.thermal_network_optimization.yearly_cost_calculations:
+                # calculates operation costs with yearly approximation
+                COP_plant = VCCModel.calc_VCC_COP(network_info.config,
+                                                  network_info.full_cooling_systems,
```
shanshanhsieh: why not using line 164 to get the exact supplied systems?

cea/technologies/thermal_network/thermal_network_costs.py
```diff
@@ -0,0 +1,563 @@
...
+                                                  centralized=True)
+                Opex_var_chiller += abs(
+                    plant_heat_yearly_kWh) / COP_plant * 1000 * network_info.prices.ELEC_PRICE
+                Opex_var_CT += CTModel.calc_CT_yearly(plant_heat_yearly_kWh)
```
shanshanhsieh: This model is problematic:
1. the heat input should be the heat rejected from the condenser, not the cooling production from plants
2. there is no bounds for this equation

cea/technologies/thermal_network/thermal_network_costs.py
```diff
@@ -0,0 +1,563 @@
...
+        Capex_chiller = 0
+        Opex_fixed_chiller = 0
+        Capex_CT = 0
+        Opex_fixed_CT = 0
+        if plant_heat_peak_kW > 0:  # we have non 0 demand
+            peak_demand_W = plant_heat_peak_kW * 1000  # convert to W
+            print 'Calculating cost of heat production at plant number: ', plant_number
+            if network_info.config.thermal_network_optimization.yearly_cost_calculations:
+                # calculates operation costs with yearly approximation
+                COP_plant = VCCModel.calc_VCC_COP(network_info.config,
```
shanshanhsieh: After checking the source of the numbers, this COP_plant already include opex_var for chiller, cooling tower, condenser water pumps and chilled water pumps? (please confirm)
Therefore, the equation calculating Opex_var_CT in line 152 is redundant. (sorry, there were probably some misunderstanding when i first requested the change)

If you can confirm that, the output of this function `calc_Ctot_cooling_plants` should be: `Opex_a_cooling_plants`, `Capex_a_CT`, `Capex_a_chiller`.
and `Opex_a_cooling_plants` = `Opex_var_plant` + `Opex_fixed_CT` + `Opex_fixed_chiller`

cea/technologies/thermal_network/thermal_network_costs.py
```diff
@@ -0,0 +1,563 @@
...
+                Opex_var_CT += CTModel.calc_CT_yearly(plant_heat_yearly_kWh)
+
+            else:
+                # calculates operation costs with hourly simulation
+                # Read in building demand
+                building_demand = {}
+                for building in network_info.building_names:
+                    building_demand[building] = pd.read_csv(
+                        network_info.locator.get_demand_results_file(building))
+                for t in range(HOURS_IN_YEAR):
```
shanshanhsieh: I think the entire hourly function is redundant since the current model calculating COP_plant does not change over each hour... but we can keep this structure as a placeholder for future addition.
Please:
1. delete Opex_var_CT (since it is already included in COP_plant, also, the heat input is not correct as explained in previous comment)
2. make note that the current model does not give different results when choosing hourly calculation/ or delete the entire session

cea/technologies/thermal_network/thermal_network_costs.py
```diff
@@ -0,0 +1,563 @@
...
+                            # go through all systems and sum up demand values
+                            if system_index == 0:
+                                disconnected_demand_t = disconnected_building_demand[system]
+                            else:
+                                disconnected_demand_t = disconnected_demand_t + disconnected_building_demand[system]
+                        peak_demand_kW = disconnected_demand_t.abs().max()  # calculate peak demand of all disconnected systems
+                        disconnected_demand_t_sum = disconnected_demand_t.abs().sum()
+                    print 'Calculate cost of disconnected loads in building ', building
+                    if network_info.config.thermal_network_optimization.yearly_cost_calculations:
+                        COP_chiller_system = VCCModel.calc_VCC_COP(network_info.config, system_string,
```
shanshanhsieh: please confirm this is correctly implemented (see FIXME in line 245)

shanshanhsieh:
@roglen update to the memory errors: thanks to @BhargavaKrishnaSreepathi , who implemented the saving of each individual, this should resolve the problem. testing on this part is ongoing now.

roglen:
@shanshanhsieh I'll have a look this weekend and try to clear things up and implement the changes. Thanks for the feedback! And thank you very much @BhargavaKrishnaSreepathi for helping with the memory error

cea/technologies/thermal_network/thermal_network_costs.py
roglen: I think this has resolved itself as the cose has chenged since the comment? At least I can't ind this section anymore.

cea/technologies/thermal_network/thermal_network_costs.py
roglen: My reasoning was that since this is the yearly calculation we just take the inputs, and the funciton in the hourly calculation evaluates every hour. But I could adapt the 'find_non_zero_demand_systems' hourly function from line 164 to evaluate the aggregated yearly values. Then I could use the function here too.

cea/technologies/thermal_network/thermal_network_costs.py
roglen: 1. Ok I see the problem, but without a model of the plant I'm not sure how to solve this, except maybe by trying to find a factor which relates plant heat to condenser heat? Some kind of plant condenser efficiency?

2. The CT function has an upper bound in the sense that it will just add more cooling towers, lower bound should be 0. But I think I am misunderstanding what you mean :) 

Please advise on how to proceed on these two points :)

cea/technologies/thermal_network/thermal_network_costs.py
roglen: I checked the source and you are right - the calculation is based on values which already contain the CT and chiller opex. I am implementing the changes as you suggested.

cea/technologies/thermal_network/thermal_network_costs.py
roglen: It seems like this is actually no longer relevant since the CT OPEX is included in the OPEX var chiller calculation.

cea/technologies/thermal_network/thermal_network_costs.py
roglen: What about the Opex_var_CT? The way I understand this, it is also included in the Opex_a_cooling_plants. 
Then it would be:
´Opex_a_cooling_plants = Opex_var_plant + Opex_fixed_CT + Opex_fixed_chiller + Opex_var_CT´

cea/technologies/thermal_network/thermal_network_costs.py
roglen: Problably this is redudant as Opex_var_CT seems to be included in the Opex_cooling_plant calculation

cea/technologies/thermal_network/thermal_network_costs.py
roglen: 1. Done :)

2. Actually it should give diferent restults because the supplied systems change depending on if we have an hourly or yearly calculation. The yearly function that I will add in my next commit checks not only if a system is "connected" but also if it has a non zero demand over the year. If the demand is always 0 then this system is excluded for the COP calculation. 

In the hourly calculation this is evaluated at each time step. So our system will have a different COP depending on which loads have nonzero demands. in the yearly calculation the COP is the same for the entrie year.

cea/technologies/thermal_network/thermal_network_costs.py
roglen: Good catch, now it is correct

cea/technologies/thermal_network/thermal_network_costs.py
roglen: I am adding this function in the next commit

cea/technologies/thermal_network/thermal_network_costs.py
shanshanhsieh: It depends on what is included in COP_plant
from my understanding, the `Opex_var_plant` calculated from COP_plant includes the operation cost of cooling towers as well, that means we can treat `Opex_var_plant` = `Opex_var_chiller` + `Opex_var_CT`

cea/technologies/thermal_network/thermal_network_costs.py
shanshanhsieh: So the COP of the current hourly calculation only changes if the supplied systems are different between each hour, is that right? 
It is independent of the outdoor temperature at the moment?
That is okay for me, since the model is quite rough for the moment anyways, however, it should probably be documented.

shanshanhsieh:
@roglen similar problems also exist in the decentralized chiller and cooling tower calculation. please revise that part as well.

cea/technologies/thermal_network/thermal_network_costs.py
roglen: I agree!

cea/technologies/thermal_network/thermal_network_costs.py
roglen: Currently in both the hourly and the yearly case the COP is a function of t`he yearly average outdoor temperature (in the calc_VCC_COP function).

For the hourly function I will try to clarify:
The supplied systems which are either defined by the optimization or given as a user input from the config file are the systems which we supply with a centralized system. The yearly function checks if one of these systems has an aggregated demand of zero over the year - if it does, the COP is adapted to ignore that system. 
The hourly function does this for every hour. So depending on the demand of that hour, the COP will change. 
E.g. supplied systems = ahu, aru:
t = 10 we have demand for ahu and aru, so the COP is calculated using ahu and aru
t = 11 we only have ahu demand, so COP is calculated using ahu only
t = 12 we only have aru demand, so COP is calculated using aru only
... etc.

I will clarify this in the function description

cea/technologies/thermal_network/thermal_network_costs.py
roglen: @shanshanhsieh 
This seems to be the only point which is still open - the CT Capex is still being calculated based on sizing of the CT to the heat demand, not to the heat rejected from the condenser. Any ideas how to fix this?

roglen:
@shanshanhsieh I made the changes discussed above except for the CT capex which is still being sized based on the heat demand and not the condenser exhaust heat. On this point I am not sure how to proceed. The other points should all be covered. 
As far as i can tell 3 issues remain:
1. CT capex
2. I am still getting a memory error if I run e.g. 10 generations with 10 individuals
3. Github checks are failing
FYI My time as officialy member of ETH is ending in one week, so at that point I will no longer be able to access ARCGis or run the scripts since they need the ARCGis interface.

shanshanhsieh:
hi @roglen, the changes seem good
for your questions:
1. i implemented the changes, basically, a rough calculation of how much heat is rejected from the condenser base on COP of chillers, please go through the code with label `FIXME` to confirm the units are right
2. if you have time, would you like to set up a quick meeting with @BhargavaKrishnaSreepathi tomorrow to go through the potential bottlenecks for the memory error?

roglen:
@shanshanhsieh the changes look good. I'm setting up a meeting with Bhargav this week to look at the memory error.

roglen:
@BhargavaKrishnaSreepathi As discussed I looked through the code for possible sources of the memory error and made an overview of the existing dataframes and dictionaries:

- population performance 
(dictionary, small, generated in thermal_network_optimization line 71, one entry per individual, filled in line 103 of thermal_network_optimization)
- outputs 
(Dataframe [n_individuals_per_generation x 24], Generated in thermal_network_optimization line 74ff, filled in line 153ff of thermal_netowrk_optimization using network_info.populations)
- network_info.cost_storage 
(Dataframe [n_individuals_per_generation x 18], generated in thermal_network_optimization line 849, filled in line 461ff of thermal_network_costs)
- network_info.populations 
(Dictionary of dictionaries, one dictionary per individual, generated in thermal_network_optimization line 88, filled with info from network_info.cost_storage in line 109ff of thermal_network_optimization)
- network_info.all_individuals 
(Dataframe [n_individuals_total x 24], generated in thermal_network_optimization line 878, filled with data from network_info.populations in thermal_network_optimization lines 893ff, written to csv in line 931) This is only used at the very end of the optimization so this is not the source of the error

Movement of Data: 
cost_storage -> populations -> outputs & all_individuals 
Probably only one Dataframe or dictionary would be necessary to do this process
(e.g. network_info.populations could maybe replace cost_storage and outputs)

Online I found that this error often comes from reading in huge files, I've marked possible large csv files that are read in with (!)
possible sources could be:
thermal_network_costs.py:
- line 85 read in edge mass flows
- line 90 read in pressure loss
- line 115 read in plant heat requirement
- line 141 read in building demand (!!) -> this generates a huge dictionary approx 65 MB with the building demands of each building, even though only some columns are actually needed
- line 235 read in building demand (!) reads in full file, only needs some columns
- line 374 read in building demand (!) reads in full file, only needs some columns
- line 527 read in network pipe information
- line 554 read in total demand file (!)

thermal_network_optimization.py:
- line 477 read in total demand file (!), only need two columns
- line 837 read in total demand file (!), only use column names (!) -> seems like a big waste of RAM

shanshanhsieh:
@BhargavaKrishnaSreepathi this branch is ready to test now!
To test:
run a case in SG with different variables:
- with/ without loop optimization
- with / without connected building optimization
- with / without optimization of number of plants
- with / without rule based approach

daren-thomas:
@BhargavaKrishnaSreepathi @shanshanhsieh what is the status of this branch? Any chance to merge it before the review meeting tomorrow?

shanshanhsieh:
since i made a lot of changes in this branch, i am no longer qualified as a reviewer for this PR
so @BhargavaKrishnaSreepathi is currently testing this branch, and it will take a couple more days until the testing is finished, my estimation is that we might be able to merge it next monday.

daren-thomas:
@BhargavaKrishnaSreepathi what is the status of your reviews?

BhargavaKrishnaSreepathi:
I finished running all the cases on this PR. The results need to be compared with Shanshan's. Currently she is on vacation, once she is back, we will check the results and then approve this.

shanshanhsieh:
Hi @BhargavaKrishnaSreepathi ,

i just updated the branch from master and here's the results from my runs, could you please compare the results from the last two rounds?
https://drive.google.com/open?id=1spEckV4_O8dh3ZtI8aMRnPLchQ1ozTlb

please go ahead to merge it if the results match, thanks!

JIMENOFONSECA:
@BhargavaKrishnaSreepathi status?

BhargavaKrishnaSreepathi:
This PR is ready to be merged, the results look good, just that the optimization parameters need to be tuned based on the case study, which is a separate research topic as such

JIMENOFONSECA:
Great, go ahead and merge :-)

_______________________

Dr. Jimeno A. Fonseca
Senior researcher and project coordinator
Future cities Laboratory
Singapore ETH-Centre

Chair of Architecture and Building Systems
ETH Zurich

Product owner
City Energy Anlayst (CEA)<http://www.cityenergyanalyst.com>

Prof.h.c in modeling and simulation
Universidad Central de Colombia
On 12 Feb 2019, 10:14 +0800, Bhargava Krishna Sreepathi <notifications@github.com>, wrote:

This PR is ready to be merged, the results look good, just that the optimization parameters need to be tuned based on the case study, which is a separate research topic as such

PR #1602, opened 2018-08-07
checked and changed zone geometry dbf to reflect variable names

Jack-Hawthorne:


daren-thomas:

PR #1603, opened 2018-08-13
remove pyliburo from conda env - it will be installed with pip

daren-thomas:
I'm able to create a conda environment from this on my computer. Let's see what the jenkins thinks about this. (`setup.py` contains a hint to collect pyliburo from our fork instead of kian we's source)


JIMENOFONSECA:
@daren-thomas should it pass jenkins? or is it good to go?

daren-thomas:
it is failing. and nope. not good to go. sorry. actually, it seems it won't ever work. can you talk to @chenkianwee? (see issue #1597).

in fact. it is my opinion that the best way would be to update the few places (create_gml, radiation_daysim) that require pyliburo to use py4design instead). the other option would be to host our fork on pypi as pyliburo. and forgo any bugfixes.

another alternative would be to extract the code from pyliburo that we actually use and incorporate that into CEA, making sure to pay attention to licenses and credits. for this I believe we should talk to @chenkianwee too first.

JIMENOFONSECA:
I go for using the fork. The option of looking at py4design code can invlove several hours of re-testing all functionality, and now that finally after about 1.5 years we have an stable daysim connection, this seems like a potential high risk to take.

I just sent a message to Kianwee, let's see what is his opinion.

JIMENOFONSECA:
@daren-thomas I just heard from KianWee, he is ok with us doing the fork into Pypi. Should we go ahead with this?

daren-thomas:
@JIMENOFONSECA ok. we can do that. BTW: what about just adding the parts we use to the CEA codebase? is there a way to get kianwee (he's already a cea developer, right?) to do that? That way the code would still be attributed to him, but we could drop the dependency. Otherwise, I'll just publish the fork.

JIMENOFONSECA:
Kianwee, would you like to contribute to this :-))? Although you have other type of license right?
On 15 Aug 2018, 16:54 +0800, Daren Thomas <notifications@github.com>, wrote:

@JIMENOFONSECA<https://github.com/JIMENOFONSECA> ok. we can do that. BTW: what about just adding the parts we use to the CEA codebase? is there a way to get kianwee (he's already a cea developer, right?) to do that? That way the code would still be attributed to him, but we could drop the dependency. Otherwise, I'll just publish the fork.

daren-thomas:
Ah... @JIMENOFONSECA I just saw. Yes, @chenkianwee used the GPL. That's actually a bit of a problem. We can't really use his code in our code without complying with the GPL (which requires us to adopt that license, I think) so... 

ugh. this here: https://opensource.stackexchange.com/questions/1640/if-im-using-a-gpl-3-library-in-my-project-can-i-license-my-project-under-mit-l

It seems more complicated than that. I wonder, @chenkianwee is there a way to use a different license for the pyliburo? The GPL is a copy left license and it's really hard to figure out what that means for our licensing. Since pyliburo / py4design is a _library_, maybe the LGPL license would be a better idea? I don't know. We can have that discussion if you like.

JIMENOFONSECA:
@daren-thomas i wonder if the commit fixing fiona and dependencies will produce an error once another library tries to get a newer version of one of fiona's dependencies.

For instance, installing geopandas will need pandas 19.0. Then when I install plotly, it requires pandas 20.0. The installation of plotly will update pandas and then make geopandas brake. Is this behavior expected?

daren-thomas:
@JIMENOFONSECA I finally figured out the library situation. Now... for the tests to pass, I had to uncomment the calibration tests. They don't reliably work and I don't know why.

This is the error message:

```
TaskError - taskid:run_calibration
PythonAction Error
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\ceatest\lib\site-packages\doit\action.py", line 383, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\ceatest\lib\site-packages\cea\tests\dodo.py", line 366, in run_calibration
    calibration_main.calibration_main(locator=locator, config=config)
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\ceatest\lib\site-packages\cea\demand\calibration\bayesian_calibrator\calibration_main.py", line 81, in calibration_main
    mu = pm.Deterministic('mu', predict_y(var1, var2, var3, var4, var5, var6))
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\ceatest\lib\site-packages\theano\gof\op.py", line 674, in __call__
    required = thunk()
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\ceatest\lib\site-packages\theano\gof\op.py", line 892, in rval
    r = p(n, [x[0] for x in i], o)
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\ceatest\lib\site-packages\theano\compile\ops.py", line 555, in perform
    outs = self.__fn(*inputs)
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\ceatest\lib\site-packages\cea\demand\calibration\bayesian_calibrator\calibration_main.py", line 58, in predict_y
    prediction = emulator.predict(input_sample)
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\ceatest\lib\site-packages\sklearn\gaussian_process\gpr.py", line 297, in predict
    X = check_array(X)
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\ceatest\lib\site-packages\sklearn\utils\validation.py", line 422, in check_array
    _assert_all_finite(array)
  File "C:\Users\darthoma\AppData\Local\conda\conda\envs\ceatest\lib\site-packages\sklearn\utils\validation.py", line 43, in _assert_all_finite
    " or a value too large for %r." % X.dtype)
ValueError: Input contains NaN, infinity or a value too large for dtype('float64').
```

(mentioned in #1609)

environment.yml
```diff
@@ -16,60 +16,48 @@ dependencies:
...
 - ephem=3.7.6.0=py27_1
 - expat=2.2.5
-- fiona=1.7.11
 - freetype=2.7=vc9_1
 - freexl=1.0.2=vc9_2
 - functools32=3.2.3.2=py27_1
 - geos=3.6.2
-- gdal=2.2.4
-- geopandas=0.3.0
+- geopandas
```
JIMENOFONSECA: @daren-thomas I prefer we install the exact version of every library to avoid compatibility issues.

environment.yml
```diff
@@ -16,60 +16,48 @@ dependencies:
...
 - libspatialite
+- libssh2=1.8.0=h0990ea7_2
 - libtiff
 - libxml2=2.9.4=vc9_4
-- matplotlib=2.0.2=py27_2
 - munch=2.2.0=py27_0
 - openjpeg=2.3.0
 - openpyxl
-- openssl=1.0.2l=vc9_0
-- pandas=0.20.3=py27_1
```
JIMENOFONSECA: @daren-thomas I prefer we install the exact version of every library to avoid compatibility issues. 

If we leave it open, a new library we install will update the version of one of these libraries without us realizing and then break compatibility with another library. is that right? maybe I am missing something.

JIMENOFONSECA:
@daren-thomas besides my comments, all tests pass.

environment.yml
daren-thomas: yes and no. the exact version will eventually not be hosted anymore. that is what happened this time. so either way, things will break. that is why we do tests - we install a fresh environment everytime a pull request is made. and we try to exercise as much code as possible. this is why writing these tests are so important and why i'd like more help with that. so while we can fix the versions exactly, we don't really remove the problem. how would you like to do it?

JIMENOFONSECA:
Got you. I will go for keeping the specific version + the test on every PR you suggest.

I guess It would be easier to see if the problem is of hosting than if the problem is about one library overwriting another one in order to work.

Is it possible that we add a restriction that any new library cannot change the version stored in the environment.yml?

Best, Jimeno
On 28 Aug 2018, 20:51 +0800, Daren Thomas <notifications@github.com>, wrote:

@daren-thomas commented on this pull request.

In environment.yml<https://github.com/architecture-building-systems/CityEnergyAnalyst/pull/1603#discussion_r213299834>:

>  - munch=2.2.0=py27_0
 - openjpeg=2.3.0
 - openpyxl
-- openssl=1.0.2l=vc9_0
-- pandas=0.20.3=py27_1


yes and no. the exact version will eventually not be hosted anymore. that is what happened this time. so either way, things will break. that is why we do tests - we install a fresh environment everytime a pull request is made. and we try to exercise as much code as possible. this is why writing these tests are so important and why i'd like more help with that. so while we can fix the versions exactly, we don't really remove the problem. how would you like to do it?

daren-thomas:
> Is it possible that we add a restriction that any new library cannot change the version stored in the environment.yml?

I'm not quite sure what you mean here... I guess what you mean is, let's not add new libraries willy nilly, but instead make sure we really need it? I expect that at some point we will want to add a library and that will require updating other libraries and I really don't see that as much of a problem - freezing the dependencies is probably not a good move. At the same time, keeping a certain amount of stability _is_ a good idea. I'll update the pull request with the current version of the environment.yml.

JIMENOFONSECA:
@daren-thomas got it, let me know when it is ready for testing

daren-thomas:
@JIMENOFONSECA I think it should work now. Please test :-)

Issue #1607, opened 2018-08-23
As a CEA developer I want to know how to contribute so I make my code readable and transparent for others to use.


JIMENOFONSECA:
The idea is to create a readme on how to contribute to CEA.

This should include ground rules about:

- Coding (not exactly PEP8, but more on how to create functions etc.)

- Variable Naming (not exactly PEP8, but more on how to understand the physics-based variables that CEA uses.

- How to create issues (or user stories). I will come with more of that in the next period.

PR #1615, opened 2018-09-08
This is a first stab to updating the documentation

JIMENOFONSECA:
@daren-thomas  Could you please verify this? I think this moves around parts of your work . Thanks!

This could be a first stab to the user story #1607 

I am missing just the part about how to name variables.

@BhargavaKrishnaSreepathi are we ready to include the new variable naming format? If not, could we reach an agreement asap on this?


docs/developer-walk-through.rst
```diff
@@ -1,140 +1,14 @@
...
-======================
+Other stuff
+===========
 
 .. toctree::
    :maxdepth: 2
    :caption: Contents
 
-   main-developer-workflow
-   how-to-add-a-new-script-to-the-cea
```
daren-thomas: I think this chapter is pretty important for new developers (or as a reference for older developers) - where is it added again?
(speakig of `how-to-add-a-new-script-to-the-cea`)

maybe we can add these in a sort of knowledge-base style catalog?

docs/developer-walk-through.rst
```diff
@@ -1,140 +1,14 @@
...
    the-configuration-file
    configuration-file-details
-   how-to-create-a-new-release
```
daren-thomas: where is this going? we kind of still need this information, but totally ok with moving it around.

docs/environment.yml
```diff
@@ -1,118 +1,208 @@
-name: cityenergyanalyst_docs
```
daren-thomas: I don't understand why you're changing this. Have you tested it to work on the readthedocs? I'd prefer this to be a separate issue (updating the environment.yml for read-the-docs) - the RTD build system is not a windows system and therefore the environment won't have the exact same versions (your changes add vc9_1 versions of some modules which don't exist in ubuntu as the compiler doesn't exist).

docs/how-to-add-a-new-script-to-the-cea.rst
```diff
@@ -1,16 +1,15 @@
-Contributing
-=============
+How to add a new script
```
daren-thomas: ah. i think this explains my previous question :-)

docs/how-to-add-a-new-script-to-the-cea.rst
```diff
@@ -1,16 +1,15 @@
...
 So you want to extend the CEA? This guide will get you up and running!
 
 The main steps you need to take are:
 
-#. copy the template script and rename it
-#. update the module-level documentation and credits
-#. develop your script
+#. start with a script template.
+#. develop your script.
 #. add your script to the ``cli.config`` file
```
daren-thomas: to be changed when the scripts.yml file changes... i can manually update no problem.

docs/how-to-contribute.rst
```diff
@@ -0,0 +1,85 @@
...
+
+The main steps you need to take are:
+
+#. Create an User Story or issue in Github.
+#. Create branch and learn how to use Github.
+#. get acquainted with variable naming.
+#. create your contribution.
+#. test your creation.
+#. Create a pull request on Github.
+#. Claim your CEA T-shirt!
```
daren-thomas: yay! t-shirts for everyone!

docs/developer-walk-through.rst
JIMENOFONSECA: yes it is kept already but re-organized

docs/developer-walk-through.rst
JIMENOFONSECA: yes it is kept already but re-organized

docs/how-to-add-a-new-script-to-the-cea.rst
JIMENOFONSECA: :-)

docs/environment.yml
JIMENOFONSECA: oh got you. I need your help with making this run

docs/how-to-add-a-new-script-to-the-cea.rst
JIMENOFONSECA: please do so, thanks!

docs/how-to-contribute.rst
JIMENOFONSECA: :-D

JIMENOFONSECA:
@daren-thomas so the PR is here since 24 days, we kind of need it to pass for Bhargava to include the new standard of variables. Would you mind checking the issue with the environment and get this baby up and running?

daren-thomas:
@JIMENOFONSECA sure. On the same page. I was just waiting for your comments ;-) (I'll have time later this week)

PR #1627, opened 2018-09-24
1224 improve runtime for solar collectors

daren-thomas:
This PR does some (minor) refactoring to the solar-collector code. One of the things realized was changing the type of many variables from np.ndarray with 1 dimension to np.array. The main boost comes with using the just-in-time compilation decorator from numba.

to use this, you either need to update the conda environment or `conda install numba` - (the `environment.yml` has been updated as part of the pull request).

@shanshanhsieh can you give this a spin and tell me what you think?


daren-thomas:
@shanshan are you working on this today? otherwise i could see myself applying some more changes, possibly adding a multiprocessing option... Also, I thought we should probably also add these changes to the PVT code?

shanshanhsieh:
@daren-thomas , i will not be working on this today, please go ahead and make the changes!
and you are right, the same changes can be applied to PVT!  (it was not mentioned last time since the operating temperature is lower, thus the simulation converges faster.)

daren-thomas:
@shanshanhsieh the PVT part still needs some work. i'll look into it on monday. but you could go ahead and see if the solar-collector stuff is working: if it's rubbish, then i don't have to create more rubbish with the PVT part...

daren-thomas:
@shanshanhsieh the pull request now also includes multiprocessing support :-) give it a spin and write a review (after adding the comments in the source of solar-collector and photovoltaic-thermal regarding the loop variable and the fortran code)

daren-thomas:
@shanshanhsieh ok. then we merge :)

PR #1629, opened 2018-09-25
1625 figure out why the radiation daysim tool hangs on sabines scenario

daren-thomas:
This pull request only contains some minor cosmetic changes to the geometry_generator script. Mainly, it adds a line printed out for each building being processed (e.g. `Processing district building Z1832`) so that when an error does occurr, it is easier to spot where.

Also, it allows using POLYGON shapefiles with a Z component - the Z component is ignored and the usual process for determining the Z component of a building footprint is used instead.


daren-thomas:
This also includes a fix for the dbf tool: apparently we were restricting `str` fields to 25 characters... changed that to 255 characters.

gabriel-happle:
@daren-thomas what should I test here?

daren-thomas:
@gabriel-happle basically that everything still works. Especially: The DBF conversions. So... The issue that was linked here has been resolved, but the fixes pushed are interesting to add to master, as they cover some edge cases we hadn't tested before.

How about testing this: Create an age.xlsx file with the `Name` column having very long names... larger than 25 characters. Roundtrip (xlsx -> dbf -> xlsx) and see if the names get "cut off" / shortened to 25 characters. Master currently does this. This one fixes it.

I'm not sure about POLYGONZ input files... this PR also allows using them - but in a very "brain-dead" fashion - they're just treated as POLYGON shapefiles without a Z-component.

daren-thomas:
@gabriel-happle I have updated the way the string lengths are calculated (instead of assigning 255 characters to each string field - that was creating errors with the way pysal reads dbf files).

I suggest we move the last two points of your review to separate issues:

- #1651 As a Practitioner, I'd like to be able to create *.xslx files from *.dbf files
- #1652 As a Practitioner, I'd like to produce Excel files from DBF files with an index of my choice

daren-thomas:
@gabriel-happle I think, if you can re-check the output please, that we can merge. I've updated from master and resolved the merge conflict.

gabriel-happle:
Works great.
I was able to test the PloygonZ conversion.
I used the process described here https://support.esri.com/en/technical-article/000012075 but in *reverse* to create a shape file with Z coordinate values.

PR #1655, opened 2018-10-08
I1619 calibration

martin-mosteiro:
This issue fixes minor issues in the calibration scripts that however caused the scripts not to run. These are, namely
- The number of variables requested by `calibration_sampling.py` was 5, but the number required by `calibration_main.py` was 6. Now both take 6 varibales.
- The triangular distribution in `calibration_main.py` was being assigned the wrong values, such that c > upper_bound for all variables. This has been corrected.
The scripts now run successfully on my computer.


daren-thomas:
@martin-mosteiro in the process of reviewing this - I am running the code as we speek. It's gonna take a while for the calibration-sampling script to run through.

Issue #1645, opened 2018-10-03
As a Contributor, I'd like to find a list of the Roles we can use for issues


daren-thomas:
This list should be part of the Documentation on RTD, either part of the [Contributing](https://city-energy-analyst.readthedocs.io/en/latest/how-to-add-a-new-script-to-the-cea.html) section or a separate section.

gabriel-happle:
I'm also wondering the same.

PR #1669, opened 2018-11-07
adding the 'roles' work

JIMENOFONSECA:
this fixes issue #1645 @daren-thomas would you mind checking this small PR?, thanks!


daren-thomas:
@JIMENOFONSECA thank you for this! I took the liberty of fixing some language issues I found - please check them.

Also, I think it might be important to differentiate user stories from bugs. In this document, we treat them as the same (bug being a subclass of user story) but actually, I think they are totally different. A user story describes a new feature that a user needs to accomplish something, whereas a bug describes an existing feature (that would ideally have been developed as part of a user story) that does not fill the need it should, as in, it does not accomplish what it should be.

If we do not make this distinction, we will end up with a ton of "user stories" that follow this pattern:

> as a (any user) i want to run feature (x) without it crashing / producing wrong results

A bug report has the following elements:

1. Steps to reproduce,
2. What you expected to see, and
3. What you saw instead.

See [Joel Spolsky's painles bug tracking article](https://www.joelonsoftware.com/2000/11/08/painless-bug-tracking) for a nice write-up on this class of issues.

I suggest we do not treat bug as a subclass of user story, but rather treat it as it's own category. Maybe we add the label `bug` on it? I suggest, besides the above, we could specify the following:

1. Steps to reproduce
  - include CEA version
  - include information about the case study / scenario and the configuration (input variables) you are using
2. What you expected to see
3. What you saw instead.
  - include full error message (including stack trace) if applicable

JIMENOFONSECA:
@daren-thomas thanks for the feedback, I updated the documentation accordingly.

Issue #530, opened 2017-03-07
Make that the floor height be calculated from Height and floor internally


JIMENOFONSECA:
Also raise error for cases that exceed the standard: E.g., 100 floors in 10 meters.

gabriel-happle:
@JIMENOFONSECA I changed the milestone, as this is not related to EcoCampus. Also I noticed that nobody is assigned to this issue!

PR #1671, opened 2018-11-08
I530 floor height

JIMENOFONSECA:
This solves issue #530 

@gabriel-happle would you mind testing this please?

Just use this zone file and try to run radiation_main. If you get an error message, then it works!

[zone.zip](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2560590/zone.zip)


JIMENOFONSECA:
@gabriel-happle weird. the changes added to this PR disappeared. @daren-thomas any idea why this happened?

daren-thomas:
@JIMENOFONSECA what do you mean by "the changes added to this PR disappeared"? Did _you_ commit changes and can't find them? If so... try pushing the changes - you might have only committed locally.

daren-thomas:
@JIMENOFONSECA another thing that might have happened: Updating from master might have undid the changes.

Issue #1511, opened 2018-06-18
CEA does not support WWR=1.0


gabriel-happle:
When running the CEA with buildings with WWR=1 the demand calculation breaks.
Radiation works.

```
Running C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/cea/demand/demand_main.py
Backend Qt5Agg is interactive backend. Turning interactive mode on.
Running demand calculation for scenario C:\cea-case-studies-local-05-06-2018\reference-case-WTP-reduced\WTP_MIX_m
Running demand calculation with weather file C:\Users\Gabriel\Documents\GitHub\CEAforArcGIS\cea\databases\weather\Singapore-2016.epw
Running demand calculation for region SIN
Running demand calculation with dynamic infiltration=False
Running demand calculation with multiprocessing=False
Running demand calculation with daysim radiation=True
Running demand calculation with stochastic occupancy=False
read input files
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm 2018.1.2\helpers\pydev\pydev_run_in_console.py", line 52, in run_file
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/cea/demand/demand_main.py", line 211, in <module>
    main(cea.config.Configuration())
  File "C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/cea/demand/demand_main.py", line 195, in main
    demand_calculation(locator=locator, gv=cea.globalvar.GlobalVariables(), config=config)
  File "C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/cea/demand/demand_main.py", line 91, in demand_calculation
    override_variables)
  File "C:/Users/Gabriel/Documents/GitHub/CEAforArcGIS/cea/demand/demand_main.py", line 133, in properties_and_schedule
    building_properties = BuildingProperties(locator, gv, use_daysim_radiation, region, override_variables)
  File "C:\Users\Gabriel\Documents\GitHub\CEAforArcGIS\cea\demand\building_properties.py", line 94, in __init__
    prop_geometry, prop_HVAC_result, use_daysim_radiation)
  File "C:\Users\Gabriel\Documents\GitHub\CEAforArcGIS\cea\demand\building_properties.py", line 268, in calc_prop_rc_model
[...]

PR #1673, opened 2018-11-08
adding flag to solve issue

JIMENOFONSECA:
this solves issue #1511 

@gabriel-happle would you like to verify if this patch solves the problem?


daren-thomas:
@JIMENOFONSECA ready to merge?

PR #1674, opened 2018-11-08
Fixing substation.calc_HEX_cooling loop with bhargava

daren-thomas:
@BhargavaKrishnaSreepathi and @daren-thomas did some refactoring on the `substation.calc_HEX_cooling` loop and introduced a bug.

This pull request fixes that bug. It includes a file `bin/test_substation_calc_hex_cooling.py` that, when run, runs both versions with different combinations of input values, printing out the difference in the calculation.

@BhargavaKrishnaSreepathi - could you check out that file and add interesting test cases? I don't really know enough about the edge cases and what are good values to test for here, so I just made some guesses. As far as I can tell, the two versions calculate the same.

BTW: as noted in the test code, there are some constraints on the input values to these functions that should be tested at the top of the function. This function could do with a unit test!


BhargavaKrishnaSreepathi:
The results obtained are exactly same before and after implementing the changes, and all the tests are passing.

PR #1690, opened 2018-11-16
i1679: Varying electrical price integration

BhargavaKrishnaSreepathi:
This PR includes electricity costs as an array for every hour of the year. This PR introduces a new `boolean parameter` to the config file named `detailed-electricity-pricing`. It also includes an excel file with the electricity costs of Singapore of 2017. It is currently added in both CH and SG databases present in `cea\databases\SIN(CH)\systems\electricity_costs.xlsx`

For cases where detailed electricity costs are not present, the code will take the values present in the LCA database for the cost of resources (already existent) and create arrays with the same cost for all 8760 hours of the year.

The way to test this PR:
1. Use any case study `cooling` or `heating`
2. Run the entire CEA test procedure with `detailed-electricity-pricing` as `false`
3. Repeat the same with the flag as `true`

The differences in the results will only be in `costs`.


cea/optimization/lca_calculations.py
```diff
@@ -112,19 +113,41 @@ def __init__(self, locator, config):
...
+                self.ELEC_PRICE = np.ones(8760) * average_electricity_price
+                self.EL_TO_OIL_EQ = resources_lca[resources_lca['Description'] == 'Electricity'].iloc[0][
+                    'PEN']  # MJ_oil / MJ_final
+                self.EL_TO_CO2 = resources_lca[resources_lca['Description'] == 'Electricity'].iloc[0][
+                    'CO2']  # kg_CO2 / MJ_final - CH Verbrauchermix nach EcoBau
+
+        average_gree_electricity_cost = resources_lca[resources_lca['Description'] == 'Solar'].iloc[0]['costs_kWh']  # MJ_oil / MJ_final
+        self.ELEC_PRICE_GREEN = np.ones(8760) * average_gree_electricity_cost
+        self.EL_TO_OIL_EQ_GREEN = resources_lca[resources_lca['Description'] == 'Solar'].iloc[0]['PEN']  # MJ_oil / MJ_final
+        self.EL_TO_CO2_GREEN = resources_lca[resources_lca['Description'] == 'Solar'].iloc[0]['CO2'] # kg_CO2 / MJ_final
```
JIMENOFONSECA: This is hard-coding again the region values. Any chance we can avoid this?

JIMENOFONSECA: @BhargavaKrishnaSreepathi

cea/optimization/slave/cooling_resource_activation.py
```diff
@@ -20,30 +20,30 @@
 __status__ = "Production"
 
 
-def calc_vcc_operation(Qc_from_VCC_W, T_DCN_re_K, T_DCN_sup_K, prices, lca, limits):
+def calc_vcc_operation(Qc_from_VCC_W, T_DCN_re_K, T_DCN_sup_K, prices, lca, limits, hour):
```
daren-thomas: here as well: we're not interested in the hour at all. or the lca object. all we really care about is  the `electricity_price` (for the hour - but what the `calc_vcc_operation` calculation does not depend on the hour at all!

cea/optimization/slave/cooling_resource_activation.py
```diff
@@ -20,30 +20,30 @@
...
+    opex_var_VCC_USD = VCC_operation['wdot_W'] * lca.ELEC_PRICE[hour]
     co2_VCC_kgCO2perhr = VCC_operation['wdot_W'] * lca.EL_TO_CO2 * 3600E-6
     prim_energy_VCC_MJperhr = VCC_operation['wdot_W'] * lca.EL_TO_OIL_EQ * 3600E-6
     Qc_CT_VCC_W = VCC_operation['q_cw_W']
-    E_used_VCC_W = opex_var_VCC_USD / lca.ELEC_PRICE
+    E_used_VCC_W = opex_var_VCC_USD / lca.ELEC_PRICE[hour]
     return opex_var_VCC_USD, co2_VCC_kgCO2perhr, prim_energy_VCC_MJperhr, Qc_CT_VCC_W, E_used_VCC_W
 
-def calc_vcc_backup_operation(Qc_from_VCC_backup_W, T_DCN_re_K, T_DCN_sup_K, prices, lca, limits):
+def calc_vcc_backup_operation(Qc_from_VCC_backup_W, T_DCN_re_K, T_DCN_sup_K, prices, lca, limits, hour):
```
daren-thomas: here, too, seems you're not interested in the hour, but in the electricity price...

cea/optimization/master/evaluation.py
```diff
@@ -171,26 +171,26 @@ def evaluation_main(individual, building_names, locator, solar_features, network
...
+        GHG_heating_tonCO2 = 0.0
+        costs_heating_USD = 0.0
+        PEN_heating_MJoil = 0.0
 
     costs_USD += costs_heating_USD
     GHG_tonCO2 += GHG_heating_tonCO2
     PEN_MJoil += PEN_heating_MJoil
 
     # District Cooling Calculations
     if gv.ZernezFlag == 1:
```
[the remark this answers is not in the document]
JIMENOFONSECA: @daren-thomas I totally agree let's remove it.. I think this was a flag to consider wood furnace also. which was the case of zernez. @BhargavaKrishnaSreepathi please take this int consideration

BhargavaKrishnaSreepathi:
@JIMENOFONSECA @daren-thomas: sure, I will do the necessary modifications, but it will take some time as I am trying to get results on another study. Will let you know when it is ready

cea/optimization/slave/cooling_resource_activation.py
BhargavaKrishnaSreepathi: It is true that the hour is not actually necessary, but in `lca`, we need `EL_TO_CO2`, `EL_TO_OIL_EQ` and `ELEC_PRICE`. Only electricity price is hour dependent

One way to do this is pass all these variables in the function call. Other way to do this is pass the lca fully and then also pass the hour. 

Is one way better than the other? I am not so sure.

BhargavaKrishnaSreepathi:
i think the PR is ready now

JIMENOFONSECA:
@daren-thomas any chance to see why this is failing? thanks! @BhargavaKrishnaSreepathi

Issue #1700, opened 2018-11-27
As a user, I would like to model buildings with unoccupied spaces (e.g. basements)


martin-mosteiro:
Since the executive course, the definition for the occupied area in CEA changed from Af (the conditioned area) to the gross floor area (GFA). This was necessary for Singapore since there are occupied unconditioned spaces. However, buildings in the European context also have unoccupied, unconditioned spaces – basements, for example. So right now CEA assumes that there are people in these spaces, which can grossly overestimate the number of people in the building, the water demand, etc.

In order to solve this, a new parameter could be defined, the net floor area (NFA), such that we'd now have:
- Gross floor area (GFA) = entire floor area of the building
- Net floor area (NFA) = occupied floor area of the building (usually NFA < GFA)
- Conditioned floor area (Af) = floor area that is heated or cooled (in Singapore, usually Af<NFA).

JIMENOFONSECA:
Well, we moved to use GFA because it is more standard with the concept of end-use energy intensity. Which helps to cover more places worldwide.

However, I understand NFA is used for the 2000-Watt society and Minergie, which makes sense to a great extend.

Ideally we have both options. To me GFA just makes things easier to compare scenarios and several locations. and well, I believe we have more certainty in the GFA of a building due to its geometry than the NFA (based on thermal zones).

Kind regards,
Jimeno

_______________________

Dr. Jimeno A. Fonseca
Senior researcher & Project coordinator
Multi-scale energy systems for low-carbon cities
Singapore ETH-Centre

Product owner & Scrum master
City Energy Anlayst (CEA)<http://www.cityenergyanalyst.com>

Prof.h.c in modeling and simulation
Universidad Central de Colombia


On 27 Nov 2018, 5:29 PM +0800, Martín Mosteiro-Romero <notifications@github.com>, wrote:

One thing that I am still not sure about: should the plots still use GFA as the reference area? Or NFA? @JIMENOFONSECA<https://github.com/JIMENOFONSECA>, what do you think?

martin-mosteiro:
I don't think we actually do, because our calculation of GFA is very rough: ground floor geometry * number of floors. So if you take any building that is not shaped like a box, the whole calculation is off (e.g., what happens with a building where the size of the floor plan decreases with increasing height? then we grossly overestimate the GFA).
I think having a parameter like this NFA allows us to adjust for issues like that – which is basically what we had originally done with the conditioned floor area in the Hochschulquartier case study.

[...]

PR #1702, opened 2018-11-27
I1700 net floor area

martin-mosteiro:
This PR does the following:
1) Redefines the basis for calculating the number of people in the building and their associated demands to the Net Floor Area (NFA) instead of the Ground Floor Area (GFA) since parts of buildings in the European context are not usually occupied. (As described in issue #1700)
2) Correspondingly introduces 'Ns' as an input in `architecture.dbf` to define the percentage of the GFA that is actually occupied (i.e., NFA = Ns * GFA)
3) Reintroduces the input 'Es', the share of the floor area that is electrified (which was previously included as an input but later moved as a constant value equal to 0.9), as an input in `architecture.dbf`.
4) Adds the parameters 'Es' and 'Ns' to the CEA archetype database. For Singapore, 'Ns' was set to 1 and 'Es' was set to 0.9 to maintain the CEA status quo for the Singapore case. For Switzerland, Ns was set equal to Hs, since the occupied floor area is generally equal to the conditioned floor area. Since the standard values for lighting and appliance demand used in CEA are actually normalized by NFA, Es was also set equal to Hs for consistency.
5) Updates the data helper to account for these new inputs.
6) Updates the CEA example reference cases (reference-case-open and reference-case-WTP) to include these new inputs in the `architecture.dbf` file.
These changes are needed in order to account for the fact that the entire gross floor area of a European building is usually not occupied (e.g., basements) while also allowing to adjust the floor area produced by CEA for irregular buildings (see #1700 for some discussion on this). As a result, the simulations for the Hochschulquartier case study have a much more realistic number of occupants and electricity demand.
After updating the unittest values (necessary because the number of occupants is now slightly different than before) and adjusting test_schedules to assertAlmostEqual instead of assertEqual, this branch passes all tests on my computer.


martin-mosteiro:
@JIMENOFONSECA This didn't break the demand calculations on my computer.
@gabriel-happle It actually shouldn't since the change only really affects the Swiss case, because all these factors were set to 1 for Singapore, so that NFA=GFA.

Issue #1652, opened 2018-10-05
As a Practitioner, I'd like to produce Excel files from DBF files with an index of my choice


daren-thomas:
Currently, the `dbf-to-excel` tool adds an index column (as found in PR #1629 by @gabriel-happle) when creating an Excel file. It would be nice to be able to create the file with an arbitrary column as index (which would be used for creating the `DataFrame` from the dbf file), possibly defaulting to `Name` if present in the file - since this is a CEA tool, I think it's ok to assume our customary index column name.

Issue #1651, opened 2018-10-05
As a Practitioner, I'd like to be able to create *.xslx files from *.dbf files


daren-thomas:
The tool `dbf-to-excel` currently can only create *.xls files. Based on the name chosen for the parameter  `dbf:excel-file`, it should either create an xlsx or an xls file.

This was found by @gabriel-happle in PR #1629.

Issue #1604, opened 2018-08-16
dbf-to-excel: clarify arguments to script


daren-thomas:
I just ran the dbf-to-excel script:

```
cea dbf-to-excel --dbf-file "C:\reference-case-open\baseline\inputs\building-properties\age.dbf" --excel-file "C:\reference-case-open\baseline\inputs\building-properties\testing.xlsx"
City Energy Analyst version 2.7.22
Running `cea dbf-to-excel` with the following parameters:
- dbf-tools:dbf-file = C:\reference-case-open\baseline\inputs\building-properties\age.dbf
- dbf-tools:excel-file = C:\reference-case-open\baseline\inputs\building-properties\testing.xlsx
Running excel-to-dbf with dbf-file = C:\reference-case-open\baseline\inputs\building-properties\age.dbf
Running excel-to-dbf with excel-file = C:\reference-case-open\baseline\inputs\building-properties\age.xls
```
Notice how the output file (`testing.xlsx`) is not used, but instead the output file is derived from the name of the input file.

Is this the behavior that we want? 

- if so, we need to change the interface to just accept `dbf-file`.
- else, we need to change the script to use `excel-file` as input.

Also, check `excel-to-dbf` for similar behavior.

daren-thomas:
@JIMENOFONSECA can you please provide some input in how the tool _should_ work? Fixing it should be a piece of cake.

JIMENOFONSECA:
So what I understand is that it should convert from excel to dbf and viceversa. If an excel file is the input, then it converts to dbf. if a dbf file is input then it is converted to excel.

daren-thomas:
Yes, that's exactly what it does. Just the question is _how_. How is the output name chosen? Is it a parameter to the script? Or is it just the name of the input file with the new extension? Currently, it pretends to do the first, but ends up doing the second. So we should just decide which way we want it.

JIMENOFONSECA:
I think the name can be selected by the user once he/she selects the output directory.

The parameters to ask the user are: 1. the location of the file, and 2. the output to locate the new file (including the name of the new file). The default name is the name of the input file.

Does it work like this?
On 17 Aug 2018, 21:52 +0800, Daren Thomas <notifications@github.com>, wrote:

[...]

Issue #1691, opened 2018-11-16
As a user, I would like to use the DBF to Excel tool on ArcScene to create new Excel files


martin-mosteiro:
The "Excel to DBF" tool on ArcScene does not accept a non-existing file as a target path for the Excel file to be created. So unless there already is an Excel file in the target folder, you cannot use the tool.

![screen shot 2018-11-16 at 14 30 18](https://user-images.githubusercontent.com/18548065/48624252-3ead6d80-e9ac-11e8-9a4d-4b460488296c.png)
(note: '*.xlsx' does not work, either)

The tool works correctly on Pycharm, so the issue must be the ArcScene tool itself.

PR #1713, opened 2018-11-30
1651 1652 dbf files fix

JIMENOFONSECA:
This solve issues #1651 #1652 and #1604 

It also solves provlems in the documentation of CEA

@martin-mosteiro could you take a look at this?

To test:
1. update cea install-toolbox
2. run utilities/dbf-to-excel-to-dbf
3. use any settings you want. you can either specify an excel file or a DBF file. The tool will convert to the other format automatically.


martin-mosteiro:
@JIMENOFONSECA I did `cea-install-toolbox` and got the following error message:
```pkg_resources.DistributionNotFound: The 'python-dateutil>=2.2' distribution was not found and is required by pycollada
```
So then I did `pip install python-dateutil --upgrade` first followed by `cea install-toolbox` and it worked.
However, when I tried to use the DBF tool in ArcScene....
screen shot 2018-11-30 at 11 54 20
screen shot 2018-11-30 at 11 53 50
That is, I still get the issue described in #1691

JIMENOFONSECA:
Wierd. The toolbox in this branch should have 4 inputs. The one you are showing has two.. did you try to do the process from the branch?

Best,
Jimeno
On 30 Nov 2018, 18:55 +0800, Martín Mosteiro-Romero <notifications@github.com>, wrote:

@JIMENOFONSECA<https://github.com/JIMENOFONSECA> I did cea-install-toolbox and got the following error message:

So then I did pip install python-dateutil --upgrade first followed by cea install-toolbox and it worked.
However, when I tried to use the DBF tool in ArcScene....
[screen shot 2018-11-30 at 11 54 20]<https://user-images.githubusercontent.com/18548065/49285049-bb554700-f496-11e8-8283-3b4228a4e08e.png>
[screen shot 2018-11-30 at 11 53 50]<https://user-images.githubusercontent.com/18548065/49285015-a11b6900-f496-11e8-9b08-286f5148ea89.png>
That is, I still get the issue described in #1691<https://github.com/architecture-building-systems/CityEnergyAnalyst/issues/1691>

JIMENOFONSECA:
Yes. These are messagges from the arcgis interface. It does this weird behavior always. Maybe an issue to address?

martin-mosteiro:
But it's not just a message, the script is actually copying the databases to the folder specified by the user. This seems neither necessary nor desirable.

daren-thomas:
@martin-mosteiro, @JIMENOFONSECA this is by design and was explicitly requested by @JIMENOFONSECA: Each time a script is run, the databases are copied.

The relevant lines are (from b12020277a038a5c347e9c44178a7ef82b3a0a02) in `cea.interfaces.cli.cli`, 55ff:

```
 # FIXME: remove this after Executive Course
    # <--
    config.restrict_to(['general:scenario', 'general:region'] + cea_script.parameters)
    cea.datamanagement.copy_default_databases.copy_default_databases(
        locator=cea.inputlocator.InputLocator(config.scenario), region=config.region)
    config.restrict_to(cea_script.parameters)
    # -->
```

Jimeno wanted to make sure the default databases are copied every single time a script was run. As far as I understand, this only happens if the file does not yet exist.

martin-mosteiro:
@daren-thomas You're right – seems like only "benchmarks", "lifecycle", "systems" and "uncertainty" were copied into my case study, but not "archetypes" since it was already there.
So @JIMENOFONSECA  this is working as designed (*). The PR passes my test!

(*) However, I think copying the databases when running scripts related to the specific case study makes sense to make sure the case study is complete. However, since this tool is a tangential tool (file conversion, completely unrelated to any other CEA scripts) I don't think copying the databases is necessary and could actually be a nuisance. But since the script doesn't break anything, I'm okay with passing it as is.

daren-thomas:
Actually, this copying of database files is totally unnecessary, since they will be copied on first access anyway (see `cea.inputlocator.InputLocator#_get_region_specific_db_file`).

Issue #1717, opened 2018-11-30
As a Practitioner, I would like to know how to model existing urban heating networks


daren-thomas:
When setting up a case study, how do I model existing urban heating networks? What files do I need to provide? How do I make sure the network is being used?

(Asking for Anja who has such a case)

JIMENOFONSECA:
@shanshanhsieh could you please adrees this issue? Thanks!

shanshanhsieh:
yes, i will write a tutorial

shanshanhsieh:
merged with PR#1717

PR #1721, opened 2018-12-06
1717 model existing thermal networks

shanshanhsieh:
this PR resolves issue #1717 
a guide on **How to prepare inputs for network simulation** is added


shanshanhsieh:
i was struggling a bit on changing the size of the image, maybe @daren-thomas can give me a hand on that?

PR #1726, opened 2018-12-13
1725 autdoc errors for demand metamodel

Jack-Hawthorne:
The main problems were with the keras callbacks missing in the MOCK_MODULES and erroneous import path for nn_settings in one of the scripts.  

Should be fine now.


daren-thomas:

PR #1733, opened 2018-12-16
1730 Autodoc errors for thermal_network/network_layout

Jack-Hawthorne:
adding the mock modules has fixed the build errors. easy one


daren-thomas:

PR #1737, opened 2018-12-18
I1735 fix full demand reporting

gabriel-happle:
this fixes the full reporting to `.xls` function of the demand calculation.

It can be activated by uncommenting line 182 in `thermal_loads.py`.

Following changes have been made additionally:

- deactivate the quick visualization (plotly) of the demand.
- some reporting `system_status` variables of cooling systems.

Assigning @shanshanhsieh and @BhargavaKrishnaSreepathi to test and merge, since they are currently using this functionality.

### DIFF
--- a/cea/demand/thermal_loads.py
+++ b/cea/demand/thermal_loads.py
@@ -13,7 +13,7 @@ from cea.demand import sensible_loads, electrical_loads, hotwater_loads, refrige
 from cea.demand import ventilation_air_flows_detailed, control_heating_cooling_systems
 from cea.technologies import heatpumps
 
-import cea.utilities
+from cea.utilities import reporting
 
 def calc_thermal_loads(building_name, bpr, weather_data, usage_schedules, date, locator, use_stochastic_occupancy,
                        use_dynamic_infiltration_calculation, resolution_outputs, loads_output, massflows_output,
@@ -179,7 +179,7 @@ def write_results(bpr, building_name, date, format_output, loads_output, locator
 
     # write report & quick visualization
     # (NOTE: uncomment to write full report for debugging purposes)
-    # cea.utilities.reporting.full_report_to_xls(tsd, locator.get_demand_results_folder(), building_name)
+    reporting.full_report_to_xls(tsd, locator.get_demand_results_folder(), building_name)
 
 
 def calc_Qcs_sys(bpr, tsd):
--- a/cea/demand/thermal_loads.py
+++ b/cea/demand/thermal_loads.py
@@ -179,7 +179,7 @@ def write_results(bpr, building_name, date, format_output, loads_output, locator
 
     # write report & quick visualization
     # (NOTE: uncomment to write full report for debugging purposes)
-    reporting.full_report_to_xls(tsd, locator.get_demand_results_folder(), building_name)
+    # reporting.full_report_to_xls(tsd, locator.get_demand_results_folder(), building_name)
 
 
 def calc_Qcs_sys(bpr, tsd):
--- a/cea/utilities/reporting.py
+++ b/cea/utilities/reporting.py
@@ -7,9 +7,7 @@ import os
 from plotly.offline import plot
 import plotly.graph_objs as go
 
-from cea.demand.thermal_loads import TSD_KEYS_HEATING_LOADS, TSD_KEYS_HEATING_TEMP, TSD_KEYS_RC_TEMP, \
-    TSD_KEYS_COOLING_LOADS, TSD_KEYS_MOISTURE, TSD_KEYS_VENTILATION_FLOWS, TSD_KEYS_COOLING_SUPPLY_TEMP, \
-    TSD_KEYS_COOLING_SUPPLY_FLOWS
+
 
 __author__ = "Gabriel Happle"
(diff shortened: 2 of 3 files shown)


gabriel-happle:
@daren-thomas any idea why the tests are failing? On my machine I can run the `dodo`. It produces a bunch of warnings related to PySAL and pandas, but otherwise it's running fine.

daren-thomas:
@gabriel-happle I'm trying to find out why it stopped working - it's a conda issue, or more specifically, it has to do with the conda packages for fiona. I think.

> It can be activated by uncommenting line 182 in thermal_loads.py.

This is worrying... could you not just use the `debug` parameter? Or create it's own parameter? I would prefer it if we didn't keep functionality hidden in the comments of the code.

cea/demand/thermal_loads.py
```diff
@@ -179,7 +179,7 @@ def write_results(bpr, building_name, date, format_output, loads_output, locator
 
     # write report & quick visualization
     # (NOTE: uncomment to write full report for debugging purposes)
-    # cea.utilities.reporting.full_report_to_xls(tsd, locator.get_demand_results_folder(), building_name)
+    # reporting.full_report_to_xls(tsd, locator.get_demand_results_folder(), building_name)
```
daren-thomas: this should be handled via a flag in the config file or something. there is one called "debug" - we could either use that one or add a new one specific to reporting here like the `use_dynamic_infiltration_calculation`)

cea/utilities/reporting.py
```diff
@@ -40,18 +38,25 @@ def full_report_to_xls(tsd, output_folder, basename):
     writer.close()
 
     # quick visualization
-    quick_visualization_tsd(tsd, output_folder, basename)
+    # UNCOMMENT THE NEXT LINE TO PRODUCE QUICK VISUALIZATIONS WITH PLOTLY OF THE DEMAND CALCULATION IN BUILDINGS
+    # quick_visualization_tsd(tsd, output_folder, basename)
```
daren-thomas: please don't hide functionality in comments. make this an option to the script.

cea/utilities/reporting.py
```diff
@@ -40,18 +38,25 @@ def full_report_to_xls(tsd, output_folder, basename):
...
 def quick_visualization_tsd(tsd, output_folder, building_name):
 
-    plot_heat_load = False
-    plot_heat_temp = False
+    # import keys
+    from cea.demand.thermal_loads import TSD_KEYS_HEATING_LOADS, TSD_KEYS_HEATING_TEMP, TSD_KEYS_RC_TEMP, \
+        TSD_KEYS_COOLING_LOADS, TSD_KEYS_MOISTURE, TSD_KEYS_VENTILATION_FLOWS, TSD_KEYS_COOLING_SUPPLY_TEMP, \
+        TSD_KEYS_COOLING_SUPPLY_FLOWS
+
+    # set to True to produce plotly graphs of selected variables
```
daren-thomas: again, this code assumes the user will change the _source_ to turn functionality on or off. let's not do that. this is what the config file is for. if you don't want to create a whole bunch of options, just have one option (quick-visualization-tsd) or something. then, it is documented and further users / researchers will learn about the functionality from the user interface.

JIMENOFONSECA:
@gabriel-happle there are some changes waiting for you to make ;-)

cea/demand/thermal_loads.py
gabriel-happle: I believe this is your code @daren-thomas. I was just fixing something that researchers here need.
I will create a new flag in the config file. Let me figure out how.

cea/utilities/reporting.py
gabriel-happle: Again, just copying what I saw above.
I will delete the code. And keep it somewhere for myself.
I think it's useful for testing and understanding what's happening.
It's not really for researchers, it's more for developers.

cea/utilities/reporting.py
gabriel-happle: See above.

gabriel-happle:
I'll work on it Thursday or Friday.

cea/demand/thermal_loads.py
daren-thomas: if it's really just for developers, you can use the `debug` flag in the config file. that's what it's for :-)

cea/utilities/reporting.py
daren-thomas: the debug flag could help?

daren-thomas:
@gabriel-happle this is the error message from the failed build:

```
ERROR: test_calc_thermal_loads (test_calc_thermal_loads.TestCalcThermalLoads)
  File "c:\users\darthoma\miniconda2\envs\ceatest\lib\site-packages\cea\tests\test_calc_thermal_loads.py", line 60, in test_calc_thermal_loads
    self.format_output, self.region)
TypeError: calc_thermal_loads() takes exactly 16 arguments (14 given)
```

it seems like some interface was changed.

daren-thomas:
@gabriel-happle I agree - generally when unittesting you want to control the variables as much as possible. This test is a bit of a hybrid and I think it's unlikely that the results of the test would change based on those two variables - I'm just going to leave it like this. Thank you :)

PR #1746, opened 2019-01-14
adding credits to the best of my knowledge

JIMENOFONSECA:
@shanshanhsieh @BhargavaKrishnaSreepathi @gabriel-happle Since you worked in the plots. Please cehck if you contributed to one or more of those scripts. I tried my best to reflect all authors but I might not remember all. thanks for your help!

### DIFF
--- a/cea/plots/demand/energy_balance.py
+++ b/cea/plots/demand/energy_balance.py
@@ -9,6 +9,14 @@ from cea.plots.variable_naming import LOGO, COLOR
 import pandas as pd
 import numpy as np
 
+__author__ = "Gabriel Happle"
+__copyright__ = "Copyright 2018, Architecture and Building Systems - ETH Zurich"
+__credits__ = ["Gabriel Happle","Jimeno A. Fonseca"]
+__license__ = "MIT"
+__version__ = "2.8"
+__maintainer__ = "Daren Thomas"
+__email__ = "cea@arch.ethz.ch"
+__status__ = "Production"
 
 def energy_balance(data_frame, analysis_fields, normalize_value, title, output_path):
     # Calculate Energy Balance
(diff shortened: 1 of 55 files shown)


cea/plots/demand/energy_balance.py
```diff
@@ -9,6 +9,14 @@
 import pandas as pd
 import numpy as np
 
+__author__ = "Gabriel Happle"
+__copyright__ = "Copyright 2018, Architecture and Building Systems - ETH Zurich"
+__credits__ = ["Gabriel Happle","Jimeno A. Fonseca"]
```
daren-thomas: needs a comment after the space

cea/plots/demand/energy_balance.py
```diff
@@ -9,6 +9,14 @@
...
+__license__ = "MIT"
+__version__ = "2.8"
```
daren-thomas: i wonder how important the `__version__` attribute is and what it should mean.

cea/plots/demand/energy_balance.py
daren-thomas: `["Gabriel Happle","Jimeno A. Fonseca"]` - there should be a space after the comma. pep8. (PyCharm hints at this with squiggly lines, similar to Word. ideally, the less such complaints that are unnecessary, the more information they provide, so it's good to fix the easy stuff by habit)

cea/plots/demand/energy_balance.py
daren-thomas: it's inconsistent among the files. sure. i'm totally fine with leaving it for now! but at some point in the future we'll have to come up with a good reason :-)

JIMENOFONSECA:
@daren-thomas OK I solved the comments, would you mind merging this? thanks!

PR #1756, opened 2019-01-22
Fix issue with `mcpcdata_sys_kWperC`

martin-mosteiro:
Previously it was incorrectly labeled as `mcpdata_sys_kWperC`


daren-thomas:
@martin-mosteiro how do i test this? i have updated from master, so the jenkins test will run through.

martin-mosteiro:
The existing tests should pass, since the changes I made do not affect the tests (and hence the tests were passing in spite of the bug being present in it).

I suppose a full test should involve running a thermal network simulation with data centers and making sure the code works – which Mahshid has already done.

Perhaps reference-case-open could be updated to contain more uses, since at the moment it has only two...

daren-thomas:
@martin-mosteiro could you please test again? and create an issue for updating the jenkins tests to catch this in the future? thank you :)

PR #1772, opened 2019-01-25
updated the install instructions to include size and time

Jack-Hawthorne:
I changed the disclaimer to occur at the start rather than the end (makes more sense).

Omitted Anaconda2 (just Miniconda2) as an install option to reduce time and uncertainty.

Created space and time estimations for the installation section to advise new comers.


daren-thomas:

PR #1778, opened 2019-01-31
I 1767 fix longwave radiation bug

gabriel-happle:
This PR fixes a bug in the long wave radiation loss calculation.
The bug was discovered by @pajotca. Thank you!

As a consequence of this fix, the radiation losses are around **-15%** smaller than the current master for the CH case.
This can result in lower heating energy demand in the range of -1%...-3% (CH case).
Cooling energy demand can be -3%...+0% lower/higher (CH case).

Assigning @martin-mosteiro and @JIMENOFONSECA as testers as they are most familiar with ISO 13790.


JIMENOFONSECA:
@gabriel-happle if somebody did the PR and you tested it, you can just go ahead an merge it!

martin-mosteiro:
The code looks good, the equations are consistent with the ISO standard.

On reference-case-open, the heating demand decreased by 2.7%, while the cooling demand increased by 0.04%. On reference-case-zurich, the heating demand was decreased by 2.0% while the cooling demand interestingly also decreased (by 0.1%).

All in all, the results seem feasible and the calculations are correct, so I'm merging!

martin-mosteiro:
Oops, just noticed @JIMENOFONSECA was also assigned. Hope he agrees ;)

PR #1785, opened 2019-02-01
Fixing calc cop ccgt

BhargavaKrishnaSreepathi:
Fixing a refactoring bug present in Master. This was affecting optimization script.

The way to test the script is to start a fresh test case and test optimization. More details can be found here 
https://city-energy-analyst.readthedocs.io/en/latest/cea-workflow-guide.html 
Look at step 9

Martin, I am assigning you such that you can test your bug in this branch and suggest if there needs to be any changes


martin-mosteiro:
The branch passes all tests and the simulation I was having issues with works fine now. The PR looks good to me!

PR #1787, opened 2019-02-05
1069 document input output variables

Jack-Hawthorne:
I'm working on documenting all input and output variables via the trace_inputlocator script.

**new things**:
* altered the trace-inputlocator graphviz and yaml template to separate ``file_path`` and ``file_name``
* added two different graphviz output types depending on if a single script (more detailed) or multiple scripts (more relational) are traced
* added dir in docs called ``script_dependencies`` containing trace_inputlocator results for various scripts (missing a lot still)
* new script in ``script_dependencies`` dir which matches the name of the graphviz file to possible scripts from cea.scripts, then reads the digraphs and inserts them into a jinja2 template. the output is then written to the docs dir as ``what-are-the-cea-scripts``

**problems**:
* the trace-inputlocator only saves the inputlocator method with no information about which module and method within the script is calling it

**to do**:
* find a way to trace the module name and method calling the inputlocator (and maybe/hopefully which specific variables are used)
* adapt the trace_inputlocator to output this new data
* run the trace-inputlocator for all scripts
* run the trace-inputlocator for all scripts in one instance (resulting in the master architecture)


daren-thomas:

daren-thomas:
image

First steps towards providing an overview of the inputs / outputs of the scripts.

JIMENOFONSECA:
@Jack-Hawthorne great effort, may I ask for some design aspects to be included? 
1. A legend text explaining how to read the figure.
2. The same width for every white box. 
3. use the "blue" color of cea for the orange boxes. The colors are included in the plots folder of CEA.

Jack-Hawthorne:
@JIMENOFONSECA no worries i'll aim to get finished by end of week!

Jack-Hawthorne:
graphviz
 @JIMENOFONSECA before i run this for most of the scripts, is this more what you're after.

i chose to write a over-arching legend for sake of repetition.

JIMENOFONSECA:
yes, that looks good. I would do a legend for each diagram. It does not matter if it is repetitive. It makes easier to get the idea straight away.

I wonder if we can split it in input and output folders. So it is easier to understand where the data comes from and where it goes to straight away.

An example of the legend could be:

Dark Blue: Main script name
Light Grey: Output folder path
Light Blue: Input folder path
White: file name
Floating text: intermediate function name



Let me know how it goes.


On 19 Feb 2019, 4:04 AM +0800, Jack-Hawthorne <notifications@github.com>, wrote:

[graphviz]<https://user-images.githubusercontent.com/36536537/52974082-97800e00-33c0-11e9-9b69-2a3385faa3af.png>
@JIMENOFONSECA<https://github.com/JIMENOFONSECA> before i run this for most of the scripts, is this more what you're after.

i chose to write a over-arching legend for sake of repetition.

daren-thomas:
@JIMENOFONSECA I'd like to diagree. I don't think the legend adds any clarity at all. The example legend will also become a problem when this kind of diagram is used to show multiple scripts, where one produces output that becomes input for another script. e.g. data-helper & demand.

Also, the arrows show very clearly the direction of data flow. The file names with the suffixes (.xlsx, .dbf) are clearly file names. The method names could be made more obvious by using a fixed-width font and adding parenthesis (`get_building_architecture`). 

I'd rather we just had this information for all scripts, than spending more time on making it look pretty. We can easily do that later on if we really need to. It doesn't add much value IMO.

Jack-Hawthorne:
ok so I've added a legend in which looks like this:
graphviz 2

the problem with only having the input, output or database label is:

- the loss of some detail in the information presented
- also it gets a little bit trickier to code for instances where the scripts looks for default databases in cea:
``..\..\users\documents\cea\databases``

apart from that is it what you were after, any more adjustments?

Jack-Hawthorne:
yay or nay on the legend - it's up to you @daren-thomas @JIMENOFONSECA
hard to make - very easy to remove

Jack-Hawthorne:
> @JIMENOFONSECA I'd like to diagree. I don't think the legend adds any clarity at all. The example legend will also become a problem when this kind of diagram is used to show multiple scripts, where one produces output that becomes input for another script. e.g. data-helper & demand.

i didn't touch the multiple scripts case, it's way too complicated -> multiple scripts is still without legend.

JIMENOFONSECA:
@Jack-Hawthorne @daren-thomas thanks for the new version. i believe the solution could be a bit  easier. Let me make myself more clear with the next diagrams. This should make things intuitive for the reader and not too repetitive.

[Untitled Page.pdf](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2880346/Untitled.Page.pdf)

Jack-Hawthorne:
@daren-thomas this legend is general and should not have any problems with multiple scripts, right?

dot is a pain in the ass when it comes to aligning things, the legend can end up interfering with the information you actually want to portray very easily and unexpectedly. 

For multiple scripts, i think its better to go for relations rather than details (you can always find the details with the single scripts). IMO it's better to keep the multiple scripts minimalist with centralised dirs (i.e. no splits between input and outputs if within the same directory). it will also be very tricky to code some scripts outputs as others' inputs using general rules (it will be very case specific). 

splitting the dirs for single scripts was tedious enough! anyway, here is the new format:

graphviz 3

this looks more like the second option in the pdf, the other is not as useful as daren has said.

Jack-Hawthorne:
@JIMENOFONSECA 
this is how demand is looking now
graphviz 4

JIMENOFONSECA:
@Jack-Hawthorne this is totally and absolutely clear! thanks for this!. The thing that the diagram big is just the pact that we use many files in that script, but visually it is quite easy to understand now.

daren-thomas:
the tests ran through locally, so I'm merging.

Jack-Hawthorne:
@daren-thomas i have a few more changes to add and to replace all the viz outputs - about half way through. i'll let you know when its ready to merge (hopefully end of day)

Jack-Hawthorne:
@daren-thomas ok the first version should be good to go.

i'm still missing a few scripts but these can be put in later when they all work on my build (some issues with plots and network-layout) 

merge away

PR #1788, opened 2019-02-06
1729 autodoc error radiation daysim

daren-thomas:
This PR fixes some problems with the documentation.


Jack-Hawthorne:
looks nice! thanks for looking into the metaclass conflict!

Issue #1754, opened 2019-01-22
As a user I would like to simulate DC networks that supply data centers


martin-mosteiro:
While running DC network simulation for a case study that includes data center, we get the following error:
```
Traceback (most recent call last):
  File "C:/Users/Assistenz/Documents/GitHub/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_matrix.py", line 3249, in <module>
    main(cea.config.Configuration())
  File "C:/Users/Assistenz/Documents/GitHub/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_matrix.py", line 3242, in main
    thermal_network_main(locator, network_type, network_name, file_type, set_diameter, config, substation_systems)
  File "C:/Users/Assistenz/Documents/GitHub/CityEnergyAnalyst/cea/technologies/thermal_network/thermal_network_matrix.py", line 437, in thermal_network_main
    thermal_network.buildings_demands, substation_systems, thermal_network)
  File "C:\Users\Assistenz\Documents\GitHub\CityEnergyAnalyst\cea\technologies\thermal_network\substation_matrix.py", line 64, in substation_HEX_design_main
    substation_HEX = substation_HEX_sizing(buildings_demands[name], substation_systems, thermal_network)
  File "C:\Users\Assistenz\Documents\GitHub\CityEnergyAnalyst\cea\technologies\thermal_network\substation_matrix.py", line 181, in substation_HEX_sizing
    '', T_DC_supply_C, thermal_network)
  File "C:\Users\Assistenz\Documents\GitHub\CityEnergyAnalyst\cea\technologies\thermal_network\substation_matrix.py", line 229, in calc_hex_area_from_demand
    cs = (abs(building_demand[m].values)) * 1000  # in W/K
  File "C:\ProgramData\Miniconda2\envs\cea\lib\site-packages\pandas\core\frame.py", line 2688, in __getitem__
    return self._getitem_column(key)
  File "C:\ProgramData\Miniconda2\envs\cea\lib\site-packages\pandas\core\frame.py", line 2695, in _getitem_column
    return self._get_item_cache(key)
  File "C:\ProgramData\Miniconda2\envs\cea\lib\site-packages\pandas\core\generic.py", line 2489, in _get_item_cache
    values = self._data.get(item)
  File "C:\ProgramData\Miniconda2\envs\cea\lib\site-packages\pandas\core\internals.py", line 4115, in get
[...]

Issue #1176, opened 2018-03-29
CEA cannot run case studies where heating and cooling seasons overlap


martin-mosteiro:
The CEA `systems_controls` database includes the dates when the heating season starts and ends, and the dates when the cooling season starts and ends. However, it is *not* possible to run CEA if the heating and cooling seasons overlap (see `calc_heating_cooling_loads`).

Does this make sense? Couldn't there be buildings that might need cooling even during the nominal heating season (e.g. highly insulated buildings with lots of internal gains)?

And if it does make sense, shouldn't it be enough to define a heating season *or* a cooling season rather than both (instead of defining them separately and then needing to check that the entire year is covered)? @gabriel-happle @JIMENOFONSECA

martin-mosteiro:
@gabriel-happle Another interesting outcome of having cooling in the "heating season"... The heat pump code (`HP_air_air`) breaks down because the temperature difference in the evaporator and condenser approaches 0 and thus the COP approaches infinity...
Seems like the issue might be that such buildings with high gains need an appropriate ventilation system (i.e., the one assigned by the archetypes might be inadequate).

gabriel-happle:
Hi @martin-mosteiro, I did some quick research about heating and cooling seasons. I.e., times of the year when heating is allowed in Europe.
It seems this only applies for residential buildings, I.e., MULTI_RES and SINGLE_RES. Or do you know any information that says otherwise?

I guess one way to move forward is to apply heating and cooling seasons only to residential buildings.
For the other buildings we can modify that the algorithm decides what to do.
Of course this requires some testing, because we want to avoid a situation where heating and cooling happens intermittently.

What do you think?

Issue #1161, opened 2018-03-27
Technical systems in archetypes have outdated codes


martin-mosteiro:
The archetypes define technical systems by their old codes, thus all heated building types built after 2006 are assigned heating by central AC. This is doesn't make any sense! @JIMENOFONSECA I assume this is a mistake and these buildings were meant to have 'T4' = floor heating (*not* 'T3' = central AC)?

JIMENOFONSECA:
you are right!

On 27 Mar 2018, 12:14 PM +0800, Martín Mosteiro-Romero <notifications@github.com>, wrote:

The archetypes define technical systems by their old codes, thus all heated building types built after 2006 are assigned heating by central AC. This is doesn't make any sense! @JIMENOFONSECA<https://github.com/JIMENOFONSECA> I assume this is a mistake and these buildings were meant to have 'T4' = floor heating (not 'T3' = central AC)?

PR #1790, opened 2019-02-07
1769 reference case open improvement

martin-mosteiro:
Reference-case-open only had two occupancy types and no mixed-use buildings, so the unit tests missed a lot of details (see, e.g. #1754).

The new reference case comprises 9 mixed use buildings covering all of the different occupancy types available in CEA. The radiation results and "building metering" were correspondingly updated. The unittest data was also created again for this new reference case.


gabriel-happle:
- `dodo` fails because of `sensitivity`
- @martin-mosteiro can you check the if the building properties make sense? E.g., `Hs, Es, Ns = 0.45` in `B001` you know better about the archetypes of Switzerland.
- I will take a look at the detailed results now and get back to you tomorrow.

martin-mosteiro:
@gabriel-happle Thanks for checking that. B01 is 50% FOODSTORE, 50% COOLROOM. For FOODSTORE, `Hs=Es=Ns=0.9` (based on CEA archetypes).

The COOLROOM archetype has a conditioned surface of 0 according to CEA archetypes (`Hs=0`) [*] and no occupants (so `Ns=0`). However they **do** have a lighting demand, so the electrified surface should **not** be 0. I will create an issue about this.

I will also fix the reference case in this PR accordingly.

gabriel-happle:
I think we are discovering some issues because of this change.
Also the internal temperatures of `B001` don't seem to make much sense. The foodstore and coolroom is only cooled during the cooling season and has setbacks as well.

image

gabriel-happle:
`B003` (Industrial and Hospital) temperature is a bit hot in Winter.
image

martin-mosteiro:
@gabriel-happle regarding `B001`, I alluded to this in issue #1176 precisely because of this behavior in buildings with mixed occupancy included COOLROOM. I wonder if it makes sense to include the COOLROOMS in single-thermal zone simulation at all, because they have set point temperatures for cooling but they don't *actually* have a space cooling demand in CEA - their cooling is entirely assigned to refrigeration loads.

martin-mosteiro:
Regarding 'B003' I faced similar issues before as well - I believe the archetype definition for future buildings in some typologies might not be feasible, since they have gigantic internal loads but no heat recovery and are highly insulated.

gabriel-happle:
@martin-mosteiro let's create some issues. I think `B001` also reveals some bug regarding the schedule of temperature set points.  
I suggest we merge this PR since it is doing what it is supposed to do.

martin-mosteiro:
@gabriel-happle Good point. While the COOLROOM is not actually part of the thermal model (`Hs=0` according to the archetypes) it _is_ setting the setpoint for the thermal zone. That is not desired behavior.

gabriel-happle:
Created #1794 regarding the COOLROOM schedule.
Other issues are related to heating and cooling season controls.
This PR is ready to be merged. 
Can we merge @daren-thomas even though the checks are failing?

JIMENOFONSECA:
image

gabriel-happle:
@martin-mosteiro I don't really understand what the `test_schedules` function does. But I had to update the test results in order to pass the tests.
You are assigning a different usage [`OFFICE` & `INDUSTRIAL`] to one of the buildings, right? And then all the values of people and loads are smaller now, due to the smaller conditioned area of building `B001`...

martin-mosteiro:
Yes, that’s what it’s doing. The goal was just to have a way to test mixed use schedules since the reference case did not have any. Given the new reference case, this shouldn’t be necessary, so I can update the tests to accommodate for that.

martin-mosteiro:
Ah! I know what that issue is. I thought we had fixed it ages ago… seems we hadn’t. I’m on it.

gabriel-happle:
@JIMENOFONSECA we tried to fix the heating coil model. 
It produces some results that seem feasible and it passes the `dodo` until it fails in the sensitivity test.
You might want to take a look at the heating coil. It's really hard to figure out what's happening without documentation.

martin-mosteiro:
I believe this issue is caused by buildings with high internal temperatures being heated by the centralized AC unit. The current code allows heating to happen even when the indoor temperature is higher than the set point. The following line (line 145 in `airconditioning_model.py`) would fix(*) the issue:
```
    ~~if t_ve_mech_after_hex >= t_sup_h_ahu or \~~
    ~~t_ve_mech_after_hex >= ta_hs_set:~~
    if T_int >= ta_hs_set or t_ve_mech_after_hex >= t_sup_h_ahu or t_ve_mech_after_hex >= ta_hs_set:
```

@gabriel-happle since you made this change originally, what do you think of this proposal?

(*) It doesn't fix the fact that this building still gets overheated in late spring/early autumn, but it does fix the fact the overheating in winter.

martin-mosteiro:
Nevertheless, this also shows that issue #1161 is still occurring. Namely, centralized AC heating is not a common feature in Switzerland, much less for new buildings. Due to this we had decided back then that this was probably a mistake (radiant floor heating was the actual system that should have been in the archetypes, not central AC). However, this update seems to either never have been implemented or have been reverted subsequently.

JIMENOFONSECA:
image

this is the errror in jenkins @martin-mosteiro you might want to check tests locally again

martin-mosteiro:
Also, in the cooling coil calculation a constant is given as `U_COOL = 450.0`. Assuming this should rather be in `constants.py` I looked through the code and found that indeed there is a constant there which is `U_COOL = 2500.0`. @shanshanhsieh are these two constants supposed to be different? If so, can we name them something more intuitive (e.g. `U_COOL_WATER_HEX` and `U_COOL_AIR_HEX` or something). Also not sure where these values come from... and since they're not defined very clearly I am not sure what values to compare them to when I Google "Typical Overall Heat-Transfer Coefficients".

daren-thomas:
hey @martin-mosteiro what is the status here? 

btw: the jenkins is complaining about `AssertionError: qhs_sys_kwh for B03 should be: 25255.62400, was 203182.03600` in the unit test `test_calc_thermal_loads_other_buildings`.

martin-mosteiro:
@daren-thomas I made some corrections as discussed yesterday, wanted to run the unit tests but I have the same issue @gabriel-happle mentioned at the meeting yesterday. so I'm updating my conda environment and then testing.

martin-mosteiro:
@daren-thomas all tests passed locally

martin-mosteiro:
As discussed in the CEA meeting, I have also changed `B01` to 25% COOLROOM and 75% FOODSTORE, so that the temperature curve now looks more normal.

image

martin-mosteiro:
`B03` does not heat up as badly as before, but it still gets too hot in spring due to the huge internal loads he has and very insulated, airtight envelope:
image
I believe this is an issue of archetype definition (are industrial buildings ever that airtight and well-insulated?) rather than thermal calculation.

JIMENOFONSECA:
I believe that this behavior is physically possible. If it is, then the problem is not the model, the problem is the building system designer who will have a building that overheats. This a great educational example…
On 22 Feb 2019, 9:45 PM +0800, Martín Mosteiro-Romero <notifications@github.com>, wrote:

B03 does not heat up as badly as before, but it still gets too hot in spring due to the huge internal loads he has and very insulated, airtight envelope:
[image]<https://user-images.githubusercontent.com/18548065/53246213-59c90100-36b0-11e9-82f5-987eabad378a.png>
I believe this is an issue of archetype definition (are industrial buildings ever that airtight and well-insulated?) rather than thermal calculation.

gabriel-happle:
After updating the mixed schedules tests, all checks are finally passing.

PR #1793, opened 2019-02-13
Correct archetypal values of Es for PARKING and COOLROOM

martin-mosteiro:
Since these uses are not conditioned (`Hs=0`) nor occupied (occupant density 0), their electrified area was also set to 0. However, they both *do* have lighting demands, so they should be electrified. For consistency with the old CEA hard-coded value for electrified area, these were now set to `Es=0.9` for both uses. This was already the case for the Singapore case study, so no changes were made there.


JIMENOFONSECA:
@gabriel-happle let's take down this PR so the Jenkis breathes for a while. Let's call it once the test pass locally.

JIMENOFONSECA:
@martin-mosteiro @gabriel-happle this is the error in Jenkins. You will need to update the unittests and test locally before pulling 
image

gabriel-happle:
Is the jenkins doing anything even if we don't commit any changes?
The error above should be already addressed in the other PR #1790.
What to do here depends on what PR is merged first.
@martin-mosteiro do you want to update the schedule tests here?

daren-thomas:
@gabriel-happle the jenkins only runs the tests when you commit changes to the pull request

martin-mosteiro:
after discussion with @daren-thomas, we decided test_schedules is quite basic but it's doing its job, so there's no need to change the schedules themselves - just update the expected results as @gabriel-happle did

gabriel-happle:
This should pass the tests now

PR #1795, opened 2019-02-14
1784 connectivity network + CEA Open Source + Goodbye ArcGIS

JIMENOFONSECA:
This makes CEA open source!

Just run the command technologies/network_layout/main.py

Probably try it out with loop and not loops.

Check how the district cooling network looks like.


shanshanhsieh:
the network in reference-case-open is 12% shorter with this update, and it looks much more reasonable!

new
old

Issue #1800, opened 2019-02-20
Check to make sure installation guides mention deprecated nature of ArcGIS interface


daren-thomas:
This includes the step `cea install-toolbox`.

Issue #1799, opened 2019-02-20
Mention deprecation of ArcGIS interface on CEA website and README.md on GitHub


daren-thomas:
The ArcGIS interface will be deprecated in near future (e.g. 2-3 Months). Communicate this via the CityEnergyAnalyst.com website as well as on GitHub.

JIMENOFONSECA:
I will put this for the 1st of May as a deadline, just to not stretch too much the release of a stable version of the dashboard, which is still our priority now.

daren-thomas:
@JIMENOFONSECA yes, this can be closed! :) good job!

Issue #1798, opened 2019-02-20
Re-add the ArcGIS interface as deprecated


daren-thomas:
A recent PR removed the ArcGIS interface. There are some reasons for not removing it quite yet, but the general idea is still to remove it mid-term. Reasons include users currently relying on that interface, published documentation/papers/tutorials, the state of the replacement interface as well as deployment in the computer rooms at ETH Zurich.

The ArcGIS interface should therefore be re-added. Separate issues will deal with the deprecation and phasing out of ArcGIS as a standard interface in the CEA.

PR #1804, opened 2019-02-21
1798 re add arcgis

JIMENOFONSECA:
This solves issues #1798 and #1799 and #1800 

To test just install the cea interface.

The new command  (following the convention of other commands e.g., cea install-grasshopper) is:

cea install-arcgis


JIMENOFONSECA:
@daren-thomas you will merge when ready right?

PR #1807, opened 2019-02-22
Calculating demand from varying setpoint temperatures

BhargavaKrishnaSreepathi:
This is the first PR in a series of PRs corresponding to CONCEPT project.

This adds in a functionality to CEA where in we can calculate the demand of every building by providing hourly set points of building in place of only the `set-point` and `set-back` temperatures.

Currently, these need to provided in `.BuildingName_temperature.xlsx` format. and the necessary columns are `time` and `temperature`.

The structure is set up in a way where the `scenario\inputs\predefined-hourly-setpoints\space heating` and `scenario\inputs\predefined-hourly-setpoints\space cooling` folders are checked for the necessary files. If the files are not present, the demand will still run, but with the default values present in CEA databases.

The way to test it is to run `data helper` - > `solar radiation` and `demand_main`.

I am also providing sample temperature files to test. Do change the values and run and test the PR

[B001_temperature.xlsx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2892673/B001_temperature.xlsx)
[B002_temperature.xlsx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2892674/B002_temperature.xlsx)
[B003_temperature.xlsx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2892675/B003_temperature.xlsx)
[B004_temperature.xlsx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2892676/B004_temperature.xlsx)
[B005_temperature.xlsx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2892677/B005_temperature.xlsx)
[B006_temperature.xlsx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2892678/B006_temperature.xlsx)
[B007_temperature.xlsx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2892679/B007_temperature.xlsx)
[B008_temperature.xlsx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2892680/B008_temperature.xlsx)
[B009_temperature.xlsx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2892681/B009_temperature.xlsx)
[B010_temperature.xlsx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/2892682/B010_temperature.xlsx)


BhargavaKrishnaSreepathi:
This also adds a new `config` parameter which is `boolean` in nature,  under the `demand` section. It reads as following `config.demand.predefined_hourly_setpoints`

daren-thomas:
```
ERROR: test_calc_thermal_loads (test_calc_thermal_loads.TestCalcThermalLoads)
  File "c:\users\darthoma\miniconda2\envs\ceatest\lib\site-packages\cea\tests\test_calc_thermal_loads.py", line 62, in test_calc_thermal_loads
    self.format_output, self.region, self.write_detailed_output, self.config)
TypeError: calc_thermal_loads() takes exactly 17 arguments (16 given)

ERROR: test_calc_thermal_loads_other_buildings (test_calc_thermal_loads.TestCalcThermalLoads)
Test some other buildings just to make sure we have the proper data
  File "c:\users\darthoma\miniconda2\envs\ceatest\lib\site-packages\cea\tests\test_calc_thermal_loads.py", line 94, in test_calc_thermal_loads_other_buildings
    self.write_detailed_output, self.debug)
  File "c:\users\darthoma\miniconda2\envs\ceatest\lib\site-packages\cea\tests\test_calc_thermal_loads.py", line 114, in run_for_single_building
    massflows_output, temperatures_output, format_output, region, write_detailed_output, debug)
TypeError: calc_thermal_loads() takes exactly 17 arguments (16 given)
```

It seems the unit test is still using an older version of the function signature - @BhargavaKrishnaSreepathi please adjust to match. From a first glance it looks as if the order of the parameters to `calc_thermal_loads()` changed and there is also a new parameter `debug`.

cea/demand/thermal_loads.py
```diff
@@ -321,13 +323,18 @@ def calc_Qhs_sys(bpr, tsd):
     return tsd
 
 
-def calc_Qhs_Qcs(bpr, date, tsd, use_dynamic_infiltration_calculation, region):
+def calc_Qhs_Qcs(bpr, date, tsd, use_dynamic_infiltration_calculation, region, building_name, config, locator):
```
daren-thomas: calculation of set points seems to be independent of any other calculations - I think it should be refactored _out_ of `calc_Qhs_Qcs`. Also, I believe the check for the  predefined temperatures file should be outside of the routine that checks for it (a separate function) so that you don't need to call the simple version from inside that. this reduces the complexity of the function signatures as well as making the flow easier.

cea/demand/thermal_loads.py
```diff
@@ -321,13 +323,18 @@ def calc_Qhs_sys(bpr, tsd):
...
     # get ventilation flows
     ventilation_air_flows_simple.calc_m_ve_required(bpr, tsd, region)
     ventilation_air_flows_simple.calc_m_ve_leakage_simple(bpr, tsd)
     # get internal comfort properties
-    tsd = control_heating_cooling_systems.calc_simple_temp_control(tsd, bpr, date.dayofweek)
-    # initialize first previous time step
+    # predefined set points for every given hour can be used to calculate the demand profile for a building
+    # a config flag is used for this, it is present in the config.demand section
+    if config.demand.predefined_hourly_setpoints:
```
daren-thomas: this clause should be `if predefined_hourly_setpoints_present(...)` and called before entering this function...

cea/demand/thermal_loads.py
BhargavaKrishnaSreepathi: Hey @daren-thomas , I do not understand your comment. Set point temperatures are further used for calculating the thermal loads. So I think the calculations should be part of the `calc_Qhs_Qcs`.

BhargavaKrishnaSreepathi:
* I do not understand why the Excel file format was chosen for this feature. Would you care to explain the reasoning?
Selecting of Excel file is from personal preference, as it is easy to modify Excel files and these files were created manually.
* Where is the documentation of the format?
The documentation of the format is present in the file `cea\demand\set_point_from_predefined_file.py`. I can add some more in the same place.
* In practice, how are these files to be created?
In practice, the MPC model (TUM CREATE) will create an excel file with the temperatures that are to be set in every building (for every hour) based on the region (i.e. a mixed building having office and residential spaces, will have two different temperatures. One corresponding to office space and one corresponding to residential space). These temperatures are then used to create a new column, which is the lowest of the set temperatures (cooling case) (As is the convention in CEA), and this creates the file of predefined-set points

cea/demand/thermal_loads.py
daren-thomas: hey @BhargavaKrishnaSreepathi what I mean is to pass the set point temperatures as a _parameter_ instead of adding all the parameters to calculate it instead. You need two functions, one computes whether you can actually do this varying setpoint calculation (true if files are present and the proper config variable is set) and the other function to perform the calculation.

That removes the reponsibility of calculating these setpoints from the `calc_Qhs_Qcs` function. Basically, a function should do one thing only.

daren-thomas:
@BhargavaKrishnaSreepathi, i checked the file for the documentation of the file format and it doesn't seem to be documented - at least, given that documentation, I don't think your example files (see PR text) could be produced ("Inside these folders the setpoints are to be provided in an excel file. The excel files need to be provided in `BuildingName_temperature.xlsx` format and the necessary columns are `time` and `temperature`").

CSV files are just as easy to edit than Excel files, but are more text-based, so play better with version control (should they ever get put into version control) and are easier to generate form a variety of other languages. Also, just as easy to consume with pandas. Do you think CSV could be used instead?

Ah, reading your last point, it seems the example files and the format specification are not totally in synch, or more, that the description could be augmented to mention the CEA convention of using the lower temperature for each zone function. Maybe a short section could be added to the Documentation regarding this? Where should we put it? Think from the user's perspective...

cea/demand/set_point_from_predefined_file.py
```diff
@@ -29,7 +29,8 @@ def calc_set_point_from_predefined_file(tsd, bpr, weekday, building_name, locato
 
     The excel files with the predefined hourly temperatures are to be stored in
     `scenario\inputs\predefined-hourly-setpoints` folder. This folder in turn has two subfolders namely `space heating`
-    and `space cooling`. Inside these folders the setpoints are to be provided in an excel file.
+    and `space cooling`. Inside these folders the setpoints are to be provided in an excel file. The excel files need
```
daren-thomas: can we keep whitespace out of folder names please? `space-cooling` and `space-heeting` would work just as well and potentially create less problems for future work.

cea/demand/thermal_loads.py
BhargavaKrishnaSreepathi: hmmm...but the original `calc_Qhs_Qcs` already had the temperature calculations within the function, I maintained the same structure. Unless you are talking of the removing the entirety of the temperature calculations, i.e. `calc_set_point_from_predefined_file` and `control_heating_cooling_systems.calc_simple_temp_control`, out of the ` calc_Qhs_Qcs`.

daren-thomas: @BhargavaKrishnaSreepathi yep. exactly. and make the set points a parameter to `calc_Qhs_Qcs`. basically, tidying up as we go. knowing how these set points were calculated is not `calc_Qhs_Qcs`'s responsibility.

docs/concept/how-to-use-predefined-hourly-setpoints.rst
```diff
@@ -0,0 +1,41 @@
...
+:Date: 26-February-2019
+
+How to calculate demand using prefined-hourly-setpoints in CEA?
+===============================================================
+
+Step 1: In the `config` file, make the `demand.predefined-hourly-setpoints` as `True`
+
+Step 2: Create `csv` files corresponding to `space-heating` and/or `space-cooling`. The files are to be named in the
+`BuildingName_temperature.csv` format. The file content should be the values of time and the corresponding set point
+temperature in the columns: `time` and `temperature`
```
daren-thomas: - this would be a good place for an "example" of the file... maybe just the first 5 rows including the header. 
- also, this documentation does not explain where to put the files. Please mention the path
- the phrase "the values of time" is a bit vague. Do you mean date + time values for each hour of the year? What is the format used to encode the date? your example in the PR looks like "dd.mm.yyyy", but e.g. in america "mm.dd.yyyy" is also common - how does the CEA deal with this? https://en.wikipedia.org/wiki/ISO_8601 goes into this a bit, but you might be piggy backing off of pandas here? if so, mention that, so a user can go check there instead.

PR #1839, opened 2019-03-26
1811 fix module doc strings

Jack-Hawthorne:


daren-thomas:

PR #1845, opened 2019-03-29
I1836 network layout

JIMENOFONSECA:
@martin-mosteiro kindly, cross-check this. The PR is ready to merge though

### DIFF
--- a/cea/technologies/thermal_network/network_layout/connectivity_potential.py
+++ b/cea/technologies/thermal_network/network_layout/connectivity_potential.py
@@ -349,7 +349,7 @@ def one_linestring_per_intersection(lines, crs):
         lines_merged = lines_merged.intersection(bounding_box)
     except:
         #if the bounding_box fails, then revert to lines merge.
-        print('bounding box method did not work, falling to more simple method, no need to worry')
+        print('bounding box method did not work, falling to a more simple method, no need to worry')
 
     # merge the result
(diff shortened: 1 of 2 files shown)


martin-mosteiro:
Tested the following:
1. Ran the network layout street with a shapefile that had disconnected streets. The following warning appears, which is nice: `bounding box method did not work, falling to a more simple method, no need to worry`. After that, the folder path to the networks was displayed and finally the following error message: `ValueError: There was an error while creating the Steiner tree. Check the streets.shp for isolated/disconnected streets (lines) and erase them,the Steiner tree does not support disconnected graphs.` This is very good, as it gives a clear explanation of what the issue is.
2. Removed the disconnected streets and ran the script again. Again, the bounding box warning appeared. The resulting network does not have a specific coordinate system, though:
```
In[6]: gpd.read_file(r'C:\Users\User\Desktop\1_Baseline\inputs\networks\DC\edges.shp').crs
Out[6]: {}
In[7]: gpd.read_file(r'C:\Users\User\Desktop\1_Baseline\inputs\networks\DC\nodes.shp').crs
Out[7]: {}
```
This was already the case with the ArcGIS version of this script, though, and the network simulation part appears to be working correctly even without a specific coordinate system.

@JIMENOFONSECA if the edges and nodes not having a specific coordinate system is okay, this PR looks okay to me.

The tests fail, but that seems to be the case for all PR's at the moment.

Issue #1005, opened 2018-01-30
Fix installation guide for Ubuntu


daren-thomas:
The prerequisites are not necessary as the guide doesn't use them. While we're at it: Go over the guide and fix up any other issues.

daren-thomas:
I think this could be as simple as updating the `environment.ubuntu.yml` file. Also, this should work similarly for euler.ethz.ch (first tests installing conda were promising, but I need to check for PREFIX as there is a limit to the amount of files I may have in my home directory)

daren-thomas:
status update: the branch `1005-fix-installation-guide-for-ubuntu` is in progress. installation of cea is working and being tested in the new `docker/Dockerfile` which creates a docker image with a fully installed cea. Stuff that still needs to be done:

- copy the `docker/environment.yml` to `environment.ubuntu.yml` (or actually export that environment once built?)
- add installation steps for Daysim (could be a script to run)
- add installation of Daysim to `docker/Dockerfile` (could be just running the script mentioned above)

daren-thomas:
@roglen do you have some input to compiling Daysim / Radiance on ubuntu? I've seen some instructions and have been able to compile a bit, but I'd like to chat with you if you have experience running the CEA on Linux.

roglen:
@daren-thomas sadly I don't think I can be of much help here. I tried to run it in the beginning, and being a complete Ubuntu newbie I couldn't figure it out. During my time using Ubuntu I always took the Radiance files from @shanshanhsieh or @BhargavaKrishnaSreepathi and just copied them in.

daren-thomas:
@roglen ok, thank you! i have been able to figure out most stuff, so i think we got this.

daren-thomas:
status update: The file `docker/Dockerfile` contains a machine-readable guide to installing CEA in a Linux environment, including building Daysim and has been tested to run the following scripts:

- `cea test`
- `cea radiation-daysim`

The documentation for installing on a clean ubuntu will need to be updated. I expect this to change anyway with the M2.12 (installer) efforts. It also depends on what steps are necessary (the docker image used is bare-bones, an standard ubuntu system might already have git & co. installed)

daren-thomas:
status update: our changes to the py4design fork have been merged to the original by kianwe and we can now continue with the development of the installation guide and the Dockerfile.

PR #1854, opened 2019-04-02
1005 fix installation guide for ubuntu

daren-thomas:
This pull request revises the installation guide for Ubuntu. I went through each step to make sure it works. There is still an issue with running `cea test` to completion, but we're most of the way there and the goal of the issue #1005 (to fix the installation guide) is met.

To test this, follow the guide step by step on an Ubuntu (18 LTS) installation - either on a dedicated machine or in a virtual machine (if you're using Docker, e.g. on Windows, you already have hyper-v installed as a requirement and can create it there. That's what I used).

Please also check for spelling errors / grammar etc.

It also adds a Dockerfile for building a Docker container that can run the CEA. That can be tested with the following steps:

- `cd docker`
- `docker build -t your-user-name:cea .`
- `docker run -it your-user-name:cea cea test`

We will probably end up publishing the image to DockerHub, maybe with the tag `cityenergyanalyst:v2.12` or whatever the current version is...


JIMENOFONSECA:
@daren-thomas we will have to trust you in this one since we do not count with Ubuntu machines around.  

It would be great to assign it though to somebody who uses ubuntu. Unfortunately, the only one was Jose, and he finished already his work...

One question. Does this work is equivalent for running things in a Linux and Mac OS?

MacOS systems probably could be used the most.  I personally would prefer to work with CEA from MacOS instead of windows....

daren-thomas:
@JIMENOFONSECA I believe a lot of the ground-work done here is applicable to OS X. We'd have to try. I _do_ have an (old) macbook I could test this on. And I _will_ be setting up an Ubuntu machine for running CEA stuff in my office as soon as I find time. We're just short on hands, remember? ;-)

Ubuntu is a Linux distribution. Not all distributions use the `apt-get` package manager (e.g. RedHat uses `rpm` or used to back in the day, I'm not à-jour anymore) etc. but it will be similar to do. Besides, having the docker images might even be easier to run.

daren-thomas:
@JIMENOFONSECA are you volunteering to be our go-to mac tester? That'd be awesome!!

JIMENOFONSECA:
hehe, cool, so should I try this branch in my mac and see how it goes?




On 4 Apr 2019, 4:53 PM +0800, Daren Thomas <notifications@github.com>, wrote:

@JIMENOFONSECA<https://github.com/JIMENOFONSECA> are you volunteering to be our go-to mac tester? That'd be awesome!!

PR #1867, opened 2019-04-09
Plot cache for dashboard

daren-thomas:
This pull request implements a caching system for plots and their data (see `cea.plots.cache.PlotCache`)

Data used by plots can be cached using `PlotCache.lookup(...)` to the folder `$project/.cache/<category_name>/<data_or_plot_name>/`. The filename for the cached data is made by using a md5 hash of the sorted parameter dictionary. This guarantees unique names for the same data with different parameters.

Each plot lists it's `input_files` in the `__init__` method. The timestamps of these files are compared with the cached data and if a newer one is found, the data is re-calculated.

The `<div/>` version of the plots are cached using the same mechanism to speed up the display of the dashboard.

### DIFF
--- a/cea/config.py
+++ b/cea/config.py
@@ -647,8 +647,15 @@ class ScenarioNameParameter(ChoiceParameter):
     @property
     def _choices(self):
         # set the `._choices` attribute to the list of scenarios in the project
-        return [folder for folder in os.listdir(self.config.project)
-                if os.path.isdir(os.path.join(self.config.project, folder))]
+        def is_valid_scenario(folder_name):
+            fodler_path = os.path.join(self.config.project, folder_name)
+            return all([
+                os.path.isdir(fodler_path),  # a scenario must be a valid path
+                not folder_name.startswith('.'), # a scenario can't start with a . like `.config`
+                ])
+
+        return [folder_name for folder_name in os.listdir(self.config.project)
+                if is_valid_scenario(folder_name)]
 
 
 class ScenarioParameter(Parameter):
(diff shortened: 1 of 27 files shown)


JIMENOFONSECA:
I am trying to add the plot 'energy demand' = 'comfort-chart' and get the next error:

chrome_5TsGLSa3IW

I also tried with several of the plots of the category 'energy demand', and get a similar error

JIMENOFONSECA:
@daren-thomas also the function of adding a new dashboard is broken too....

daren-thomas:
Good testing! I'm on it!

JIMENOFONSECA:
@daren-thomas your welcome!. while I was at it I also found #1868 #1869 and #1870

daren-thomas:
It should work now. I've tested adding new dashboards, deleting dashboards (remove all plots in a dashboard to delete it), adding plots to dashboard, moving plots around. Also, made sure `.cache` does not show up as a scenario...

daren-thomas:
I think #1868, #1869, #1870 are probably unrelated to this PR - I'd like to address them in a separate branch. Feel free to merge this when you're ready :)

JIMENOFONSECA:
alright the cache seems to work, the only thing is that when I create new dashboards and then switch among them, the data seem to be reloaded (it makes it slow, especially with the the 'heating reset' diagram. 

is this an error?

daren-thomas:
@JIMENOFONSECA as far as I can tell it's not the data being reloaded, it's in plotly itself - so the way plotly works is that it creates the plot with javascript on the page. To test this, open up the heating reset schedule created by the old plots command (the html file in outputs/plots/basic/demand) and wait for it to show up in the browser. Then refresh. This is the same behavior as the dashboard - ergo not a caching problem.

Issue #1827, opened 2019-03-18
Config file is getting reset to default values after running scripts


BhargavaKrishnaSreepathi:
After running scripts, the `cea.config` present in the local computer is being reset to default values. This has been observed while running the `optimization_main.py` script, also while running few scripts in a different branch.

@shanshanhsieh:  you can add in your experience regarding the same.

@daren-thomas: Any idea why this is happening?

daren-thomas:
@BhargavaKrishnaSreepathi how do i reproduce this behavior? Can you please provide exact steps?

BhargavaKrishnaSreepathi:
1. For a cooling case study (take one of the singapore case studies).
2. Change the config file representing the cooling scenario. Make the flag `district-cooling-network` as `True`. Replace all the `DH` in the config file with `DC`. 
3. Save the config file and close it.
4. Run `cea\optimization\optimization_main.py` or run `cea\plots\optimization\main.py`. We observed the behaviour while we ran these scripts.

daren-thomas:
@BhargavaKrishnaSreepathi what exactly got changed? All the values? Some specific ones? How did you run the scripts?

BhargavaKrishnaSreepathi:
@daren-thomas: Everything in the `general` section of the config file got changed to default values. I ran the scripts from Pycharm (not the command prompt)

JIMENOFONSECA:
@daren-thomas I have the same problem when I run a new installation of CEA. all values get reset.

@daren-thomas I also see a quirky behaviour. When I do a fresh installation of CEA, I have to run a random program to make the cea.config show up. But should not it be ready after installing it?

daren-thomas:
@JIMENOFONSECA what exactly do you view as "run a new installation of CEA"? what are the steps?

regarding the `cea.config`: that is created on first access. We don't really have an installer. That's where this file would be created.

JIMENOFONSECA:
"run a new installation of CEA". I mean, I do a fresh installation of CEA.
________________________________

daren-thomas:
to clarify: you mean, follow the [installation guide](https://city-energy-analyst.readthedocs.io/en/latest/installation-on-windows.html) steps 1-10?

JIMENOFONSECA:
Sorry, I mean these two steps:


  *   Conda env update
  *   Pip install .[dev]

_______________________

Dr. Jimeno A. Fonseca
Senior researcher and project coordinator
Future cities Laboratory
[...]

PR #1872, opened 2019-04-11
1827 config file is getting reset to default values after running scripts

daren-thomas:
While trying to reproduce #1827, I made some small changes to optimization_main.py - cleaning up. This is a really small PR, but I guess we could do a lot of these to tidy up optimization and get it ready for cloud.

BTW: I can't reproduce #1827. I do not see the config file being reset no matter how many times I run optimization or optimization plots.

### DIFF
--- a/cea/optimization/optimization_main.py
+++ b/cea/optimization/optimization_main.py
@@ -2,7 +2,9 @@
 multi-objective optimization of supply systems for the CEA
 """
 
+from __future__ import print_function
 from __future__ import division
+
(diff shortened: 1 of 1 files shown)


JIMENOFONSECA:
@daren-thomas does the optimization run in your computer with this branch? let me know if it does... thanks (a couple of generations will do)

daren-thomas:
yes. it does run. i haven't run it to completion, but it runs.

PR #1879, opened 2019-04-11
added the schedule information to tutorials

Jack-Hawthorne:
basically just summarised and copied @martin-mosteiro documentation into rst format

hope it looks ok


daren-thomas:

martin-mosteiro:
@Jack-Hawthorne Thanks for that. "The only parameter that needs to be estimated is μ, which we randomly draw for each occupant ~from a normal distribution~." It's actually not drawn from a normal distribution, so I would delete that bit.

daren-thomas:
good job! thanks!

Issue #1813, opened 2019-03-01
Add a zone helper tool to allow users to create zone shape file from a polygon


JIMENOFONSECA:
I have been trying to work out to run a project in the cea dashboard (to be renamed) from scratch. It seems we are lacking an essential feature: "A landing page" for the dashboard.

The idea of this "landing page" is to point users to the first thing they should do once they open CEA:

1. Select or create a new project in CEA.

I created a diagram to present how the workflow could be, and the series of panels and "views" we should create when somebody opens the CEA dashboard.

The first panel in the next figure is what the landing page could be. The rest is the workflow to where the landing page should take users to.

![screen shot 2019-03-01 at 3 00 33 pm](https://user-images.githubusercontent.com/8973186/53621820-c9039f80-3c32-11e9-9c8b-6819449eafa5.png)

JIMENOFONSECA:
https://idethz-my.sharepoint.com/:o:/g/personal/ubjt71fbg4_ethz_ch/Es0fFNX5BEJIsW6llzWVWkcBV5TpDQrM17p3_TJbyKwB8Q?e=WTpKv4

PR #1880, opened 2019-04-11
1869 dashboard plot settings are not working

### LINKED ISSUE #1869: dashboard plot settings are not working
While working inside the plots, I noticed that they are not being recalculated when I change their settings.

I am running the case study that bhargava sent us to test optimzation MIX_low_density.

This is an example for the 'Energy Balance Plot'

1. First I ask CEA to show me the accumulated energy balacne for all buildings

chrome_riajHxdDZg

2. The results show me just the balance for B001, But it should be for the entire district

chrome_JW2tZ4PXeH

3. Now I try to ask CEA to show me the results for any other building B009

chrome_6D53i6VCOL

4. The results are exactly the same as before. Notice the scale in the Y axis. It is about the same This means that the chart is not updating.

chrome_mukNam8qR1

daren-thomas:
This fixes #1869 for plots in the demand category that are only supposed to work for a single building.

To test it, create a dashboard with two (or more) of these three plot types:

- demand/energy-balance
- demand/comfort-chart
- demand/heating-reset-schedule

Change the building in the parameters. Observe new results (they only differ by a little in the scenario I checked - WTP_CBD_h)


cea/default.config
```diff
@@ -869,6 +869,10 @@ buildings =
 buildings.type = BuildingsParameter
 buildings.help = a list of buildings to run graphs or leave blank to simulate district
 
+building =
+building.type = SingleBuildingParameter
+building.help = the building to plot
+
```
JIMENOFONSECA: @daren-thomas mmm this might create problems if the user selects values for 'buildings' and 'building'. I would prefer if we keep the buildings parameter only. Perhaps there could be a flag inside the plots to indicate that they for only one building at the time. however, most work for several buildings. Should we skype over this?

daren-thomas: @JIMENOFONSECA actually, the user only sees the `building` parameter in the dashboard for the single building plots. The other plots only see the `buildings` parameter. I don't see an issue here.

JIMENOFONSECA: the issue is when we access the local config file from %home% to use CEA in pycharm. it is confusing to have two variables. I imagine this to be the case, when calling the plots from  tools/plots

daren-thomas: I actually think we should work towards removing the plots* scripts from the cea / tools section of the dashboard. I just haven't removed them yet since we haven't converted all plots. 

The problem being solved here is, for single building plots, _what_ plot is being produced? Explicit is better than implicit: Implicitly the first one in the list? Having the building a separate type (SingleBuildingParameter) also allows to display a more appropriate control in the dashboard.

What about renaming the parameter?

Also, I'm don't think optimizing for running the CEA from PyCharm is the way to go. As per our website, the dashboard is the official interface. And we need to start using it internally as well, because otherwise it will never be tested enough. Running a script that's under development from PyCharm is of course only logical :)

daren-thomas: We could also update the help for the two parameters to explain?

JIMENOFONSECA: I see your point. However, I would not trust on removing the tools yet, nor on refactoring at least until we finish the dashboard. To me, we are still in the early stages.  3D visualization is needed, A workflow from the starting to end of the analysis is needed e.g., #1813 .

+ the plots tool will be still necessary since they produce tables. So in the meantime, it would be good to do the integration of the dashboard without making it harder to run it from the other interfaces.

daren-thomas: Sure. That's why I haven't removed the tool yet. Also, we should add the tables to the plots! I was hoping to get to that at some stage after converting all the plots. I was very optimistic about how long it would take to get there. Creating an issue right now! #1882 :)

I don't understand this part:

> But to make it consistent, it needs to be renamed in all the plot tools.

Only the [plot] tool currently has the building parameter. What do you mean? Also, do you have a good suggestion for what to rename the parameters? I really think keeping them as they are is a good option.

(part of the problems we have here - and that i had to solve separately for the plotting framework - is that the `[plots]` section is actually a configuration for multiple tools, each plot being it's own tool in a sense)

JIMENOFONSECA: I was refering to your point about: What about renaming the parameter? 

I do prefer to keep them as they are.

I see the issue. However, I am not convince that adding a new variable could solve it. I see it as a source of problem in the near future (while we partly use pycharm and the plot tools).

So maybe we can keep the functionality of until now of the plots? and rather make the trick inside the interface just for these plots?

I imagine you can create a flag, than when a plot only uses single building data, then you show a list instead of a multiple choice right?

OR

what about changing the functionality of that building parameter to something like this:

building =
building.type = SingleBuildingParameter
building.help = Indicate the building ID for the Energy Demand, Comfort Plot and heating-reset-scehdule Plot, If empty we use the first building in the 'buildings variable' or your district.

Then I could live with the two parameters under the same umbrella.

daren-thomas: I too can live with this solution. I will implement the change today.

daren-thomas: @JIMENOFONSECA I just looked at the code again... The dashboard plots are a bit separate from the `cea plots` script. The functionality is not changed at all. So I updated the help text to indicate that only the dashboard plots are affected. The way the original plots structure is set up, it is very hard to make a small change. Once we have the plots all converted, we can mimic the plots tool functionality with a script similar to the `if __name__ == __main__` part of `cea.plots.categories`. I would like to merge as is.

PR #1885, opened 2019-04-12
Restrict length of year to 8760 hours

martin-mosteiro:
The length of a year is hard-coded as 8760 hours, which is reasonable for a "standard" year.

When running a simulation with measured weather data, however, the measured data might be longer than that - if it's a leap year.

The most practical solution (and consistent in order to make results comparable cross-year) is to force the weather data to be 8760 hours long.

### DIFF
--- a/cea/utilities/epwreader.py
+++ b/cea/utilities/epwreader.py
@@ -28,7 +28,7 @@ def epw_reader(weather_path):
                   'snowdepth_cm', 'days_last_snow', 'Albedo', 'liq_precip_depth_mm', 'liq_precip_rate_Hour']
 
     result = pd.read_csv(weather_path, skiprows=8, header=None, names=epw_labels).drop('datasource', axis=1)
-    result = result.loc[0:8759]
+    result = result.loc[0:HOURS_IN_YEAR]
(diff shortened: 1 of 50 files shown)


martin-mosteiro:
Makes sense... should that be implemented with this PR?

martin-mosteiro:
@daren-thomas I added the change you suggested.

cea/utilities/epwreader.py
```diff
@@ -27,8 +28,9 @@ def epw_reader(weather_path):
                   'snowdepth_cm', 'days_last_snow', 'Albedo', 'liq_precip_depth_mm', 'liq_precip_rate_Hour']
 
     result = pd.read_csv(weather_path, skiprows=8, header=None, names=epw_labels).drop('datasource', axis=1)
-    result['date'] = pd.Series(pd.date_range(str(result["year"][0])+"/1/1", periods=8760, freq='H'))
-    result['dayofyear'] = pd.date_range(str(result["year"][0])+"/1/1", periods=8760, freq='H').dayofyear
+    result = result.loc[0:8759]
```
daren-thomas: um... to be consistent, this should actually be `result.loc[0:HOURS_IN_YEAR-1]`, right?

martin-mosteiro: You're right. Done!

Issue #1901, opened 2019-04-23
Ensure there is at least one occupant per building function


martin-mosteiro:
The case study in Almere includes some unorthodox homes that are single-occupant lofts of 14 m2. Since the area is less than the assigned area for a single-family home according to the CEA archetypes, there are no occupants all year round. Hence the cooling systems are never turned on and the building overheats in summer.

This could be solved by adapting the occupancy schedule such that each use in the building for which there is actual occupants (i.e., not cool rooms, server rooms, etc.) should have at least one occupant in the building occupancy schedule.

martin-mosteiro:
Taken care of in PR #1903

PR #1903, opened 2019-04-23
Make each function that has occupants have a minimum of one occupant

martin-mosteiro:
As described in PR #1901, it makes sense to ensure that any building function that should have occupants should have at least one occupant, even if the m2 of this function in a given building is less than the archetypal occupant density (i.e., if there are 20m2 of residential space in a building, then there should be at least one occupant even if the archetypal occupant density is 60 m2/person).

This does not impact _reference-case-open_ since none of the building has a function with a very small share of the total occupied area.


gabriel-happle:
I updated from master.
@martin-mosteiro, I don't really understand? This is correcting the occupant capacity or the occupancy at every hour?

martin-mosteiro:
@gabriel-happle Thanks for the feedback. Since the number of people at each hour is a positive integer, both statements are exactly the same (i.e., if sum(x) == 0, then max(x) < 1 and vice versa).

However, I agree that your statement is more intuitive, so I'll change that line accordingly.

PR #1905, opened 2019-04-23
Add floor cooling to CEA systems

martin-mosteiro:
Floor cooling is added to the emissions systems database as T5 with operating temperatures of 18/21 (same as ceiling cooling) due to comfort reasons. The cooling power is based on @FionaCollins's thesis. The calculation of the temperatures in the RC-model is based on the method described in SIA 2044 (i.e., fhc,cv = 0.1 for floor cooling).


martin-mosteiro:
I'm assigning @gabriel-happle as a reviewer since this mostly affects the demand calculations.

gabriel-happle:
I updated from master.
If the tests are still passing I suggest we merge this.

gabriel-happle:
I tested this with the reference-case-cooling (region SG). It works great.
I'm advising to merge this when all the checks are completed.

Issue #1828, opened 2019-03-20
Make one-for-all installer for windows


JIMENOFONSECA:
After some discussion with @daren-thomas in the Epic, I believe we should go the executable route. At the best we develop a Cross-platform installer of CEA.

WORKFLOW
The installer should prompt to the user the next option:

Indicate where CEA will be installed.
Then the installer should ask the user to select one option of the next:

1. Repair existing installation of CEA
2. Do a fresh installation of CEA.
3. Update to the latest version of CEA.
4. Install the Developer version of CEA.

OTHER REQUIREMENTS

the installation process should be fast ( less than 10 min)

DELIVERABLES
Installer of CEA for Windows.

daren-thomas:
The installation process being less than 10 minutes is probably an unrealistic goal. We should talk about options. I have some ideas.

JIMENOFONSECA:
Sure, lets put this point in the agenda of next monday’s meeting. Would 4:00pm singapore work for you?

_______________________

Dr. Jimeno A. Fonseca
Senior researcher and project coordinator
Future cities Laboratory
Singapore ETH-Centre

On 20 Mar 2019, 17:26 +0800, Daren Thomas <notifications@github.com>, wrote:

The installation process being less than 10 minutes is probably an unrealistic goal. We should talk about options. I have some ideas.

daren-thomas:
i sent you a calendar invite :) see you then!

JIMENOFONSECA:
Folder structure:
(inside the Documents folder)

-CEA
---Repository
---Projects

gabriel-happle:
Separate issues were created: #1916 #1915 #1914 #1913 #1911

PR #1908, opened 2019-04-24
1828 make one for all installer for windows

daren-thomas:
This PR implements an installer for Windows.

Try this: https://www.dropbox.com/s/3l67vmd90f7pn6m/Setup_CityEnergyAnalyst_2.12.0a3.exe?dl=0

I'll add more information on how to create the installer (setup NSIS etc.), what it does exactly and documentation for the website tomorrow. Just a sneak peak for now :)

(check out the start menu when done for the nifty cea shortcuts)


JIMENOFONSECA:
A couple of comments while testing: 
1. we might want to put the credits in here and the year of the copyright from 2016-2019
image

JIMENOFONSECA:
Another issue is that the icon has a white background. It should be transparent right? Take a look at the corners. thanks!

image

JIMENOFONSECA:
I believe I had implemented the option to open automatically the cea interface. Now when I click CEA dashboard it opens this. 

I believe it should directly open the browser (no the console). We can open the console with other executable as I see, right?

image

JIMENOFONSECA:
4th issue. The last three issues where found while running CEA with development version installed. I also did a separate installation and saved it in the folder CityEnergyAnalyst2.

I run the console and try to run the dashboard. It does not work. 
image


I also tried to run the CEA dashboard directly form the executeble. and it does not work (it opens the console and closes it quickly)

JIMENOFONSECA:
5th Issue: Daysim does not work. I think this has to do with the fact that C:/Daysim/bin (a parameter inside the solar radiation tool. does not exist anymore. I believe CEA should then know that daysim is installed in the cea/dependencies folder?

image

daren-thomas:
@JIMENOFONSECA thank you for the feedback! yep. there are still a couple of issues. btw: the license agreement is taken from the LICENCE file in our root folder. that needs to be updated too! i have created a separate issue for that (#1912). The credits could be shown on a separate page. But I don't think they improve the usability and nobody is going to check the installer to figure out the credits, so i think that's maybe just a vanity project.

I have asked @shizhongming on slack to provide a gradient towards dark - the background really is transparent, but the icons have a gradient towards the background. maybe the gradient is not necessary at all...

regarding 3rd issue: that's actually fixed in this pull request already - but you'd have to use the development version and switch that repo to this branch for the changes to be visible - a bit of a chicken and egg problem that maybe easier to solve with a quick merge and then work out the issues as separate issues incrementally?

regarding 4th issue: i have created an issue for that. (#1913)

regarding 5th issue: yes. i am aware of this. the installer bundles daysim binaries, but i'm not sure if i have everything yet. (#1914)

Since these issues don't really break existing things in the CEA, I propose merging this PR, but not closing the issue, instead, makinging #1828 dependent on these issues being resolved. That way, we can have smaller pull requests, each addressing a specific issue.

PR #1919, opened 2019-04-29
1813 Zone Helper

JIMENOFONSECA:
This PR solves issue #1813 which extracts the Zone.shp file from OSM.

It also creates the occupancy and the age file.

This is part of some work we need to do to create the "new project" feature inside the new CEA dashboard.

How to test:
1. uise the site.shp file stored in building_geometry
2. run zone_helper.py
3. check that the outputs: zone.shp, occupancy.dbf and age.dbf are created

@reyery please check that it runs in your computer.


reyery:
@JIMENOFONSECA i have tested the zone_helper.py tool and successfully created the 3 files. Should be able to merge to master but might need to change back the default route of the dashboard that I have accidentally changed in a commit, 90efb96, so that the dashboard can work as per normal when merged

JIMENOFONSECA:
@reyery alright, please go ahead with the changes and merge. thanks!

Issue #1924, opened 2019-05-02
Create case study for executive course


JIMENOFONSECA:
This entails finding an area, potentially in Zurich of about 20 buildings, where to run simulations, including all the stack until optimization.

Issue #1920, opened 2019-04-29
As user i would like to get the occupancy file from OSM information


JIMENOFONSECA:
This can be an improvement to #1813 where the parameters inside the occupancy file can be filled in with OSM data

PR #1929, opened 2019-05-03
adding automatic occupancy of site

JIMENOFONSECA:
@reyery this solves issues #1924 and #1920, 

To test, just use the site shapefile and run the zone_helper. It should create the zone.shp, occupancy and age.dbf based on data in Open street maps


reyery:
@JIMENOFONSECA running zone_helper creates the 3 files on my end. Is there anything that I should take note regarding the generated files?

Issue #1921, opened 2019-04-29
As a user I would like to create a new project/select my scenario as soon as I start the CEA dashboard


reyery:
https://github.com/architecture-building-systems/CityEnergyAnalyst/issues/1813#issue-415968438

A landing page like the one mentioned above could be provided to the user, so that they can manage different projects such as creating a new project or opening an existing project.

Issue #1917, opened 2019-04-29
Allow user to create zone shapefile from map during project creation


reyery:
With the help of a map when creating a new project, the user should be able to create a zone shapefile based on a selection on the map using a lasso like tool. This can be achieved with the help of the zone helper tool.

PR #1931, opened 2019-05-06
1917 create zone shapfile from map

reyery:
This PR solves #1917 which aims to allow users to create a zone.shp file based on a selection from a map when they create a new scenario. Besides the map tool, a basic landing page interface/flow is added to the dashboard interface, which allow users to access a 'project management' page. The user experience of the landing page could be further improved in #1921.

With the added features, users should now be able to:
- Create a new project in the landing page
- Open an existing project in the landing page
- Create a new scenario in the current project
- Create zone.shp file using the map tool when creating a new scenario
- See the current active project and scenario in the dashboard

To test it:
- Create a new project from the landing page
- Create a new scenario and select the checkbox to create the zone.shp file
- Use the map tool to select a zone

The zone.shp file should be created in the new project directory, together with the age and occupancy files that were generated with default parameters.


JIMENOFONSECA:
a small change, maybe we can split the box into one that says latitude and the other longitude? people usually confuse which one is which :-), I do :-D

image

PR #1939, opened 2019-05-06
Split coordinates to lat lon input

### LINKED ISSUE #1934: separate lat from lon in new project page
a small change, maybe we can split the box into one that says latitude and the other longitude? people usually confuse which one is which :-), I do :-D

image

reyery:
This PR resolves #1934. It separates the coordinates input in the map tool to two separate latitude and longitude inputs.


reyery:
@JIMENOFONSECA I have added a 'go' button. Should be working now.

PR #1944, opened 2019-05-07
Add data_source to input table in dashboard

### LINKED ISSUE #1935: databases lack some arguments in the dashboard
The next databases are missing a new attribute that helps the user to know where the data was extracted from.

The attribute is called "data_source" it is of type txt.

The next databases lack it in the dashboard:

zone.shp,
district.shp
age.shp
occupancy.shp

image

reyery:
This PR resolves #1935. The 'data_source' column is now added to the inputs.yml for the four databases. It should also be noted that the columns in shape files are truncated to 10 characters due to specification of the format.

### DIFF
--- a/cea/datamanagement/district_geometry_helper.py
+++ b/cea/datamanagement/district_geometry_helper.py
@@ -130,9 +130,9 @@ def clean_attributes(shapefile, buildings_height, buildings_floors, key):
         # Check which attributes the OSM has, Sometimes it does not have any and indicate the data source
         if 'building:levels' not in list_of_columns:
             shapefile['building:levels'] = [3] * no_buildings
-            shapefile['data_source'] = "CEA - assumption"
+            shapefile['REFERENCE'] = "CEA - assumption"
         else:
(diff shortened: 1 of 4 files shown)


reyery:
@JIMENOFONSECA this currently does not work for the occupancy data since it does not contain the 'data_source' column, otherwise there should not be any problems.

JIMENOFONSECA:
@reyery one question, is it possible that we do not show the column if it does not exist? but still have full functionality. I see that some user will have the column (those who used our automated way. The others will not have it.

What do you think? Another way is to create a button in the panel to help the user know where does the data come from in every row (that is what the data_source column is doing, nothing else

daren-thomas:
I'd like to add 2 cents here: I have this mid-term goal of formalizing the schemas of all input and output data - This seems like it would make that harder, by adding an "exception".

Instead, why don't we just require the data_source column, if we really want it. It could default to `<null>` values.

reyery:
Added lines to check for missing columns to set them to null, should work with occupancy now.

PR #1959, opened 2019-05-14
1942 fix default dashboard plot

### LINKED ISSUE #1942: Delete default dashboard and stop loading/creating plots by default
So the CEA dashboard creates random plots once we start a scenario.

It would be good to stop this behavior. Let the user go to the Dashboard and create a new dashboard on its own.

This of course, after running some tools.

### LINKED ISSUE #1946: zone.cpg file being created when creating new project with dashboard
When creating a new project using the dashboard interface, a zone.cpg file will be created in the new project path.

For example, if the new project is saved as ` C:\new-project\`, a zone.cpg file will be created in the folder structure ` C:\new-project\baseline\inputs\building-geometry\zone.CPG`.

This also happens as long as the project in the current config does not exist, when the page is loading.

i.e. If the project in the current config ` C:\old-project\` does not exist anymore, loading another page would still create ` C:\old-project\baseline\inputs\building-geometry\zone.CPG`.

reyery:
This PR resolves #1942 and resolves #1946. The dashboard does not create a default dashboard and plot at every page load anymore.

New features include:
- Able to create and delete dashboards from the new dashboard management page
- Able to delete scenarios from the project overview page


reyery:
The dashboard is still not able to handle the exception when a new plot is created if the required files are not present. Will fix it in a new issue.

reyery:
@JIMENOFONSECA I have also added the new workflow for opening and creating new projects in this branch that you have mentioned earlier.

reyery:
Added geocoder to the map when creating zone file. User is now able to search for a location and the coordinates would be auto generated.

PR #1963, opened 2019-05-14
1868 dashboard file directory opener is not working

### LINKED ISSUE #1868: dashboard file directory opener is not working
@daren-thomas I am trying to change the project dir, but I cannot navigate to any folder outside of the current parent directory (E:/). The project i want to use is in C:/. 

I think CEA should allow to locate files in any hard drive.

image

daren-thomas:
This PR implements a drive selection dropdown for file- and folder dialogs in the CEA. This fixes #1868.

To test it, start the dashboard and change project location - you should be able to choose the drive on windows.


JIMENOFONSECA:
@reyery kindly review this PR, if this works please merge

cea/interfaces/dashboard/tools/templates/folder_listing.html
```diff
@@ -16,6 +16,15 @@ <h4 class="modal-title" id="cea-folder-dialog-title">{{title}}</h4>
     </ol>
   </nav>
 
+  {% if drives %}
+  <select class="form-control cea-drive-select" id="cea-drive-select"
+          onchange="folder_dialog_navigate_to('{{fqname}}', $(this).val(), '.')">
+    {% for drive in drives %}
+    <option value="{{ drive.lower() }}" {% if breadcrumbs[0][:2].lower() == drive.lower() %} selected {% endif %}>{{ drive }}</option>
```
reyery: Could change the drive to include the trailing slash i.e. `C:\` since `os.path.abspath` changes the input `c:.` to the working directory

```suggestion
    <option value="{{ drive }}\" {% if breadcrumbs[0][:2].lower() == drive.lower() %} selected {% endif %}>{{ drive }}</option>
```

JIMENOFONSECA: @daren-thomas I appreciate if you can check this...

daren-thomas:
Thank you @reyery - I was able to reproduce the bug you mentioned and accepting your fix solved it!

@JIMENOFONSECA I think we're good to go.

reyery:
@daren-thomas I just noticed that the weird behavior seems to affect file listing also, but I was unable to fix it with the same change to the folder listing.

JIMENOFONSECA:
@daren-thomas @reyery solution?

reyery:
@daren-thomas ok, so the fix for the folder_listing actually works for the file_listing. I found out that the problem that I faced was that the drive dropdown actually shows another option `E:` when my laptop only has one drive. Apparently, scanning for logical disk on the macbook running windows shows the card reader as a logical disk. Not sure how you want to handle this.

daren-thomas:
@reyery doesn't that just mean that the card reader is a drive? doesn't windows explorer also show that drive? does showing it present a problem?

reyery:
@daren-thomas true, but windows explorer does not show the drive when there is no card inserted, though it technically exists. It still shows up in the drive selection when it does nothing.

test

daren-thomas:
@reyery i think we can just not care about that. thank you for pointing it out! (from your .gif, it looks like it just reverts back to `C:\`, right? merging.

Issue #1896, opened 2019-04-18
Blog entry for Milestone M2.13


daren-thomas:
The Blog should be about how to use Grasshopper and CEA. Potentially, only about creating the Zone.shp file and the occupancy and age files. The rest (district, streets, terrain) can be done with the blog published in M2.12

JIMENOFONSECA:
In reference to issue #481 we will give it a shot and include video tutorials in all Blog entries.

shizhongming:
I am creating a series of blogs on CEA at Grasshopper. The first blog is about the installation of the CEA Grasshopper interface, producing .dbf files. Yet need to include a video.

shizhongming:
@daren-thomas I just finished the first blog on Squarespace....could you have a look? Thanks.

daren-thomas:
@shizhongming how do i access it?

shizhongming:
@daren-thomas Jimeno told me you have credentials to the Squarespace..nvm, i just made it public. Please use this link.
 https://cityenergyanalyst.com/blog-1

daren-thomas:
hey @shizhongming ah. i understand. yes. i probably should have figured that out on my own. sorry.

This is fantastic! I _love_ it!

I do have some small comments:

- the workflow for the installation is different with the introduction of the installer: 
  - step 2 should be: launch the CEA Console from the Start menu
  - step 3 is not required
  - step 4, formatting of indented list
- in "Creating .dbf files", the sentence "If they are renovated, keep the corresponding cells as “0“." should probably be "If they are *not* renovated"
  - check formatting of nested lists

shizhongming:
@daren-thomas thanks for the comment!
However, the `cea install-grasshopper` is currently not working under the cea console. the errors are as follows. could you have a look? Thanks~ (I unpublished the blog for the moment)

C:\Users\Zhongming\Documents\CityEnergyAnalyst
λ cea install-grasshopper
City Energy Analyst version 2.13.5a1
Running `cea install-grasshopper` with the following parameters:
- general:debug = False
  (default: False)
Traceback (most recent call last):
  File "C:\Users\Zhongming\Documents\CityEnergyAnalyst\Dependencies\Python\Scripts\cea-script.py", line 11, in <module>
    load_entry_point('cityenergyanalyst', 'console_scripts', 'cea')()
[...]

PR #1973, opened 2019-05-17
Fix for cea install-grasshopper

daren-thomas:
A tiny Bugfix for the `cea install-grasshopper` command: I did a `git log --full-history -- cea/concept_parameters.py` according to [this answer on stackoverflow](https://stackoverflow.com/questions/6839398/find-when-a-file-was-deleted-in-git) - I'm pretty sure that file never ever existed and explains the bug @shizhongming is having with https://github.com/architecture-building-systems/CityEnergyAnalyst/issues/1896#issuecomment-493307814

@shizhongming would you mind testing this to see if it solves the problem? thank you!


shizhongming:
@daren-thomas just tested... `cea install-grasshopper` works fine now...as Jimeno has already merged~ ;)

Issue #1987, opened 2019-05-21
CEA tests don't run 


gabriel-happle:
Console tests stop after calibration. `dodo` on pycharm continues running.

CONSOLE OUTPUT:

```
...
>Metropolis: [var2]
>Metropolis: [var1]
100%|##############################################################################################################################| 1100/1100 [00:07<00:00, 151.45it/s] 100%|##############################################################################################################################| 1100/1100 [00:05<00:00, 213.26it/s] The number of effective samples is smaller than 25% for some parameters.
This application failed to start because it could not find or load the Qt platform plugin "windows"
in "".

Reinstalling the application may fix this problem.
```

PYCHARM OUTPUT:

```
...
>Metropolis: [var2]
>Metropolis: [var1]
100%|##########| 1100/1100 [00:06<00:00, 170.20it/s]
100%|##########| 1100/1100 [00:05<00:00, 212.48it/s]
The gelman-rubin statistic is larger than 1.05 for some parameters. This indicates slight problems during sampling.
The number of effective samples is smaller than 25% for some parameters.
.  run_emissions_operation:open
.  run_thermal_network

...
```

gabriel-happle:
Actually, I take it back. After pointing pycharm to the environment created with the installer, the `dodo` does not work either.

gabriel-happle:
output from pycharm in master

```
.  run_calibration
Only 100 samples in chain.
Sequential sampling (2 chains in 1 job)
CompoundStep
>Metropolis: [sigma]
>Metropolis: [var6]
>Metropolis: [var5]
>Metropolis: [var4]
>Metropolis: [var3]
>Metropolis: [var2]
>Metropolis: [var1]
100%|##########| 1100/1100 [00:09<00:00, 111.20it/s]
100%|##########| 1100/1100 [00:05<00:00, 191.50it/s]
The number of effective samples is smaller than 10% for some parameters.

Process finished with exit code 1
```

daren-thomas:
```
calculating thermal hydraulic properties of DH network  ...  time step 179
calculating thermal hydraulic properties of DH network  ...  time step 183
calculating thermal hydraulic properties of DH network  ...  time step 190
calculating thermal hydraulic properties of DH network  c.alculating thermal hydraulic properties of..  time step  D175H
 network  ...  time step 193
calculating thermal hydraulic properties of DH network  ...  time step 187
calculating thermal hydraulic properties of DH network  ...  time step 196
[...]

PR #1997, opened 2019-05-22
1987 fix qt conf for installer

daren-thomas:
This PR addresses #1987 - `cea test` was not running through on a fresh install. The issue has to do with the QT_PLUGIN_PATH. The solution was to update the `CityEnergyAnalyst\Dependencies\Python\qt.conf` file to point to the proper paths.

This new installer does just that. It can be tested here: https://www.dropbox.com/s/cgf9jkcaikzo618/Setup_CityEnergyAnalyst_2.14.exe?dl=0

@gabriel-happle would you mind testing it please?

Steps:
- download installer from link above
- install (doesn't really matter which type)
- open CEA Console
- run `cea test`
- expect it to complete

Once this is merged, we can update the binary in the release 2.14.


daren-thomas:
ok. i've updated the installer binary for the release 2.14.

PR #2010, opened 2019-05-27
Import input files during scenario creation

### LINKED ISSUE #2005: Dashboard: Create new project from import data not working
The functionality of `import input files` in the new project tool is not functional.
image

reyery:
This PR resolves #2005. It enables the function to import individual input files such as `zone`, `district`, `streets`, `terrain`, `age` and `occupancy`.

To test, go to the scenario creation page and try to import each of these files. All selected files should appear in their respective folders in the new scenario folder. If the `zone` and `district` files were chosen, their accompanied files, `*.dbf`, `*.prj`, `*.shx` and `*.cpg`, should also be copied over.


reyery:
@daren-thomas because I am using the `show_open_file_dialog` method for the file selection, could I include the path parameters in the config file even though it is not a script? Since currently I just copied the html code for the `new_project` script, and there's definitely a better way to handle this.

daren-thomas:
@reyery i guess you can just add that as a parameter to the config file... i do see how it's a bit of a hack. let's get back to that at a later time, maybe.

reyery:
@daren-thomas an educated guess as in if the user enters a path in a selection for the first time, the other selections would be auto populated from the same scenario based on the first selection?

reyery:
@daren-thomas anyway I found out that an empty `district.shp` file will be created, if it does not already exist, when District input is selected in the dashboard. I am planning to modify this method to only check if the file exists.
https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/f2df5d85a2bafcabcb0291bddf8490f2f810c792/cea/inputlocator.py#L576-L579

daren-thomas:
@reyery when I perform the steps above (creating with import method, just specifying `zone.shp` and `terrain.tif` and nothing else), I get the following error:

```
ERROR:flask.app:Exception on /landing/create-scenario/save [POST]
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\interfaces\dashboard\landing\routes.py", line 131, in route_create_scenario_save
    calculate_occupancy_file(geopandas.read_file(zone), cea_config.zone_helper.occupancy_type, locator.get_building_occupancy())
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\datamanagement\zone_helper.py", line 152, in calculate_occupancy_file
    if zone_df.loc[index, "category"] == "yes":
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\pandas\core\indexing.py", line 1472, in __getitem__
    return self._getitem_tuple(key)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\pandas\core\indexing.py", line 870, in _getitem_tuple
    return self._getitem_lowerdim(tup)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\pandas\core\indexing.py", line 1027, in _getitem_lowerdim
    return getattr(section, self.name)[new_key]
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\pandas\core\indexing.py", line 1478, in __getitem__
    return self._getitem_axis(maybe_callable, axis=axis)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\pandas\core\indexing.py", line 1911, in _getitem_axis
    self._validate_key(key, axis)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\pandas\core\indexing.py", line 1798, in _validate_key
    error()
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\pandas\core\indexing.py", line 1785, in error
    axis=self.obj._get_axis_name(axis)))
KeyError: u'the label [category] is not in the [index]'
```

ideas?

reyery:
which `zone` file did you use to import? Seems like that zone file does not have the `category` column. Mine somehow has the the `category` column and works well on my side.

reyery:
hmmm... seems like it only works for `zone` files created by the `zone-helper` as the `category` column is added by it.

reyery:
@daren-thomas we could use/update `cea.datamanagement.databases_verification` to check for the input files first?
it does not seem updated though.
https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/65b8285cfac9bbc17c3e8687d448c15de54af335/cea/datamanagement/databases_verification.py#L9-L13

daren-thomas:
so, the zone shapefile i used was produced by `cea cooling-case-workflow` and that used `reference-case-cooling` in the examples folder. what is the "category" column for?

reyery:
It was added by `zone-helper` to help create the `age` and `occupancy` files. Was probably a mistake to use the methods from `zone-helper` to estimate the files, since not all zone files are created by it.

I could reuse the methods in `create-new-project` (which probably is useless now in the dashboard interface), but they do not seem very smart.
https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/65b8285cfac9bbc17c3e8687d448c15de54af335/cea/datamanagement/create_new_project.py#L72-L79

daren-thomas:
so... the create-new-project methods will be removed, right? and replaced with the zone-helper stuff? so... we can just elevate the `category` column to standard, _or_ update the zone-helper code to add a "dummy" category column if none is in the .shp file... what do you think?

reyery:
Would making `category` standard require changing the zone files for the reference cases? I am not sure if that column is necessarily required, since it is generated from open street maps data to make a logical guess for the occupancy type of the building.

I feel that users should just edit/create a occupancy file if they want to set the occupancy for the buildings, instead of adding a `category` column into zone files that were not generated by `zone-helper`. That is unless `category` column can be useful for something else (maybe like labeling the buildings??), in that case we should just make it standard.

For now I will just set the default to `MULTI_RES` for dataframes that do not have the column, since I just realised that the method would still work without a `category` column, which sets every building to the same occupancy type.

daren-thomas:
@reyery I just checked - your changes totally work now. thank you :)

PR #2022, opened 2019-05-29
1949 3d visualization

### LINKED ISSUE #1949: Include 3d visualization and layers for showing inputs.
The idea is to include basic 3d and 2d visualization of:

1. Buildings (inputs/networks/building_geometry/district.shp and inputs/networks/building_geometry/zone.shp)
2. Streets (inputs/networks/streets.shp)
3. District heating networks (inputs/networks/DH/node.shp, inputs/networks/DH/edge.shp)
4. District cooling networks (inputs/networks/DC/node.shp, inputs/networks/DC/edge.shp)
5. District electrical networks (inputs/networks/EL/node.shp, inputs/networks/EL/edge.shp)

Functionality:
The user should be able to switch between the 2d and 3d view in the map.
The user should be able to turn on and off the layers (i.e., zone.shp, and district.shp)

@daren-thomas @reyery which technology should we use for this? should we try it inside the dashboard? or instead, use QGIS and some python scripting (3rd party software) to let the user visualize this?

reyery:
This PR resolves #1949 by replacing the Leaflet map to a Mapbox map with deck.gl.

Users are now able to:
- View `zone`, `district`, `streets` and `DC`/`DH` networks on the map
- Select the visibility of each layer
- Change from 2D to 3D view
- Get information about the object by hovering over it
- Change the map style to dark

To test, open a project with all the files needed for each layer (i.e `zone`, `district`, `streets` and `DC`/`DH` networks). Play around with different settings and make sure it performs as expected and does not slow down the computer.

Potential problems:
`DC` and `DH` networks support naming and this is not able to handle that yet, only the default networks.


reyery:
Hmmm... Not sure when test build failed. I did not change anything with the scripts.

daren-thomas:
@reyery this is the error message produced:

> Exception: Temperature in RC-Model of building B01 out of bounds! First occured at timestep = 6398. The results were Tint = 50.174614814, theta_c = 50.1430530041, theta_m = 49.7147250076, Check building geometry and internal loads! Building might be too small in size or architecture parameter Hs = 0.6 might be too small for this geometry. Current bounds of range for RC-model temperatures are between -30.0 and 50.0.

The full trace can be found here: https://ceajenkins.ngrok.io/job/run%20cea%20test%20for%20pull%20requests/1612/console

daren-thomas:
@reyery I found one issue so far: when viewing the zone, the building data shows up in a little tooltip. i _love_ this feature! my only issue with is this: clicking on a building does not reliably open the properties editor anymore. i think it might be the tooltip blocking the "click". i wonder if moving it a little bit away from the cursor could help?

cea/interfaces/dashboard/inputs/routes.py
```diff
@@ -106,6 +106,25 @@ def route_geojson(db):
...
+    # TODO: Get a list of names and send all in the json
+    name = ''
+    edges = locator.get_network_layout_edges_shapefile(type, name)
+    nodes = locator.get_network_layout_nodes_shapefile(type, name)
+    network_json = df_to_json(edges)
+    network_json['features'].extend(df_to_json(nodes)['features'])
+    return jsonify(network_json)
+
+
+@blueprint.route('/geojson/others/streets')
```
daren-thomas: why is the route `/geojson/others/streets` and not `/geojson/streets`?

reyery:
@daren-thomas currently the properties editor only works when you are in the respective input view (i.e. when in zone input, the editor only opens when clicking zone buildings which are the ones in blue). I have not touched how the input section works and it's scripts but I am intending to combine the whole thing into one eventually.

cea/interfaces/dashboard/inputs/routes.py
```diff
@@ -146,4 +165,10 @@ def route_table_post(db):
...
         cea.utilities.dbf.dataframe_to_dbf(table_df, location)
 
-    return render_template('table.html', pk='Name', table_name=db, table_columns=db_info['fieldnames'])
\ No newline at end of file
+    return render_template('table.html', pk='Name', table_name=db, table_columns=db_info['fieldnames'])
+
+
+def df_to_json(file_location):
+    table_df = geopandas.GeoDataFrame.from_file(file_location)
+    table_df = table_df.to_crs(epsg=4326)  # make sure that the geojson is coded in latitude / longitude
```
daren-thomas: `4326` should probably be a constant somewhere. hmm... [googled it](https://gis.stackexchange.com/questions/3334/difference-between-wgs84-and-epsg4326) it seems this is just a name for the reference system we use. maybe we could find other places that use this?

cea/interfaces/dashboard/inputs/static/css/deckgl.css
```diff
@@ -0,0 +1,16 @@
+#tooltip:empty {
+  display: none;
+}
+#tooltip {
```
daren-thomas: this is probably where you can specify that the tooltip is to be moved a bit to the side. idk. also, might not be the actual issue, just my guess.

cea/interfaces/dashboard/inputs/static/js/map.js
```diff
@@ -0,0 +1,274 @@
...
+var deckgl;
+var jsonStore = {};
+// Make sure the layers maintain their order
+var layers = [new GeoJsonLayer({id:'streets'}), new GeoJsonLayer({id:'dh_networks'}),
+    new GeoJsonLayer({id:'dc_networks'}), new GeoJsonLayer({id:'zone'}), new GeoJsonLayer({id:'district'})];
+var extruded = false;
+var currentViewState;
+
+const jsonUrls = {
+    'zone': 'http://localhost:5050/inputs/geojson/zone',
```
daren-thomas: use relative url's here, since the dashboard should also work with other ports / hosts (like, when we move to the cloud)

cea/interfaces/dashboard/inputs/static/js/map.js
```diff
@@ -0,0 +1,274 @@
...
+        jsonStore['zone'] = json;
+        createLayer('zone');
+        $('#zone-cb').prop('hidden', false);
+        currentViewState = {latitude: (jsonStore['zone'].bbox[1] + jsonStore['zone'].bbox[3]) / 2,
+            longitude: (jsonStore['zone'].bbox[0] + jsonStore['zone'].bbox[2]) / 2,
+            zoom: 16, bearing: 0, pitch: 0};
+
+        deckgl = new DeckGL({
+            container: 'mapid',
+            mapboxApiAccessToken: 'pk.eyJ1IjoidWJlcmRhdGEiLCJhIjoiY2pudzRtaWloMDAzcTN2bzN1aXdxZHB5bSJ9.2bkj3IiRC8wj3jLThvDGdA',
```
daren-thomas: i'm not sure we're suppost to publish api access tokens in the repository. could you please double-check this?

cea/interfaces/dashboard/inputs/static/js/map.js
```diff
@@ -0,0 +1,274 @@
...
+
+        ...options
+    }));
+}
+
+function updateTooltip({x, y, object, layer}) {
+    const tooltip = document.getElementById('tooltip');
+
+    if (object) {
+        tooltip.style.top = `${y}px`;
```
daren-thomas: ah. _here_ is where you'd move it a bit....

cea/interfaces/dashboard/inputs/templates/table.html
```diff
@@ -70,9 +68,41 @@ <h3>Inputs</h3>
                 <h2>{{ table_name }}</h2>
                 <div class="clearfix"></div>
               </div>
-              <div id="mapid" class="x_content" style="height: 500px;">
-
+              <div id="style-group">
```
daren-thomas: i think this could be styled a bit better. looking at the template [Gentelella Alela!](https://colorlib.com/polygon/gentelella/index.html) it should probably be added in the title thing on the right with an icon. what do you think? maybe use these icons:

- [fa-cubes](https://colorlib.com/polygon/gentelella/icons.html#/cubes) for 3d
- oh, idk. maybe icons are too hard to sell... instead, you could add it to the wrench icon dropdown menu? that _does_ hide it though...

image

cea/interfaces/dashboard/inputs/routes.py
reyery: `route_geojson` will catch `/geojson/streets` and throw a `404` because it's not in `inputs.yaml`. I could include it into `inputs.yaml` file or create a condition in `route_geojson` .

reyery:
> 
> 
> @reyery this is the error message produced:
> 
> > Exception: Temperature in RC-Model of building B01 out of bounds! First occured at timestep = 6398. The results were Tint = 50.174614814, theta_c = 50.1430530041, theta_m = 49.7147250076, Check building geometry and internal loads! Building might be too small in size or architecture parameter Hs = 0.6 might be too small for this geometry. Current bounds of range for RC-model temperatures are between -30.0 and 50.0.
> 
> The full trace can be found here: https://ceajenkins.ngrok.io/job/run%20cea%20test%20for%20pull%20requests/1612/console

@daren-thomas I don't have access to the jenkins server, so I can't see the full trace.

cea/interfaces/dashboard/inputs/static/js/map.js
reyery: ah yes, my mistake

cea/interfaces/dashboard/inputs/static/js/map.js
daren-thomas: @reyery regarding the api access token. did you look into this?

daren-thomas:
@reyery thank you for your changes! i'd _love_ to merge, but the API token question is still bothering me. Are we supposed to be putting that into source control? I'm not so sure...

cea/interfaces/dashboard/inputs/static/js/map.js
reyery: @daren-thomas hmm I'm not too sure how to handle this also. Technically the key is a "public" key according to mapbox. We could add the api key to a file and `gitignore` it? I'm still looking into whether we can run it without a api key, since it only provides us with the map styles.

daren-thomas:
does "public" mean anyone is allowed to use it? was it granted to your account? where did you get the key from?

reyery:
@daren-thomas I have removed the api key and changed the map to make use of the openstreetmaps raster tiles instead of mapbox tiles for the time being (does not look as good as vector tiles, but works). I think we should probably not expose the key on this repo as the key is tied to an account and has tile request limits, which can be exploited. The only way I can think of is to make use of an environment variable which could be created using the installer.

reyery:
I think we can just merge this to master first (after testing). For now I do not know what to add to it, any other additions can be added later.

cea/interfaces/dashboard/inputs/routes.py
reyery:
```suggestion
from cea.utilities.standardize_coordinates import get_geographic_coordinate_system
    table_df = table_df.to_crs(get_geographic_coordinate_system())  # make sure that the geojson is coded in latitude / longitude
```
we could change it to the current method that we have now, since they sort of mean the same thing.

PR #2023, opened 2019-05-29
Steiner tree avoids connecting buildings with no systems

martin-mosteiro:
The thermal network layout script used to connect all buildings except the ones the user asks not to be connected. This means that even buildings that do not have heating or cooling systems got connected, which made subsequent simulations slow - or made figuring out which buildings should not be connected at all cumbersome.

This PR creates a function that joins the list of disconnected buildings provided by the user with a list of all the buildings that do not have heating or cooling systems so that these don't get connected either.

This is what reference-case-open looks like for a district cooling network before this fix (below, left). However the middle building doesn't have cooling at all, so after this fix this is what the network looks like (below, middle). I finally tested what happened when passing B01 as a disconnected building. Both B01 and B05 are disconnected. Same happens when B01 and B05 are passed as disconnected buildings, as expected (below, right).

### DIFF
--- a/cea/technologies/network_layout/main.py
+++ b/cea/technologies/network_layout/main.py
@@ -6,6 +6,7 @@ import cea.inputlocator
 from cea.technologies.network_layout.connectivity_potential import calc_connectivity_network
 from cea.technologies.network_layout.steiner_spanning_tree import calc_steiner_spanning_tree
 from cea.technologies.network_layout.substations_location import calc_substation_location
+from cea.utilities.dbf import dbf_to_dataframe
 
 __author__ = "Jimeno A. Fonseca"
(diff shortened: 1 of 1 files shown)


shanshanhsieh:
hey @martin-mosteiro , 
it seems like this function is a duplicate of line 32-46 in `substations_location.py`
have you tested those lines before? that seems to be working for the DC case i tried.
could you please remove the old ones if that is not working? 
i would to keep the option, `buildings-with-demand`, in the config though

martin-mosteiro:
@shanshanhsieh oops... good catch! Seems this was added after @emanuel30 started his simulations. I do think this should have a more intuitive name, so I'll delete my lines and rename the option in the config file.

daren-thomas:
closing pull request, martin want it to happen this way.

PR #2035, opened 2019-05-31
fixes building energy balance plots

gabriel-happle:
- change sign of `I_rad` (line 170 and 171)
- plus some pep8 in the file

Testing by running the energy balance plot and checking if I_rad is < 0 
(previously it was >0)

### DIFF
--- a/cea/plots/demand/energy_balance.py
+++ b/cea/plots/demand/energy_balance.py
@@ -167,7 +167,8 @@ def calc_monthly_energy_balance(data_frame, normalize_value):
     :return:
     """
 
-    # invert sign of I_rad, I_rad was previously negative but it was changed to be consistent with ISO 13790
+    # invert the sign of I_rad. I_rad is a positive term in ISO 13790 and therefore positive in the rest of the code.
+    # I_rad is energy being irradiated from the building to the sky. It is a heat loss in the energy balance.
     data_frame['I_rad_kWh'] = -data_frame['I_rad_kWh']
 
     # calculate losses of heating and cooling system in data frame and adjust signs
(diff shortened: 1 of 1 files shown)


cea/plots/demand/energy_balance.py
```diff
@@ -159,25 +158,25 @@ def calc_graph(analysis_fields, data_frame):
...
     """
     calculates heat flux balance for buildings on hourly basis
 
-    :param data_frame:
+    :param data_frame: demand information of building in pd.DataFrame
+    :param normalize_value: value for normalization of thermal energy fluxes, usually GFA
     :return:
     """
 
+    # invert sign of I_rad, I_rad was previously negative but it was changed to be consistent with ISO 13790
```
daren-thomas: the comment "I_rad was previously negative but it was changed" is hard to understand without a historical context of the code. Is there a better way to state this?

What is I_rad like before changing signs? Why do we change signs?

cea/plots/demand/energy_balance.py
```diff
@@ -159,25 +158,25 @@ def calc_graph(analysis_fields, data_frame):
...
+
     # calculate losses of heating and cooling system in data frame and adjust signs
     data_frame['Qhs_loss_sen_kWh'] = -abs(data_frame['Qhs_em_ls_kWh'] + data_frame['Qhs_dis_ls_kWh'])
     data_frame['Qhs_tot_sen_kWh'] = data_frame['Qhs_sen_sys_kWh'] + abs(data_frame['Qhs_loss_sen_kWh'])
     data_frame['Qcs_loss_sen_kWh'] = -data_frame['Qcs_em_ls_kWh'] - data_frame['Qcs_dis_ls_kWh']
     data_frame['Qcs_tot_sen_kWh'] = data_frame['Qcs_sen_sys_kWh'] - abs(data_frame['Qcs_loss_sen_kWh'])
-    data_frame['Qcs_tot_lat_kWh'] = -data_frame['Qcs_lat_sys_kWh']  # change sign because this variable is now positive in demand
-
-
+    # change sign because this variable is now positive in demand
```
daren-thomas: here again "is now positive in demand" refers to historical context of the code (i'm guessing) and i think that adds a dimension we shouldn't be adding. a comment describing _why_ the value needs to be changed (just for the plot, right?) would be interesting. (also, the totals is just the system stuff?)

cea/plots/demand/energy_balance.py
```diff
@@ -253,13 +253,10 @@ def calc_monthly_energy_balance(data_frame, normalize_value):
...
 def main():
     import cea.config
-    import cea.inputlocator
     import cea.plots.cache
     config = cea.config.Configuration()
-    locator = cea.inputlocator.InputLocator(config.scenario)
-    # cache = cea.plots.cache.PlotCache(config.project)
     cache = cea.plots.cache.NullPlotCache()
-    EnergyBalancePlot(config.project, {'building': locator.get_zone_building_names()[0],
+    EnergyBalancePlot(config.project, {'building': config.plots.building,
```
martin-mosteiro: @gabriel-happle This was actually me, not @daren-thomas (his user was signed in on the computer I made these edits from).

Changing this line allows you to run energy_balance.py with a selected building from the config file. Before, you could either run all plots for a building specified in the config file (even if you didn't need/want the other plots) or you could run the energy_balance plot for whatever the first building in the zone shapefile is. I think this change adds useful functionality.

gabriel-happle:
@daren-thomas and @martin-mosteiro I updated the comments.

PR #2034, opened 2019-05-31
I1932 process schedules for lab

gabriel-happle:
This was supposed to be a small issue.
Many other things were found and fixed.
- Process schedules added to LAB and HOSPITAL in SG (mirroring CH)
- reading of all currently present process schedules in occupancy model (used to be only INDUSTRIAL)
- processing internal loads and other archetype information for relatively new uses of  'LAB', 'MUSEUM', 'LIBRARY', 'UNIVERSITY' (they were just ignored previously)
- creating an exception for unknown occupancies in `occupancy.dbf`

assigning @martin-mosteiro for review and @daren-thomas because of the exception.

PR #2052, opened 2019-06-06
Thermal network simulation config file

martin-mosteiro:
This PR changes a few lines in `thermal_network.py` so that whether the representative week per month method is defined by the `thermal_network` section of the config file, not `thermal_network_optimization`. 

Up until now the variable was called as `config.thermal_network_optimization.use_representative_week_per_month`, so whether the representative week method is used or not was defined not by the thermal-network section of the config file, but by the thermal-network-optimization section (i.e., when running thermal network simulation the optimization configuration overrides the simulation configuration).

### DIFF
--- a/cea/technologies/thermal_network/thermal_network.py
+++ b/cea/technologies/thermal_network/thermal_network.py
@@ -436,7 +436,7 @@ def thermal_network_main(locator, network_type, network_name, file_type, set_dia
     thermal_network.t_target_supply_df = write_substation_temperatures_to_nodes_df(thermal_network.all_nodes_df,
                                                                                    thermal_network.t_target_supply_C)  # (1 x n)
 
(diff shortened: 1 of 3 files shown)


martin-mosteiro:
@shanshanhsieh your latest commit does not pass the tests - but I believe Jenkins wasn't working anyway?
Did it pass the tests locally on your computer?

shanshanhsieh:
yep, cea test ran through on my laptop

martin-mosteiro:
Thanks @shanshanhsieh for the fix in the optimization script. The tests pass locally on my computer. As described in #2034 Jenkins was having some problems so it's being restarted, so I think this PR should be passed pending Jenkins approval.

martin-mosteiro:
Issue #2064 will also be solved by this PR

PR #2055, opened 2019-06-10
1961 project overview improvements

### LINKED ISSUE #1961: Image for scenarios in project overview to make them more identifiable
Along with a short description, an image that represents each scenario in the project overview would make it easier for users to identify and differentiate them.

reyery:
This PR resolves #1961, each scenario in the project overview will now have an image of the area with the building polygons from the `zone.shp` file. The description of each scenario now shows the centroid coordinates of the area (more useful information could be added to the description). In the absence of a `zone` file, no image would be shown and a warning would be shown in the description of the scenario.

Other fixes:
- Description input when creating a new scenario is removed.

- Also made changes to prevent resetting tool parameters to default from changing the active scenario.

- Remove project blueprint.


reyery:
This PR requires a new library `staticmap` which needs to be included in the `dependencies` zip file for the installer

cea/interfaces/dashboard/landing/routes.py
```diff
@@ -46,7 +47,22 @@ def route_project_overview():
...
     # Get the list of scenarios
     scenarios = get_scenarios(project_path)
-    return render_template('project_overview.html', project_name=project_name, scenarios=scenarios)
+
+    # Get scenario descriptions
+    descriptions = {}
+    for scenario in scenarios:
+        descriptions[scenario] = {}
+        scenario_path = os.path.join(project_path, scenario)
+        zone = os.path.join(scenario_path, 'inputs', 'building-geometry', 'zone.shp')
```
daren-thomas: please don't create paths like this. use the `InputLocator`. everywhere. i'm planning to move the whole locator thing away from using paths in scripts, but DataFrames instead - so we can have a `CeaServerInputLocator` class that can `locator.df.zone` (or similar) and it would work even if the project files are not on the current computer. this would be the case, say, in a distributed context.

cea/interfaces/dashboard/landing/routes.py
```diff
@@ -160,6 +176,41 @@ def route_open_project_scenario(scenario):
     return redirect(url_for('inputs_blueprint.route_table_get', db='zone'))
 
 
+@blueprint.route('/get-image/<scenario>')
+def route_get_images(scenario):
+    cea_config = current_app.cea_config
+    project_path = cea_config.project
+    zone_path = os.path.join(project_path, scenario, 'inputs', 'building-geometry', 'zone.shp')
```
daren-thomas: here again - let's use the locator to get hold of these files. instead of assuming they're inside a specific folder structure. that's what the `InputLocator` does: it decouples the storage location (folder structure) from all the scripts and makes it a single point to change, if necessary. also, the next step is to decouply storage _type_ (files vs. mysql or whatever).

reyery:
hmmm... just realized that the image would not update if the user somehow changes the zone file after the preview image is created.

PR #2067, opened 2019-06-12
Updated documentation to include trace-inputlocator

Jack-Hawthorne:
has some inclusions once the glossary PR is finished

### DIFF
--- a/docs/how-trace-inputlocator-works.rst
+++ b/docs/how-trace-inputlocator-works.rst
@@ -1,0 +1,98 @@
+:orphan:
+
+Metadata acquisition
+====================
+
+trace-inputlocator
+------------------
+The CEA contains a tool for the acquisition of metadata specific to each scenario, called via the command line interface
+by `cea trace-inputlocator --scripts <script_name>`. As the name suggests, the trace-inputlocator module records which methods are
+called from `inputlocator.py` during script execution. Having established which script calls which inputlocator method (and subsequently which file),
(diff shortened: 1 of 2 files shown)


daren-thomas:

docs/how-trace-inputlocator-works.rst
```diff
@@ -0,0 +1,98 @@
...
+    - 'created_by' : list of scripts which are responsible for writing the file (empty list if input file)
+    - 'used_by' : list of scripts which read the file (empty list if no dependencies exist)
+    - 'file_path' : the absolute path of the file
+    - 'file_type' : the file extension (e.g. 'xls' or 'dbf')
+    - 'schema' :
+        - variable or sheet name (for xls files)
+            - 'sample_data' : the last non-null entry of the file
+            - 'types_found' : list of python types encountered when reading the file, including None types
+
+Note! The first time metadata is generated, the order of the scripts is important as the 'created_by' list is updated each time a script is run. For example,
```
daren-thomas: I'm surprised by this. Can we meet tomorrow to discuss this?

Jack-Hawthorne: what is surprising about it? if the data is already there but the trace-inputlocator has not been run for a dependency, then how is it supposed to know which script created the data?

daren-thomas: well... the Note! sounds like it's a problem, but the explanation suggests it just works... and would just update the `created_by` when it get's there?

PR #2080, opened 2019-06-18
Bugfix: Fix terrain boundary caused by rounding error

### LINKED ISSUE #2014: Error when running the Solar Radiation component
This case study was not able to run the solar radiation. This error is indicated for one of the buildings (CEA1004):
_ValueError: ERROR: this usually happens when buildings do not overlap with the terrain, try to check if this is the case_
I ran the terrain helper before and didn't identify nothing special about the building geometry that can be causing this.

Case study file is attached here (expires in one week): https://we.tl/t-DQOCzi1jdc

reyery:
This PR fixes #2014. Made sure that the difference between the boundary `max` and boundary `min` divided by `grid_size` would generate a whole number, so casting it into an `int` would not lose any precision.

### DIFF
--- a/cea/datamanagement/terrain_helper.py
+++ b/cea/datamanagement/terrain_helper.py
@@ -80,6 +80,14 @@ def terrain_elevation_extractor(locator, config):
     x_max = bounding_box_district_file[2] + extra_border
     y_max = bounding_box_district_file[3] + extra_border
 
+    # make sure output is a whole number when min-max is divided by grid size
+    x_extra = grid_size - ((x_max - x_min) % grid_size)/2
+    y_extra = grid_size - ((y_max - y_min) % grid_size)/2
+    x_min -= x_extra
+    y_min -= y_extra
+    x_max += x_extra
+    y_max += y_extra
+
     ##TODO: get the elevation from satellite data. Open-elevation was working, but the project is dying.
     # if elevation is None:
     #     print('extracting elevation from satellite data, this needs connection to the internet')


cea/datamanagement/terrain_helper.py
```diff
@@ -80,6 +80,14 @@ def terrain_elevation_extractor(locator, config):
     x_max = bounding_box_district_file[2] + extra_border
     y_max = bounding_box_district_file[3] + extra_border
 
+    # make sure output is a whole number when min-max is divided by grid size
```
daren-thomas: i'm curious... can't you just use `math.ceil(x_max)` and `math.floor(x_min)` to get to the next integer?

reyery: Oh, what I meant was that the difference of `max` and `min` i.e. `max - min` would be a multiple of `grid_size`, so that the result would be a whole number when divided by `grid_size`.

reyery: it's because of these lines, which truncates the decimals when cast to `int`. Some times it causes the terrain boundary to have one less row/column then it's supposed to.

https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/e59f12da4015c7cb31b8be47dcbb5ed45d6cc03b/cea/datamanagement/terrain_helper.py#L102-L103

Cound just do `math.ceil()`, but the terrain boundary would be lopsided i.e. right boundary more than the left, so instead I added the difference on both ends.

daren-thomas: thank you for explaining!

Issue #2082, opened 2019-06-19
Add network plots to Dashboard


daren-thomas:
Inside `cea.plots.thermal_networks.main.plots_main`, there are three plots that I have not added as part of #1984, as they require some re-thinking of how to show that data. Currently, a png image is produced. For the Dashboard a more interactive and visually interesting plot would be similar to how the `cea/plots/supply_system/map_chart.py` works for the supply system plots. The supply system plots are currently non-functional (see #2081)

The three plots in question:

- DC__network_layout.png
- DC__pressure_loss_networkaggregated.png
- DC__pressure_loss_networkpeak.png

daren-thomas:
Add these maps to the "map" part of map-style dashboard.

- how? does user select the plot to show?
- maybe a dropdown or something.

JIMENOFONSECA:
@daren-thomas so these are ready to add they should be part of the new supply maps #1578  (which show 3D information)

gabriel-happle:
@reyery and @daren-thomas what is the status of this issue?

daren-thomas:
@gabriel-happle There is a PR which addresses it. It hasn't been merged yet.

PR #2083, opened 2019-06-19
1984 add thermal networks plots to the dashboard

daren-thomas:
This PR adds thermal-networks plots to the dashboard with the exception of the network plots (the png ones) as mentioned in #2082 - these are blocked by supply-system plots not working to test a more "dashboardy" way to produce these plots. also, they will require quite some work!

To test:

- add all the "Thermal networks" plots to a Dashboard
- try changing the parameters
- try opening a scenario that is missing thermal network data


daren-thomas:
@JIMENOFONSECA as per discussion on slack, i'm merging this now with the expectation that you check up on the plots later this week.

PR #2093, opened 2019-06-26
1978 create an uninstaller for the cea

daren-thomas:
This PR updates the CEA installer to also uninstall - a shortcut is added to the menu and the uninstaller tries to uninstall all the various CEA files (including the `cea.config` - the assumption being, if you're uninstalling, you're _done_ with the CEA!).

To test this:

- create a backup of your `cea.config` (it's in your profile folder)
- install from here: https://www.dropbox.com/s/mj7t2prgon2zotr/Setup_CityEnergyAnalyst_2.17a1.exe?dl=0
- uninstall
- check:
  - `cea.config` file should be gone
  - `Documents/CityEnergyAnalsyst` folder should be gone
  - Desktop shortcuts (if created) should be gone
  - start menu folder "City Energy Analyst" should be gone


JIMENOFONSECA:
It does not work as planned. Kindly test thoroughly and upload again.

Neither the folder nor the shortcuts are erased after uninstalling CEA

image.

image

JIMENOFONSECA:
@daren-thomas On the other hand, CEA does not work if installed in Program files (even with Admin privileges). that kind of sucks...

daren-thomas:
> 
> 
> @daren-thomas On the other hand, CEA does not work if installed in Program files (even with Admin privileges). that kind of sucks...

What do you mean "does not work"? That's a very vague description...

daren-thomas:
Can you please check the contents of "CityEnergyAnalyst"? I suspect it's not removed because some files were still in use (PyCharm maybe locking some files or - as in my case - the CEA Console was open).

I'm doing some more testing, but the default case should work with the /rebootok option to the deleting. And the rest should be separate issues that can then be prioritized - we _do_ have bigger fish to fry and inferring from "that kind of sucks..." i see you too spent a lot of time testing this. it's frustrating.

daren-thomas:
Deletion of desktop shortcuts _does_ work. At least on my computer.

daren-thomas:
Installing to "C:\Program Files" produces the following error message:

image

daren-thomas:
Some thoughts: Due to the way we're installing, it's probably not a good idea to install in ProgramFiles at all anyway. Especially since we're adding (in the Developer version) a whole git repository there. So. I think we should just create a Known Issue about that and move on. Issue #2113 describes just that (installing seems to progress if the installer is started with "Run as Administrator", but I don't feel comfortable recommending this route).

In our original design goals for the Installer, installing to "Documents" was the default choice. I still think that's the best idea so far.

daren-thomas:
@JIMENOFONSECA is it at all possible to merge this anyway now? And create issues for remaining problems? This PR also includes some minor improvements for installing to folders that contain spaces and I'd like to have it as part of release 2.17.

daren-thomas:
@JIMENOFONSECA especially since you moved it to the sprint _after_ the current one? that's in four weeks time. I'd rather not have PR's waiting for so long.

PR #2133, opened 2019-07-03
FEATURE: New 500 status error page

### LINKED ISSUE #2132: More informative `500` status error message for dashboard
Currently in the dashboard, an error in the program will cause flask to respond with a generic `500` message i.e. `Internal Server Error`. Some users have been using this error message to report new issues however this does not provide us with any information regarding the cause of the error.

It will be good if the `500` status page could provide information such as the stack trace so that users are able to craft more informative issues and allow us to squash bugs more efficiently.

reyery:
This PR resolves #2132.

Features
---
- New `500` status error page that shows the user the stack trace of the error to help with more informative error/issue reporting.

### DIFF
--- a/cea/interfaces/dashboard/base/routes.py
+++ b/cea/interfaces/dashboard/base/routes.py
@@ -60,6 +60,8 @@ def not_found_error(error):
     return render_template('errors/page_404.html'), 404
 
 
-@blueprint.errorhandler(500)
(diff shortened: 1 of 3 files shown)


reyery:
Testing
---
- Try to create an error
Edit any of routes in the dashboard to throw an error e.g. `raise IOError` and go to that route in the browser.
It should show the stack trace of the error.

PR #2158, opened 2019-07-13
edited schedules documentation

Jack-Hawthorne:
Fixed some issues with the schedules documentation, namely the error in formula for T01.


daren-thomas:

PR #2159, opened 2019-07-15
added more uses to reference-case-cooling

daren-thomas:
This PR updates the file `cea/examples/reference-case-cooling.zip` to have the following uses:

- B001: MULTI_RES, SINGLE_RES
- B002: OFFICE, RETAIL
- B003: RESTAURANT, SCHOOL
- B004: FOODSTORE, GYM
- B005: HOSPITAL, HOTEL
- B006: INDUSTRIAL, LIBRARY
- B007: COOLROOM, SWIMMING
- B008: PARKING, SERVERROOM
- B009: GYM, SCHOOL
- B010: MULTI_RES, OFFICE

To test, run `cea extract-reference-case --case cooling` and then set up data-helper, radiation-daysim and demand to run it with region SG and weather Singapore.


gabriel-happle:
The case study runs.
I suggest we merge this and at the same time we need to check some potential issues:

- How are construction and other archetypical properties assigned in buildings with exactly 50:50 occupancy use mix?
- Energy use of B004 (Foodstore & Gym) seems very low. Why?
- Energy use of B007 (Coolroom & Swimming) seems very low. Why?

newplot

daren-thomas:
@gabriel-happle would you mind adding your points as separate issues? then we can merge.

gabriel-happle:
@daren-thomas, they already are. see the links behind the bullet points.

daren-thomas:
@gabriel-happle alrighty then. care to merge?

Issue #2150, opened 2019-07-09
Internal server error when running Plots basic and Dashboard plots


DespaxJulien:
Hi,

when I try to run the `Plots basic` tool, the dashboard turn into an error page indicating 

> Internal server error

 (see picture attached).

I get the same problem when trying to add plots on the dashboard.

Steps I followed:

- Running `Solar radiation (Daysim engine)` (which worked fine: I got the CSV files in my scenario folder)

- Running `Demand` (which worked fine: I got the CSV files in my scenario folder)

- Running `Plots basic`. Then I got the error

Does this problem comes from my computer, or is it something you've experienced/know about?

Thanks for your help!

Here is the error landing page:

![Capture](https://user-images.githubusercontent.com/51714760/60904111-c533f980-a272-11e9-895d-9cf9229c270b.PNG)

reyery:
Hey @DespaxJulien thanks for reporting this issue. Could you paste the error from the Dashboard console?

If you need to see the plots, in the meantime you could try out the Dashboard feature.

<img width="297" alt="Capture1" src="https://user-images.githubusercontent.com/34395415/60936123-9cf1dc80-a2fe-11e9-8634-0f4d87f7ed9c.PNG">

DespaxJulien:
@reyery the issue is that there is no dashboard console when I run `Plots Basic`.

Like, when I click on `Plots basic`, the `Internal Server Error` displays instantly, without any console.

I tried to visualize the plots in the Dashboard feature, as you suggested, and I get this:

- For most of the plots (like `Load curve`, `Energy demand`, `Energy supply`...), I get the same `Internal Server Error` as soon as I click on the `Add plot` button.

- For some other plots (like `Heating reset schedule` or `Comfort chart`) , I get another error (see picture attached) displayed on the dashboard interface (but no console starts).

- Finally, the `Energy balance` works, but it is the only plot that works (in the `Energy Demand` category).

I hope my explanation is clear. I could send my files if neccessary.

![Capture1](https://user-images.githubusercontent.com/51714760/60952859-5a2f0500-a2fc-11e9-85a0-1eaa8deec7fd.PNG)

daren-thomas:
@DespaxJulien The Dashboard console @reyery means is the minimized console that is opened when you run the "CEA Dashboard" shortcut. It shows up in the taskbar:

[...]

PR #2160, opened 2019-07-16
2150 Feature: cea rename-building

daren-thomas:
This PR tries to address #2150 (internal server error for plots without b001) but ultimately I couldn't reproduce the error. Instead, this PR contains:

- a bugfix for PlantNodeParameter (for plotting scenarios that don't have a thermal network defined)
- a new script `cea rename-building` that allows the user to rename a building in all the input files. `data-helper` and all following scripts in the workflow need to be re-run after this script to avoid errors.


daren-thomas:
@reyery would you mind testing this, please? it's similar to work you did with deleting a building, so I guess you're close to the source on this one.

cea/utilities/rename_building.py
```diff
@@ -0,0 +1,69 @@
+"""
+A simple CEA script that renames a building in the input files - NOTE: you'll have to re-run the simulation and
+analysis scripts to get the changes as only the files defined in ``inputs.yml`` (the files you see in the CEA Dashboard
+input editor) are changed.
+
+This is the script behind ``cea rename-building --from <building> --to <building>``
```
reyery: Is this supposed to be ``cea rename-building --old <building> --new <building>`` ?

reyery:
I am getting this error, probably because the the default config for `old` is empty? I'm not sure.

```
City Energy Analyst version 2.18.1
Running `cea rename-building` with the following parameters:
- general:scenario = C:\tesss\eweq
  (default: C:\reference-case-open\baseline)
- rename-building:old = B1012
Process Process-6:
  File "C:\Users\Admin\Documents\CityEnergyAnalyst\Dependencies\Python\lib\multiprocessing\process.py", line 267, in _bootstrap
    self.run()
  File "C:\Users\Admin\Documents\CityEnergyAnalyst\Dependencies\Python\lib\multiprocessing\process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\Admin\Documents\CityEnergyAnalyst\CityEnergyAnalyst\cea\interfaces\dashboard\tools\worker.py", line 19, in run_script
    script(**kwargs)
  File "C:\Users\Admin\Documents\CityEnergyAnalyst\CityEnergyAnalyst\cea\api.py", line 26, in script_runner
    cea_script.print_script_configuration(config)
  File "C:\Users\Admin\Documents\CityEnergyAnalyst\CityEnergyAnalyst\cea\scripts.py", line 40, in print_script_configuration
    print("  (default: %s)" % default_config.get(parameter.fqname))
  File "C:\Users\Admin\Documents\CityEnergyAnalyst\CityEnergyAnalyst\cea\config.py", line 178, in get
    return self.get_parameter(fqname).get()
  File "C:\Users\Admin\Documents\CityEnergyAnalyst\CityEnergyAnalyst\cea\config.py", line 333, in get
    return self.decode(encoded_value)
  File "C:\Users\Admin\Documents\CityEnergyAnalyst\CityEnergyAnalyst\cea\config.py", line 633, in decode
    assert self._choices, 'No choices for {fqname} to decode {value}'.format(fqname=self.fqname, value=value)
AssertionError: No choices for rename-building:old to decode 
```

cea/utilities/rename_building.py
daren-thomas: you're right! fixed!

daren-thomas:
@reyery I don't know how to reproduce the AssertionError. What did you run? It looks like there are no buildings in your scenario?

reyery:
@daren-thomas how is your default for `rename-building:old` = `B01` when the one in default config is empty?

image

I thought that the `SingleBuildingParameter` uses the `ChoiceParameter` `decode` method:
https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/cbe88dedf6dfb6aef73590042dd443d965a74ccd/cea/config.py#L646-L651

Since running the script from the cli/dashboard invokes `api`, which decodes and prints out default parameters, I don't see how it is setting it to `B01`. Is there something I am missing?

daren-thomas:
@reyery I just tried this again - on a different computer that had run the heating-case workflow. And it just works. Are you sure your setup is good?

reyery:
@daren-thomas I tried this on a fresh windows VM with the 2.18 installer and it seems to work. I guess there might be something wrong with my setup.

reyery:
@daren-thomas ok.. after ripping my hair out I finally found the reason why I'm getting this error. Turns out that the building geometry files for my `reference-case-open` folder located in `C:\reference-case-open\baseline` as indicated in `default.config` are missing. 

So since default `_choices` is taken from the `zone` file, it will return an empty error.
https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/9c5eca3e4a14aee2b29a2c2a62721987b2390f9e/cea/config.py#L760-L763

Running `cea extract-reference-case` solved it.

This means that it is potentially dependent on the existence of the `reference-case-open` folder (and its contents) for it to work.

daren-thomas:
ah! ok. that explains a lot. I guess, if no zone geometry can be found, then an error is totally correct! I updated the error message:

```
C:\Users\darthoma\Documents\CityEnergyAnalyst
λ cea-config rename-building
City Energy Analyst version 2.18.1
Configuring `cea rename-building` with the following parameters:
- general:scenario = c:\Users\darthoma\Documents\CityEnergyAnalyst\projects\reference-case-open\baseline
  (default: C:\reference-case-open\baseline)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\Scripts\cea-config-script.py", line 11, in <module>
    load_entry_point('cityenergyanalyst', 'console_scripts', 'cea-config')()
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\interfaces\cli\cea_config.py", line 45, in main
    cea_script.print_script_configuration(config, verb='Configuring')
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\scripts.py", line 38, in print_script_configuration
    parameter_value = parameter.get()
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\config.py", line 333, in get
    return self.decode(encoded_value)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\config.py", line 647, in decode
    if str(value) in self._choices:
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\config.py", line 765, in _choices
    raise cea.ConfigError("Either no buildings in zone or no zone geometry found.")
cea.ConfigError: Either no buildings in zone or no zone geometry found.
```

reyery:
So this means that the user must have the `reference-case-open` folder for it to work? What if the user deletes it?

daren-thomas:
@reyery no, it means the user needs to be working on a scenario that has zone geometry. which is a totally ok requirement, since renaming a (non-existant) building in a scenario without zone geometry doesn't really mean much anyway.

reyery:
@daren-thomas the thing is that the tool does not run (in cli and dashboard) if the zone geometry file does not exist in the `reference-case-open` folder even if the user is working on a scenario that has zone geometry.

daren-thomas:
wait. you're sure? how do i re-create this please? and which folder is it referencing? i can hardly imagine this is the case? and if so, phew. thanks for not merging!

reyery:
Deleting the `reference-case-open` will cause this error, since running from cli/dashboard will cause the script to try to decode the buildings of the default scenario found in `default.config`.
https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/e673970bd68737eb1b0ca01cf5b8d18c40b1f960/cea/default.config#L1-L8

daren-thomas:
@reyery ok, now I finally understand. Wow. That was a sneaky little bug - it was only showing up for the printing of the default value. I've changed that now to be more parameter-specific: Parameters now use their own decoding strategies to show the default value. It should work now. Sorry for being so dense about this.

reyery:
Nice! It works now even without `reference-case-open`. Merging

PR #2161, opened 2019-07-16
don't create the inputs folder when opening a project + scenario

daren-thomas:
How to test:

- create a new project folder, with an empty scenario folder inside
- start CEA Dashboard
- choose "Open Project" and select the project folder
- in the project overview, choose the empty scenario folder
- you should see an image like the one below
- check in Windows Explorer to confirm the scenario folder is still empty

image

NEXT:

- in project overview, choose "Create new scenario"
- try out all the possibilities
  - "Generate new input files" should work
  - "Copy input folder from other scenarios in the project." should work
  - "Import input files" should work


daren-thomas:
@reyery would you mind testing this and merging please?

PR #2162, opened 2019-07-17
FEATURE: New Map Dashboard layout

### LINKED ISSUE #2140: Redesigning the dashboard
Currently our dashboard layout is vertically stacked with one plot on top of another. This would mean that users would have to scroll down to see all the plots that they have added.

We could improve on this layout by making the plots all in one page instead so that the user can see all the plots in one glance without having to scroll down the page (the whole point of a dashboard).

One of the layouts could be like this (example from [Dash](https://dash-gallery.plotly.host/Portal/))
Screenshot_2019-07-05 Dash

We can have one that will show visualizations on a map, accompanied by supporting graphs/plots at the side.

### LINKED ISSUE #2156: Tables should be a link in the plots (do not visualize them)
So plots have some tables. However, it would be better to include a button on the top right of every plot that takes me to this table (like a pop-up or a new screen)

It is rather difficult to use them like they are.

@daren-thomas would you mind creating a class and a structure to handle this event? thanks!

reyery:
This PR resolves #2140 and resolves #2156.

New Features
---
- New map layout for dashboards

Screenshot_2019-07-17 City Energy Analyst Test

- Allow user to choose between 2 different dashboard layouts when creating new dashboard 
i.e. `Row` (default) and `Map` layout
- Duplicate dashboards
- Title of plots show the scenario that it is based on
- Button to open the table of the plot in a popup


reyery:
Testing
---
- Create new dashboard and select the `Map` layout
  - Dashboard with one map widget and four empty plot widgets created
- Add new plot in the layout
- Click `change` icon in the plot title to change the plot
  - Plot should change
- Resize browser window of `Map` layout
  - Widgets and plots should be reactive and will resize and rearrange accordingly
- Click `table` icon in the plot title
  - Table of plot will appear in a popup. If table does not exist, there should be a message that says so
- Click `setting` icon in the plot title and change the scenario of the plot
  - Plot should be correct and title should also change to reflect the scenario name
- Remove some input files required by one of the plots and refresh
  - There should not be a `table` icon in the plot title

cea/interfaces/dashboard/plots/routes.py
```diff
@@ -65,22 +71,28 @@ def route_new_dashboard():
...
     plot_cache = current_app.plot_cache
-    dashboard_index = cea.plots.new_dashboard(cea_config, plot_cache, request.form.get('name'), request.form.get('description'))
+    dashboard_index = cea.plots.new_dashboard(cea_config, plot_cache, request.form.get('name'),
+                                              request.form.get('description'), request.form.get('layout'))
     return redirect(url_for('plots_blueprint.route_dashboard', dashboard_index=dashboard_index))
 
 
-@blueprint.route('/dashboard/<int:dashboard_index>/<func>')
+@blueprint.route('/dashboard/<func>/<int:dashboard_index>')
 def route_manage_dashboards_function(dashboard_index, func):
```
daren-thomas: why is this one function and not 3 separate functions?

- `route_delete_dashboard`
- `route_rename_dashboard`
- `route_duplicate_dashboard`

cea/interfaces/dashboard/plots/static/js/dashboard-plots.js
```diff
@@ -7,12 +7,12 @@ $(document).ready(function() {
     console.log("assigned categories!");
     console.log(categories);
 
-    $("#cea-plot-category").on("change", function(e){
+    $("[id|=cea-plot-category]").on("change", function(e){
```
daren-thomas: hey, cool! new syntax! i've never seen this before, what's it called?

cea/interfaces/dashboard/plots/static/js/dashboard-plots.js
```diff
@@ -70,20 +74,35 @@ function load_all_plots() {
         let x_table_id = "#x_table-" + dashboard_index + "-" + plot_index;
 
         $.get("../div/" + dashboard_index + "/" + plot_index, function(data){
```
daren-thomas: yeah, I'm sorry about that line: i _hate_ concatenating strings. that's why i've started using `"".format()` in python. at the time i wrote that, i didn't know about template strings yet - only now reading up on all the new features javascript has. it's come a long way since 1999!

cea/interfaces/dashboard/plots/static/js/dashboard-plots.js
```diff
@@ -70,20 +74,35 @@ function load_all_plots() {
...
-                $(x_content_id).children().replaceWith(data);
+            $(x_content_id).children().replaceWith(data);
+            $('#table-btn-'+plot_index).show();
         }).fail(function(data) {
-            $(x_content_id).children().replaceWith("ERROR: " + $(data.responseText).filter("p").text());
-            console.log("error creating plot:");
-            console.log(data);
+            // Server error
```
daren-thomas: nice! :)

cea/interfaces/dashboard/plots/static/js/dashboard-plots.js
```diff
@@ -70,20 +74,35 @@ function load_all_plots() {
...
         });
 
-        $.get("../table/" + dashboard_index + "/" + plot_index, function(data){
-                $(x_table_id).children().replaceWith(data);
-        }).fail(function(data) {
-            $(x_table_id).children().replaceWith("");
-            console.log("error creating plot:");
-            console.log(data);
-        });
+        // $.get("../table/" + dashboard_index + "/" + plot_index, function(data){
```
daren-thomas: i think we can just delete old code - no need to keep it around!

cea/interfaces/dashboard/plots/templates/manage.html
```diff
@@ -69,7 +73,7 @@ <h2 class="pull-left">Manage Dashboards</h2>
   {{ super() }}
   <script>
     function openSettings(dashboard_index, func) {
-        $.get('./' + dashboard_index + '/' + func, {}, function(html) {
+        $.get('./' + func + '/' + dashboard_index, {}, function(html) {
```
daren-thomas: yeah, i still think this needs to be refactored (see above) -  https://refactoring.com/catalog/removeFlagArgument.html

cea/plots/__init__.py
```diff
@@ -63,20 +63,34 @@ def dashboard_yml_path(config):
     return dashboard_yml
 
 
-def default_dashboard(config, cache, name='Default Dashboard', description=''):
+def default_dashboard(config, cache, name='Default Dashboard', description='', layout='default'):
```
daren-thomas: i'm wondering, do we really need 'default' as a layout type? can we not just have 'map' and 'rows'? and just set that in `default_dashboard` as the default parameter value? (then you don't need extra logic to handle the 'default' layout)

cea/plots/__init__.py
```diff
@@ -99,7 +113,11 @@ def __init__(self, config, dashboard_dict, cache):
         try:
             self.description = dashboard_dict['description']
         except KeyError:
-            self.description = None
+            self.description = ''
+        try:
+            self.layout = dashboard_dict['layout']
+        except KeyError:
+            self.layout = 'default'
```
daren-thomas: again. here we're introducing a third value for `self.layout`... which i guess maps to rows?

daren-thomas:
I _love_ the scenario-name in the plot title feature! so simple. so effective.

daren-thomas:
when duplicating a dashboard, the new description should not be "copy of {original_dashboard.name}", it should be original_dashboard.description.

daren-thomas:
Why don't we show the plot table in a modal dialog? i think that would look better, what do you think? where there any special reasons for opening a new window?

cea/interfaces/dashboard/plots/routes.py
reyery: hmm.. I probably thought it was a good idea and didn't want to keep copying the same few lines of code.

cea/interfaces/dashboard/plots/static/js/dashboard-plots.js
reyery: Oh its a jQuery selector for element attributes that matches element attribute values starting with the string followed by a hyphen. Was using something that required the same logic.

cea/interfaces/dashboard/plots/static/js/dashboard-plots.js
reyery: Will help to change the string concatenations to templates (though Internet Explorer is the only browser that doesn't support it, but shouldn't be an issue)

Yeah and there's so many javascript frameworks now too! Actually I have been trying to see if I can convert the front end to use **Electron** and **React** during my free time. Pretty cool stuff.

cea/interfaces/dashboard/plots/static/js/dashboard-plots.js
reyery: Oops, I commented it out for testing and forgot to remove it entirely

cea/plots/__init__.py
reyery: yeah makes sense. Because I originally wanted to name the original layout as 'default' but then realized that it did not mean anything / does not describe the layout.

cea/interfaces/dashboard/plots/static/js/dashboard-plots.js
daren-thomas: yeah, I've been looking into electron myself lately and it _would_ solve the IE issue... but instead of React, I was looking into Vue.js - it seems React is to Vue what Django is to Flask?

cea/plots/__init__.py
daren-thomas: good thinking!

cea/interfaces/dashboard/plots/static/js/dashboard-plots.js
```diff
@@ -109,7 +109,8 @@ function open_table(element) {
     let height = 500;
     let width = 900;
     let table = window.open(`../table/${element.dataset.dashboardIndex}/${element.dataset.plotIndex}`,
-        'popUpWindow', `height=${height},width=${width},location=no,menubar=no,status=no,titlebar=no,resizable`);
+        `table-${element.dataset.dashboardIndex}-${element.dataset.plotIndex}`,
+        `height=${height},width=${width},location=no,menubar=no,status=no,titlebar=no,resizable`);
     table.onload = function() {
```
daren-thomas: Is there a reason we're not doing this in a modal dialog? I think the UI would be better, no?

reyery: the modal kinda blocks the plot so it would be difficult if the user wants to refer to it.

daren-thomas:
OK. duh. my dashboard was using the "default" layout. And we're just accepting any string and mapping that to a template. I'll add some sanitizing and check again.

daren-thomas:
So. Fixed that small bug, merging: this is good work. I did add a bunch of issues (#2170, #2171, #2172) that we can use to tidy up more, but they don't break the CEA, so we can merge! :)

Issue #2195, opened 2019-07-23
BUG: vectorized get_heating_system_set_point, get_heating_system_set_point fail if setpoint temperatures are integers


gabriel-happle:
`np.vectorized(get_heating_system_set_point)`, and `np.vectorized(get_heating_system_set_point)` can fail if setpoint temperatures are integers

This is bug reported by @JIMENOFONSECA in a case study in Zurich.

I am relatively confident, that this is related to the behavior of `numpy.vectorize` and the fact that `numpy.nan` is defined as `float`
https://docs.scipy.org/doc/numpy/reference/constants.html?highlight=numpy%20nan#numpy.nan

```The data type of the output of vectorized is determined by calling the function with the first element of the input. This can be avoided by specifying the otypes argument.```
https://docs.scipy.org/doc/numpy/reference/generated/numpy.vectorize.html

To address this properly, the input data-type of the `dbf` and `excel` files should be ensured.
Maybe @daren-thomas and @reyery can help with this.

To fix this bug here, I will implement the `otypes` argument.

PR #2196, opened 2019-07-23
I2195 fix vectorized setpoint

gabriel-happle:
This fixes bug #2195 . 

How to test:
- run the demand of @JIMENOFONSECA's case study in master with region CH and weather Zug, it should fail with error: `ValueError: cannot convert float NaN to integer`
- run the demand of the same case study in this branch: it should pass

Let me know if you need the case study. I can send it to you.


daren-thomas:
confirmed behavior as in description for testing. this works. thank you :)

PR #2198, opened 2019-07-23
macOS installation guide

martin-mosteiro:
I wrote an installation guide for CEA on macOS. Here it is.


daren-thomas:
ok. I'll have a look at this soon.

docs/installation-guide-mac.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Installation guide for macOS
+============================
+
+Follow these instructions to install the CityEnergyAnalyst (CEA) on a macOS system (tested with Mojave 10.14.5) from the source
+
+Prerequisites
+~~~~~~~~~~~~~
+
+* Download and install Git (64-bit).
```
daren-thomas: actually, just typing `git` in the `terminal` app start a prompt "Would you like to install the tools now"?  for developer tools... that git should be enough!

docs/installation-guide-mac.rst
```diff
@@ -0,0 +1,62 @@
...
+* Download and install Github Desktop (64-bit).
```
daren-thomas: This isn't strictly a pre-requisite. It's just nice-to-have if you're a beginner with git... for documenting the Installation, actual git commands are easier... There are also alternatives (`fork`, `gitk`, `gitkraken` etc.)

docs/installation-guide-mac.rst
```diff
@@ -0,0 +1,62 @@
...
+* Download and install Miniconda(64-bit) for Python 2.7.
```
daren-thomas: the .pkg-installer is the easiest

docs/installation-guide-mac.rst
```diff
@@ -0,0 +1,62 @@
...
+* Download and install CMake.
```
daren-thomas: tested with ``cmake-3.15.0-Darwin-x86_64.dmg`` but newer should work too

daren-thomas: btw: for `cmake` to be available in the terminal, you need to add `export PATH="/Applications/CMake.app/Contents/bin":"$PATH"` to `.bash_profile`.

daren-thomas: or... just type it before the cmake call (then it's not permanent... i like this better!)

docs/installation-guide-mac.rst
```diff
@@ -0,0 +1,62 @@
...
+* Install DAYSIM version >= 4.0: Open Terminal and run the following commands
+
+  - ``git clone https://github.com/MITSustainableDesignLab/Daysim.git``
+  - ``mkdir build && cd build && cmake -DBUILD_HEADLESS=on -DCMAKE_INSTALL_PREFIX=$HOME/Daysim ../Daysim && make && make install``
```
daren-thomas: I get errors in this step...

docs/installation-guide-mac.rst
```diff
@@ -0,0 +1,62 @@
...
+  - ``cp ./CMakeLists.txt /Daysim/CMakeLists.txt``
```
daren-thomas: this can never work

docs/installation-guide-mac.rst
martin-mosteiro: Can't say for sure, but that did not appear to work on my terminal until I manually installed. I might be wrong.

docs/installation-guide-mac.rst
martin-mosteiro: In that case, that step should be deleted from all tutorials ([e.g.](https://city-energy-analyst.readthedocs.io/en/latest/installation-on-windows-manual.html))

docs/installation-guide-mac.rst
martin-mosteiro: Not sure what that means.

docs/installation-guide-mac.rst
martin-mosteiro: `mkdir build && cd build && cmake -DBUILD_HEADLESS=on -DCMAKE_INSTALL_PREFIX=$HOME/Daysim ../Daysim && make && make install`
This step appears to have worked for me...? 🤷‍♀

docs/installation-guide-mac.rst
martin-mosteiro: Again, appears to have worked for me... although I might have gotten an error message when doing that 🤷‍♀

docs/installation-guide-mac.rst
daren-thomas: yeah. I know what you mean. hm.

docs/installation-guide-mac.rst
daren-thomas: there were two installers for miniconda. a bash installer and a .pkg installer. the bash installer needs to be installed from the terminal. which one did you use? (following my Docker example, probably the bash one? with curl to get it?

docs/installation-guide-mac.rst
daren-thomas: ok. it's just, there _is_ no `./CMakeLists.txt`. I don't mean to bash your guide. I was just taking notes while following it and gave up here with the error message and I needed to go home. I will have another look at this soon and see if I can work out some stuff. Maybe we can even write it as a shell script to run. That would make things so much easier!

daren-thomas:
OK. I've gone through a bunch of steps and can now run `cea test` (though it does throw some small errors because the numbers don't _quite_ match up - results are off by roughly a factor of 0.001) and also `cea dashboard` starts up.

There is still the issue of installing DAYSIM. First, I believe this should go _after_ the CEA installation and second, I think we could provide a shell script to do this.

daren-thomas:
Actually, let's pre-compile the DAYSIM binaries and just ship them. Tell the user to copy them to /usr/bin or something.

daren-thomas:
please don't merge yet: i'm still working on this. the main obstacle right now is getting DAYSIM to work.

daren-thomas:
we decided to clean up a version that explicitly does not include DAYSIM and merge.

docs/installation-guide-mac.rst
```diff
@@ -0,0 +1,83 @@
...
+  * (this is not strictly necessary, but we find it easier for novice git users)
+
+* Download and install `Miniconda`_ (64-bit) for Python 2.7.
+
+  * (the Mac OS X 64-bit `.pkg installer`_ is probably the easiest way to install Miniconda on your Mac)
+
+* Download and install `CMake`_.
+
+  * (CMake is required for compiling DAYSIM)
+  * (this guide assumes you installed CMake to ``/Applications/CMake`` and have it in your PATH)
```
martin-mosteiro: What do you mean by this? Is CMake a directory or an application? By default, CMake gets installed to ``/Applications/CMake.app``. Also, how is CMake added to the ``PATH``?

martin-mosteiro: Also, how is CMake added to the ``PATH``?

martin-mosteiro:
Tested locally on my Mac. Daysim indeed does not work on Mac as currently installed in my computer. Unfortunately I do not have the skills to figure out what the issue is. However the radiation script is necessary for anything CEA does, so unless we get Daysim to work I don't think it makes any sense to publish this installation guide.

martin-mosteiro:
I also found that pyproj needs to be upgraded in order to be able to run the dodo. A simple `pip install pyproj --upgrade` did the trick, but should probably be part of the CEA installation itself? @daren-thomas

daren-thomas:
@martin-mosteiro what is our final verdict on this? is CEA on mac feasible? or do we scrap this PR?

martin-mosteiro:
I've only actually tried five features on Mac:
1 - radiation: does _not_ work
2 - data helper: works
3 - demand: works
4 - unit testing: does _not_ seem to work

So users would need access to a Windows or Ubuntu computer to run the radiation, and developers would likewise need access to a Windows or Ubuntu computer for unit testing. This means neither users nor developers can really use CEA without access to another computer (or Parallels).

I think this makes the installation guide basically useless.

@daren-thomas *iff* there's any chance we can solve the Daysim issue on Mac, then it would be very useful to me as a user, and it would make the tool viable to new users on Mac. If this is not feasible, then I don't think it makes sense to publish this guide.

daren-thomas:
@martin-mosteiro I've added a disclaimer that the installation doesn't totally work and isn't supported - I'm hoping someone will pick it up and continue to fix it.

@JIMENOFONSECA: I don't see this as a priority to continue on, but I think we can merge for now - we have a big warning that this doesn't work. And adventurous souls can see how much they get up and running... (development of most modules should be possible, but running a whole workflow including radiation-daysim is currently not supported)

PR #2199, opened 2019-07-23
Heating case workflow

daren-thomas:
This PR adds a new script `cea workflow` (currently only CLI interface supported) that allows running a "workflow.yml" file. The cooling-case-workflow script is removed and the functionality is merged into the workflow script. To test, in the CEA Console, do:

- `cea workflow --workflow heating-case --resume on`
- `cea workflow --workflow cooling-case --resume on`

These scripts will fail after a while, but show exactly how far our workflows currently go.

An example (first few steps) of a workflow.yml file: 

```
- config: default
  "general:multiprocessing": on
  "general:project": "{general:project}/../reference-case-open"
  "radiation-daysim:daysim-bin-directory": ${RAYPATH}
- script: extract-reference-case
  parameters:
    destination: "{general:project}/.."
    case: open
- config: .
  "general:scenario-name": baseline
- script: data-helper
- script: radiation-daysim
- script: demand
- script: emissions
- script: operation-costs
- script: network-layout
  parameters:
    network-type: DH
- script: lake-potential
```

This runs a workflow based on the `default` config file (`default.config`) - other values are `user` (use `cea.config`), `.` (use the one provided to the script or, as in the second config step, the one currently modified) and a path to a user-specified config file can also be used. Inside a config step, environment variables are interpolated (e.g. `${RAYPATH}`) most notably, all the original config variables are saved in the format `${cea_[section-name]_[parameter-name]}` - so elements of the user config file (like daysim bin path and current project) can be used in conjunction with the default config. 

The `script` steps specify a cea script (similar to how to run via CLI) and an optional dict of parameters.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -15,3 +15,6 @@ recursive-include cea/interfaces/dashboard *.html *.js *.css *.otf *.eot *.svg *
 
 include cea/plots/naming.csv
 include cea/glossary.csv
+
+include cea/tests/heating-case.yml
+include cea/tests/cooling-case.yml
--- a/cea/config.py
+++ b/cea/config.py
@@ -430,6 +430,23 @@ class WeatherPathParameter(Parameter):
         return weather_path
 
 
+class WorkflowParameter(Parameter):
+    typename = "WorkflowParameter"
+    examples = {
+        "heating-case": os.path.join(os.path.dirname(__file__), "tests", "heating-case.yml"),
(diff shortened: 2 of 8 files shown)


daren-thomas:
@JIMENOFONSECA this might be a good script to run during your optimization thing every evening before you go home - you can then see how far it got when you come back in the morning.

also, it can be used for parameter studies etc.

PR #2232, opened 2019-08-07
Update occupancy_model.py

gabriel-happle:
The normalization of the mixed-use schedules is now taking the area-share of use into account during normalization.
After averaging the schedules all values are multiplied by the `normalization_value * area_represented` where the `area_represented` is the total fraction of area of different uses with non-zero contribution to the schedules.


gabriel-happle:
This bug affects the results. Therefore the unittest results have to be updated.

B01 is affected as follows:

image

The building is 25% COOLROOM and 75% FOODSTORAGE
- COOLROOM has 8 W/m2 refrigeration loads
- FOODSTORAGE has 0 W/m refrigeration loads
-> the new schedule applies only to 25% of the area, hence the roughly 74% reduction in `Qcre_sys_kWh`

- COOLROOM has 0 warm water consumption
- FOODSTORAGE has 2 lpd warm water consumption
-> the new schedule applies to 75% of the area, hence the roughly 21% reduction in `Qww_sys_kWh`

gabriel-happle:
Regarding the other buildings, we expect changes in mixed use buildings where one of the archetype internal loads is a zero value.
The affected buildings are:
- B01 : COOLROOM/FOODSTORE -> difference see above
- B03 : HOSPITAL/INDUSTRIAL -> process electricity in hospital is zero vs non-zero in industrial
- B07 : SERVERROOM/OFFICE -> appliances and warm water in office vs data in serverrooms
- B09 : OFFICE/PARKING -> appliances and warm water in office

image

gabriel-happle:
@martin-mosteiro I'm tagging you here so you see what is going on in your occupancy model.

gabriel-happle:
The test for the mixed-use schedule creation uses a building of 50% OFFICE and 50% INDUSTRIAL

As expected with the fix, the schedule values for process heat and process electricity are reduced to 50%, because they apply only to 50% of the area.

The new value is correct `Aef * share_industry * hourly_schedule_value * monthly_schedule_value` = `3373 m2 * 0.5 * 0.8 * 0.22 = 297 m2`.

gabriel-happle:
@JIMENOFONSECA I don't know why the jenkins does not like it.
With the changes to the tests (explained above) my local tests in console and pycharm pass.
```
.  run_unit_tests
...........................
Ran 27 tests in 155.099s

OK
```

gabriel-happle:
Closing this as a non-issue. New related issues #2233 and #2234 were created instead.

PR #2251, opened 2019-08-15
Change naming of weather files

JIMENOFONSECA:


daren-thomas:
First quick test: `locator.get_default_weather()` creates a `RuntimeError: maximum recursion depth exceeded in cmp`.

This PR hasn't been tested properly yet. Please don't merge.

daren-thomas:
Where is the rationale for this change? For documentation purposes, connecting this to an issue that explains _why_ we're changing the naming would be a good idea.

daren-thomas:
The build failed because of an exception:

> Exception: Temperature in RC-Model of building B05 out of bounds! First occured at timestep = 4195. The results were Tint = 50.1732998021, theta_c = 49.7425616687, theta_m = 49.3880330959, Check building geometry and internal loads! Building might be too small in size or architecture parameter Hs = 0.95 might be too small for this geometry. Current bounds of range for RC-model temperatures are between -30.0 and 50.0.

JIMENOFONSECA:
@daren-thomas I erased the weather files that we normally do not use. I keep those that we created on our own and some of meteonorm

JIMENOFONSECA:
@daren-thomas any update?

PR #2253, opened 2019-08-16
Feature: Supply system map plot

### LINKED ISSUE #2237: introduce supply systems map plot
this is a map presenting the district heating, and cooling networks after optimization

reyery:
This PR resolves #2237 

Feature
-
- Users are now able to add the map of supply systems as a plot widget in the dashboard.

systemmap

Testing
-
- Run network optimization and add multiple supply system maps to dashboard.
  - Each map should be able to work i.e. toggle network/layers/3d (even when adding maps showing the same network/parameters)


daren-thomas:
@reyery I can't test this because the `thermal-network-optimization` script doesn't run...

daren-thomas:
oh wait... i can run it from pycharm...

reyery:
@daren-thomas were you able to run it? Here is a project you could use to try it https://we.tl/t-rD1nsLoZwm

daren-thomas:
i'm still running the optimization. i'll kill that task just before the meeting, so... thank you for the project. i don't think i have time to test it before the CEA SCRUM.

JIMENOFONSECA:
@daren-thomas to me I can merge this. It works in my computer. Let me know. I believe we will need this to pass to see how it works with the implementation of react of reynold soon

Issue #2256, opened 2019-08-16
decentralized script fails in cli, dashboard, workflow


daren-thomas:
The decentralized script fails because the parameter `{data-helper:region}` is not configured for this script. Which is correct: It shouldn't be configured for that script.

The problem can be traced back to `cea.optimization.preprocessing.decentralized_building_main.disconnected_building_main`, line 38ff:

```
    if (buildings_name_with_heating != [] and config.data_helper.region != 'SG'): 
        #FIXME: temporal fix to avoid heating calculation in SG
        decentralized_buildings_heating.disconnected_buildings_heating_main(locator, total_demand,
                                                                            buildings_name_with_heating,
                                                                            config, prices, lca)
```

This is a bug, anyway, as region `SG` should not be used to test whether heating is necessary or not.

I'm not the domain expert here, but my hunch is, heating should only be done for buildings that have a heating system? Is there a way to test for that? I'm pretty sure there has to be, right?

JIMENOFONSECA:
@shanshanhsieh I beleive you integrated this change right? can you remind us the rationality please?

shanshanhsieh:
changed to checking if the no space heating systems are installed (T0)

Issue #2255, opened 2019-08-16
thermal-network-optimization does not run from cli / dashboard / workflow


daren-thomas:
The `thermal-network-optimization` script has some problems running through. Mainly, this is because of how the `scripts.yml` defines the parameters that a script expects. 

It's important (especially for scripts like `thermal-network-optimization`) for the parameters to be explicitly defined, as we intend to run such scripts in other contexts (e.g. Euler / Amazon cloud etc.).

I suggest a refactoring:

- rename `Network_info` to `NetworkInfo` (pep8) and put in a separate file
  - document it better.
- remove the `NetworkInfo.config` attribute and replace with the specific fields used
  - `network_info.config.network_type` becomes `network_info.network_type` etc.
- the constructor of `NetworkInfo` should accept a `cea.config.Section` instead of a `cea.config.Configuration` parameter that has the values to be copied - that way, we can pass either `config.sections["thermal-network"]` or `config.sections["thermal-network-optimization"]`.
- remove the `gv` parameter while we're at it.
- replace the call to `thermal_network.main` (which expects a `config` object) with a call to `thermal_network.thermal_network_main` (which still expects a `config` object, but we'll be refactoring that too
- class `ThermalNetwork` needs to be refactored to not store a `config` object - same strategy as for `NetworkInfo`.
  - also, what is the difference between `NetworkInfo` and `ThermalNetwork`? That difference should motivate a better (more explanatory) name for `NetworkInfo`. I think `ThermalNetwork` can stay.

Pro Tip: Start the refactoring of `thermal_network.py` first - this script already works, so it can be tested while refactoring.

daren-thomas:
Status update: `2255-thermal-network-optimization-refactoring` branch.
- created NetworkInfo, NetworkLayout and ThermalNetwork objects
- `thermal-network` works again
- `thermal-network-optimization` doesn't work yet
- problems: 
  - `network_info` should contain info needed to create `ThermalNetwork` object
  - updating of `network_info` not explicit
  - (see translate_indiviual, cost function and objective_function)
  - good starting place: look for `network_info.config` and remove slowly.
  - replace `network_info.config.thermal_network` with `thermal_network` (when to create?)
  - look into encoding / decoding individuals (maybe this can be refactored out to an "adaptor" class

daren-thomas:
[...]

Issue #2135, opened 2019-07-03
cooling-case-workflow crashes during multi-criteria-analysis


daren-thomas:
The `cooling-case-workflow` has stopped working. The issue seems to be in the `multi-criteria-analysis`. It looks like a variable (`df_current_individual`) was renamed or something.

```
Running multicriteria with scenario = c:\users\assist~1\appdata\local\temp\2019-07-01-16-32-46-cooling-case-workflow\reference-case-cooling\baseline
Running multicriteria for generation = 2
compiled data of all individuals in this generation:  2
Traceback (most recent call last):
  File "C:\Users\Assistenz\Documents\CityEnergyAnalyst\Dependencies\Python\Scripts\cea-script.py", line 11, in <module>
    load_entry_point('cityenergyanalyst', 'console_scripts', 'cea')()
  File "c:\users\assistenz\documents\cityenergyanalyst\cityenergyanalyst\cea\interfaces\cli\cli.py", line 61, in main
    script_module.main(config)
  File "c:\users\assistenz\documents\cityenergyanalyst\cityenergyanalyst\cea\tests\cooling_case_workflow.py", line 88, in main
    run(script, **kwargs)
  File "c:\users\assistenz\documents\cityenergyanalyst\cityenergyanalyst\cea\tests\cooling_case_workflow.py", line 55, in run
    f(config=config, **kwargs)
  File "c:\users\assistenz\documents\cityenergyanalyst\cityenergyanalyst\cea\api.py", line 28, in script_runner
    script_module.main(config)
  File "c:\users\assistenz\documents\cityenergyanalyst\cityenergyanalyst\cea\analysis\multicriteria\main.py", line 587, in main
    multi_criteria_main(locator, config)
  File "c:\users\assistenz\documents\cityenergyanalyst\cityenergyanalyst\cea\analysis\multicriteria\main.py", line 61, in multi_criteria_main
    config), ignore_index=True)
  File "c:\users\assistenz\documents\cityenergyanalyst\cityenergyanalyst\cea\analysis\multicriteria\main.py", line 366, in preprocessing_cost_data_DC
    if df_current_individual['Data Centre'][0] == 1:
NameError: global name 'df_current_individual' is not defined
```

daren-thomas:
@JIMENOFONSECA is this something you could handle?

JIMENOFONSECA:
[...]

PR #2257, opened 2019-08-16
FEATURE: lazy loading of cea.api module

daren-thomas:
This PR fixes some stuff with the cooling-case workflow. To test, run:

```
cea workflow --workflow cooling-case
```

from the CEA Console. The cooling-case does not run through due to issues #2256, #2255 and #2135 (which I can't test, because the workflow doesn't get that far)

Still, I think merging this now is a good idea. As it addresses all the issues with the workflow up to the decentralized script.

In addition (as per the PR title) it adds a cool feature: Lazy loading of the `cea.api` module. This will increase the startup-speed of the `cea dashboard` script as well as `cea workflow` by not loading cea script modules until they're actually used.


daren-thomas:
@JIMENOFONSECA could you please test this and merge? I'm missing this functionality in other branches and would like to build up on top of it.

Issue #2152, opened 2019-07-11
`data-helper` must respect existing scenario databases


gabriel-happle:
The `data-helper` creates a copy of the archetypes databases in the scenario folder.
If I want to modify my archetypes, I will change values in this local databases.
But then how do I apply these changes?
If I run the `data-helper` again, it will overwrite the archetypes with the default database and hence never be able to apply changes...

We must have an option to use the existing scenario databases!
Or am I missing something?

This is related to #2036

daren-thomas:
Maybe the data-helper needs to be split up into two levels?
OR: don't overwrite archetypes if they're already in the folder?

SO: data-helper should not create / alter `inputs/technologies`. Instead, this should be created by `new-scenario` tool.

OR: data-helper should split "technologies" from databases parameter, add it as a boolean (overwrite/create technologies folder)

building models vs. region-specific data.

Also: re-think the naming everywhere!

gabriel-happle:
yes. This is fixed in #2265

Issue #2036, opened 2019-05-31
Archetypes Database is not really flexible accross multiple scenarios


gabriel-happle:
@shizhongming approached me with a question:
- how to modify archetypical construction across multiple (>100) scenarios?

Turns out, that is actually not easily possible. It seems either one has to modify the cea databases (and potentially run into trouble when updating from master/installing a new version) or one has to manually copy-paste/modify the database 100 times.

I think one possible enhancement could be to set up a custom path (i.e., in the data-helper config, maybe even replacing the `region` parameter) to a separate location of local databases.
I.e., when running the data-helper, the user could choose whether the default cea databases should be used or whether the databases should be taken from another location.

What do you think @daren-thomas and @JIMENOFONSECA ?

gabriel-happle:
Maybe in this way, we could create a repository of database structures from different users to be shared and exchanged.
I.e., the workflow would be to copy one of the existing databases and play around with the valid parameters while keeping the same structure, just in another folder anywhere on the computer.

daren-thomas:
I kind of like this idea. I think it could work similar to the weather files, which are either the name of a built-in weather file _or_ the path to a weather file. The region parameter (how cool is it that it's only used in the data-helper now!?) could be just that. Either a built-in region name _or_ the path to a region database folder structure.

daren-thomas:
let's copy archetypes as part of scenario creation and de-couple it from data-helper. let's revisit the workflow here.

daren-thomas:
Let's make the RegionParameter also accept a folder that will be used instead.

PR #2265, opened 2019-08-19
FEATURE: allow user-specified region databases for the data-helper

daren-thomas:
This PR updates the `data-helper` tool to allow the user to specify a path for `{data-helper:region}` instead of one of the pre-assigned region values (e.g. "CH" and "SG"). This allows users to create their own region templates.

The data-helper does its best to figure out if the path is a valid region template - defined as containing the same files as the default template (CH database).

In addition, the "technology" database choice was removed, as it didn't work like the other choices - it has been replaced with the parameter `{data-helper:overwrite-technology-folder}` which communicates more clearly what happens - the user can now run the data-helper multiple times _without_ overwriting the regional database (`inputs/technology/*`) by setting `overwrite-technology-folder` to `False`. This addresses #2152.

To test this, try the following:

- test the data-helper in CEA Console by setting the `--region` parameter to
  - SG and CH
  - an invalid path (non-existant)
  - an invalid path (not a region template)
  - the `inputs/technology` folder of another scenario
- test the data-helper tool in the CEA Dashboard, similar to above

This should solve the user story in #2036.

### DIFF
--- a/cea/config.py
+++ b/cea/config.py
@@ -635,7 +635,7 @@ class ChoiceParameter(Parameter):
     typename = 'ChoiceParameter'
 
     def initialize(self, parser):
-        # when called for the first time, make sure there is a `.choices` parameter
+        # when called for the first time, make sure there is a `._choices` parameter
         self._choices = parse_string_to_list(parser.get(self.section.name, self.name + '.choices'))
 
     def encode(self, value):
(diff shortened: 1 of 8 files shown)


daren-thomas:
@gabriel-happle due to the way the Dashboard works, it's kind of hard to test for the paths to be correct. I suggest making this a separate issue that we fix with it's own PR: #2267

Then, we can merge this PR and address the error message at a later time, what do you think?

Issue #2246, opened 2019-08-13
Remove old cooling-case-workflow script


daren-thomas:
When creating the `cea workflow` script, the workflow yaml file was introduced and both heating-case and cooling-case work implemented using that.

This left the file `cea/tests/cooling_case_workflow.py` unused. It should be deleted.

PR #2266, opened 2019-08-19
Delete cooling_case_workflow.py

daren-thomas:
This PR addresses #2246 - the old cooling-case-workflow script is now removed and replaced with the `cea workflow --workflow cooling-case` script.


daren-thomas:
@reyery I'm just going to go ahead and merge this since it's been lying around for a while and doesn't break any functionality...

Issue #2053, opened 2019-06-07
Process schedules for LAB and HOSPITAL are questionable


martin-mosteiro:
As mentioned in #1932 and #2034, I don't think the process schedules for LAB and HOSPITAL as defined in CEA make sense. Since there are no devoted schedules for these uses in the SIA standard, these were created as a copy of the process schedules for INDUSTRY, however this archetype was created based on data from a heavy manufacturing company in Switzerland. The schedules for processes follow a very specific pattern that likely makes sense in such a context (e.g. the ramp down during Saturdays), but I don't think it is especially representative for labs and hospitals (especially since these are occupied 7 days a week).

I would suggest making these schedules mirror the corresponding function's electricity demand patterns instead (both in CH and SG). This will nonetheless be a simplifying assumption, but one that is based on the specific function being modeled and not some large-scale manufacturer in Switzerland.

The alternative, of course, is researching appropriate process energy schedules for the corresponding functions.

martin-mosteiro:
I'm co-assigning @gabriel-happle since this came up in an issue he created.

martin-mosteiro:
Probably worth getting your opinion on what to do with the SG process schedules, though @gabriel-happle

gabriel-happle:
@martin-mosteiro I would have to do some research. no idea about it right now.

Issue #1932, opened 2019-05-06
COOLROOM and LAB do not have archetypical process schedules defined


gabriel-happle:
While working on #1794 it became evident that COOLROOM and LAB do not have process schedules defined for calculating refrigeration loads and process electricity consumption respectively.

In `occupancy_model.py` line 578 it is obvious that only INDUSTRIAL has a process schedule defined.

process schedules have to be added to the `occupancy_schedules.xlsx` database and the `read_schedules` function has to be adjusted.

martin-mosteiro:
@gabriel-happle The issue with COOLROOMs is addressed in issue #1930.

Regarding LABs, I remember creating a process schedule for these back when I created the LAB archetype (just copied the INDUSTRIAL process schedule), but maybe it got lost along the way? And in any case I don't think a large manufacturer (as used to define the process schedules for INDUSTRIAL) is a valid reference for a LAB. Maybe these should just use the electricity schedules as process schedules?

gabriel-happle:
@martin-mosteiro let me know when #1930 is fixed.

Regarding the LAB, in that case, it would be easier to just increase the equipment load and get rid of the process load.

martin-mosteiro:
I agree. Although in more general terms, I think the "process" calculations in general are not very clear at the moment. Besides, demands for process heating or cooling are probably served by dedicated, separate systems (like in COOLROOMs and SERVERROOMs).

martin-mosteiro:
However, getting rid of the process schedule in labs would mean that LABs can never have process electricity. Which I guess is fair if we assume that processes in LABs are just lumped into appliances.

JIMENOFONSECA:
@gabriel-happle @martin-mosteiro do not get rid of this. They are used in optimization as sources of waste heat, and sources that need specific boilers (due to the temperature needs)

martin-mosteiro:
@gabriel-happle #1994 will solve part of the issue described here. The missing part is the process schedules for LABs.

gabriel-happle:
@martin-mosteiro cool. I will copy the Industry schedule for now. And then make a PR soon

gabriel-happle:
@martin-mosteiro you were right! there is a process schedule for LAB, but only in the CH archetypes.
I was checking the reference-case-cooling at the time I made this issue.
I will just add the same schedule to the SG archetypes and test quickly in both regions.

gabriel-happle:
[...]

PR #2275, opened 2019-08-22
2063 process cooling

martin-mosteiro:
Added process cooling as an energy demand. The archetypal value for process cooling is 0.0 for all occupancy types and follows the schedule for processes. Unless the user provides a process cooling demand for a given building, the output will be 0.


cea/demand/occupancy_model.py
```diff
@@ -84,8 +85,8 @@ def calc_schedules(list_uses, archetype_schedules, bpr, archetype_values, stocha
         for schedule in ['people', 've', 'Qs', 'X', 'Vww', 'Vw']:
             schedules[schedule] = np.zeros(HOURS_IN_YEAR)
         # electricity and process schedules may be greater than 0
-        for schedule in ['Ea', 'El', 'Qcre', 'Ed', 'Epro', 'Qhpro']:
-            codes = {'Ea': 1, 'El': 1, 'Ed': 1, 'Epro': 3, 'Qhpro': 3,'Qcre': 3}
+        for schedule in ['Ea', 'El', 'Qcre', 'Ed', 'Epro', 'Qhpro', 'Qcpro']:
+            codes = {'Ea': 1, 'El': 1, 'Ed': 1, 'Qcre': 1, 'Epro': 3, 'Qhpro': 3, 'Qcpro': 3}
```
JIMENOFONSECA: Why Qcre changed?, I believe it is 3. It should not be accounted as an internal load
COOLROOMS
SERVERS
QCPRO
QHPRO
are all isolated from the R-C model under the assumption that they have their own HVAC system. Maybe I am missing something?

cea/demand/thermal_loads.py
```diff
@@ -101,14 +104,16 @@ def calc_thermal_loads(building_name, bpr, weather_data, date_range, locator,
...
+        #CALCULATE SPACE CONDITIONING DEMANDS
         tsd = calc_set_points(bpr, date_range, tsd, building_name, config, locator)  # calculate the setpoints for every hour
         tsd = calc_Qhs_Qcs(bpr, tsd, use_dynamic_infiltration_calculation)  #end-use demand latent and sensible + ventilation
         tsd = sensible_loads.calc_Qhs_Qcs_loss(bpr, tsd) # losses
         tsd = sensible_loads.calc_Qhs_sys_Qcs_sys(tsd) # system (incl. losses)
         tsd = sensible_loads.calc_temperatures_emission_systems(bpr, tsd) # calculate temperatures
         tsd = electrical_loads.calc_Eauxf_ve(tsd) #calc auxiliary loads ventilation
         tsd = electrical_loads.calc_Eaux_Qhs_Qcs(tsd, bpr) #calc auxiliary loads heating and cooling
+        tsd = calc_Qcs_sys(bpr, tsd) # final : including fuels and renewables
+        tsd = calc_Qhs_sys(bpr, tsd) # final : including fuels and renewables
```
JIMENOFONSECA: please take a second look at this. this seems a double calculation.... there is this function calc_Qhs_sys_Qcs_sys.
for renewables, they are accounted different, that is inside E_hs_PV etc...

cea/tests/test_calc_thermal_loads.config
```diff
@@ -1,7 +1,7 @@
 [test_calc_thermal_loads]
 value_columns = ["E_sys_kWh", "Qcdata_sys_kWh", "Qcre_sys_kWh", "Qcs_sys_kWh", "Qhs_sys_kWh", "Qww_sys_kWh", "Tcs_sys_re_C", "Ths_sys_re_C", "Tww_sys_re_C", "Tcs_sys_sup_C", "Ths_sys_sup_C", "Tww_sys_sup_C"]
-values = [333157.804, 0.0, 35057.826, 136873.46899999998, 164227.413, 24425.425, 0.0, 0.0, 26186.16, 0.0, 0.0, 171600.0]
+values = [332591.53400000004, 0.0, 35057.826, 136848.49699999997, 164146.24, 24425.275, 0.0, 0.0, 26186.16, 0.0, 0.0, 171600.0]
```
JIMENOFONSECA: why did it change? Qcs_sys? Qcpro should not contribute at all at this...

cea/demand/occupancy_model.py
martin-mosteiro: Cool rooms don't have process schedules in CEA. In fact, refrigeration was already calculated based on the electricity schedule before I mistakenly changed it a while back.

cea/tests/test_calc_thermal_loads.config
martin-mosteiro: It doesn't – but I also changed the process schedules now based on discussions in #1932 and #2053.

martin-mosteiro:
Not sure why this isn't passing. It passes locally on my computer. On Jenkins, it appears to be an issue with `calc_termal_loads` but I already updated the unit test data accordingly (53b69bccfab23c6da52e0abee20750a6cd2f3274), so I'm not sure why the Jenkins is still comparing to the old unit test data.

JIMENOFONSECA:
@daren-thomas maybe you can provide access to everyone to the Jenkins? So Martin can check the build? this will take some load from you.

martin-mosteiro:
@JIMENOFONSECA I've checked it, long story short: the thermal loads test fails due to missing updated results from the corresponding test data creator. However... I _did_ create and push that data. And the tests pass locally on my computer.

@daren-thomas Said it might be an issue from having tested on a Mac. I will try it on a Windows computer now.

Issue #1271, opened 2018-05-03
broken links on the CEA website


amraladdin:
Seems some links on the CEA website are broken. Here is what I found so far:
http://city-energy-analyst.readthedocs.io/en/386-create-readthedocs-style-api-documentation/index.html
You can find this broken link, navigating to: https://cityenergyanalyst.com/overview -> Choose any feature from the left list pane -> then scroll down till the end of the page and click on the **“CEA Developers manual and API”** link (Check an example in the image below).

![image](https://user-images.githubusercontent.com/16256867/39567209-83f371de-4ebe-11e8-902e-adc900b51fc4.png)

Could someone please fix them?
@JIMENOFONSECA @gabriel-happle @BhargavaKrishnaSreepathi @shanshanhsieh @martin-mosteiro @shizhongming

JIMENOFONSECA:
on it

JIMENOFONSECA:
@amraladdin done

amraladdin:
Super!

PR #2286, opened 2019-09-03
Fix to #1271

### LINKED ISSUE #2271: optimization tool: Missing geothermal potential of the scenario. Consider running lake potential script first.
This is an issue posted by @kerembeygo (transferred from #2259)

After running `water-body-potential` tool, the `optimization` tool raises this error. (NOTE: the error message needs to be updated to reflect the name change in the tool. Also, the input files system should have caught this first!)

JIMENOFONSECA:
@daren-thomas this should fix #2271 let me know if this works


cea/optimization/optimization_main.py
```diff
@@ -147,10 +150,10 @@ def check_input_files(config, locator):
         raise ValueError(
             "Missing sewage potential of the scenario. Consider running sewage heat exchanger script first.")
     if not os.path.exists(locator.get_lake_potential()):
-        raise ValueError("Missing lake potential of the scenario. Consider running lake potential script first.")
+        raise ValueError("Missing water-body potential of the scenario. Consider running lake potential script first.")
```
daren-thomas: "Consider running _lake potential_ script first"... I think only half of that error message was updated. Additionally, I think the locator method (`get_lake_potential`) could also be updated to `get_water_body_potential`.

cea/scripts.yml
```diff
@@ -248,8 +248,15 @@ Optimization:
     description: Run optimization for the given scenario
     interfaces: [cli, arcgis, dashboard]
     module: cea.optimization.optimization_main
-    parameters: ['general:scenario', 'thermal-network:network-type', 'thermal-network:network-names',
-                 'optimization', 'decentralized', 'general:multiprocessing']
+    parameters: ['general:scenario', 'general:multiprocessing', optimization]
+    input-files:
+      - [get_street_network]
+      - [get_total_demand]
+      - [get_lake_potential]
```
daren-thomas: note, when updating the InputLocator method this would need to change too...

daren-thomas:
@JIMENOFONSECA there are still some small changes here for you to make before we can merge.

cea/optimization/optimization_main.py
JIMENOFONSECA: done!

JIMENOFONSECA:
@daren-thomas please merge if the PR is ok now, thanks

PR #2288, opened 2019-09-03
FEATURE: New React Dashboard

### LINKED ISSUE #2163: Keep plots in their slots in map dashboard when deleting / adding plots
When deleting the plot next to the map (by clicking the X top-right), I expected that plot slot to be free again. instead, the plots all "moved up". I then went and tested adding a plot to an empty map dashboard, by selecting the last slot (lower-right position) and there the same behavior. 

Also, changing the order of the plots doesn't seem to be possible. Could this be added?

reyery:
This PR changes the current dashboard view to use **React** so that interacting within the view will not refresh the whole page.

Changes
-
- Plotly scatter plots to use `webgl` rendering instead of `svg` to improve rendering time
- Removed `description` property of plots since it is rarely used
- Able to add and remove plots in position when using grid layout (resolves #2163)
- Editing/Adding plots to dashboard will not refresh page
- "Set scenario" button that changes the scenario parameters of every plot of the dashboard to a particular value
- Removed map layout (map can be instead added to the dashboard by using `supply system map` plot)
- Has the same functions as the previous version minus a few unimplemented ones i.e. show table, move plots up/down


reyery:
@JIMENOFONSECA @daren-thomas would need your help to test this on your computers to make sure that it is still compatible with other `dashboard.yaml` files (do make a copy of them first just in case something happens). I will remove all the code of the old dashboard once this is up to standard.

Testing
-
*** **Remember to backup your dashboard.yaml file if it is important** ***

### View Dashboard
Plots should show the loading spinner when the plot is still being calculated in the backend or it is not its turn to be rendered (so apparently **Plotly** can only render one plot at a time, more data heavy plots will take longer to load)

Unfortunately, the UI interactions might freeze when it is rendering a plot or slow down if there are too many data heavy plots (largely depends on the performance of the computer). The only solution I can think of is to pass every other calculation to background workers while keeping React on the main thread of the browser.

- Change the dashboard by using the select box in the top left
  - The selected dashboard should load up

### Add/Change/Edit Plot
I am planning to somehow merge these functions together since it would be good to be able to select the desired parameters when adding/changing a plot, instead of adding/changing a plot -> wait for plot to load -> set desired parameters -> wait for plot to load again

- Edit/Change Plot using the edit icon on the top right of the plot
  - Changes should be applied in position while every other plot is unaffected

- Add Plot using the add plot button either found in empty plot widgets or on the bottom right
  - New plot should be added without affecting other plots

### Delete Plot
Deleting plots should work well for the grid layout where an empty plot widget would take its place. However, deleting plots in the row layout would re-render every plot that was below it. This is due to how React deals with rendering an array of components and the key value that identifies each of the elements that helps it decide what to re-render. 

To put it simply, I would need to find another way of setting a unique key for each element instead of using `{dashboard index}-{plot index}`.

### New/Duplicate Dashboard
- Click the create/duplicate dashboard buttons, fill up the form and click OK
  - It should switch to the new dashboard after it is created

### Set Scenario
- Click the set scenario button, select the scenario and click OK
  - The selection should include all the present scenarios in the project
  - Every plot that has the `scenario-name` parameter would be changed to the selected scenario, those without would be unchanged.

daren-thomas:
I just opened a dashboard from some other testing, as suggested above and there was an error. This is how the error is displayed to the user: 

image


To test, the project:

    Current Project: CASE-EXECUTIVE-COURSE
    Scenario: baseline

dashboard.yml:
```
- layout: grid
  name: row
  plots:
  - category: comparisons
    parameters:
      urban-energy-system-scenarios: [Baseline, none, none, Baseline, '2', '1', Baseline,
        '2', '2']
    plot: comparisons_annual_costs
  - category: supply-system
    parameters: {generation: 1, individual: '2', scenario-name: baseline, timeframe: monthly}
    plot: dispatch_curve_electricity
  - category: supply-system
    parameters: {generation: 50, individual: '10', scenario-name: baseline, timeframe: monthly}
    plot: dispatch_curve_heating
  - category: supply-system
    parameters: {generation: 1, individual: '13', scenario-name: baseline, timeframe: monthly}
    plot: requirements_curve_electricity
```

daren-thomas:
Stuff that works (this is awesome btw!)

- plots load with the spinner
- changing a plot / editing parameters does not require reload of the page (yay)
- changing dashboard works and is nice and fast!
- deleting a plot even in row layout didn't seem to re-draw the whole thing?
- duplicating a dashboard works!

Stuff that didn't work

- flask-restplus was not installed with `pip install -e .`, needs to be added to `setup.py`
  - oops. i just saw there was a new version
- the dropdown for toggling the legend is confusing ("on", "off", but you need to toggle it a couple of times to figure out what it does) - also, i think it should be some kind of toggle button and not a dropdown box.

Additional thoughts:
- I'm not sure if the "set scenario" functionality fits. I _do_ see it's usefulness. I guess... I guess if users create a new scenario and then copy a dashboard then they could use this to update all the plots in one go. So it's kind of cool! It's also a bit weird from a purist perspective. From the python zen: "practicality beats purity".


I vote to merge once the error above (the plot not showing with the html code being printed) is fixed and we create a separate issue for the dropdown toggling of the legend. Unless that can be done quickly here?

daren-thomas:
BTW: How is the legend toggle implemented? And: shouldn't it be part of the plot header? like... next to the change plot thing?

image

there?

reyery:
> - deleting a plot even in row layout didn't seem to re-draw the whole thing?

it doesn't redraw the whole thing, the plots below it would re-render since their index has changed. It is being rendered from an array of plots using the `map()` function ([React: List and Keys](https://reactjs.org/docs/lists-and-keys.html)). 

Currently I am using the `{dashboard_index}-{plot_index}-{hash}` as the key to render the list of plots, this works well to differentiate same plots (with the same parameters) in the dashboard but since it relies on `plot_index`, when it changes, the plot will be re-rendered by React. We could take out `plot_index` but having the same keys makes React unhappy.

So I want to come up with another way of differentiating these plots when they are on the same dashboard. I was thinking of giving the plots some sort of unique ID when it is first created so at least even with the same parameters (hash) we can still tell that they are different entities. And then we can use that as the key for the React component list.

reyery:
> I just opened a dashboard from some other testing, as suggested above and there was an error. This is how the error is displayed to the user:

@daren-thomas Ah, the error is intentionally shown to the user like this when there it receives a `500` error (runtime error in **Flask**). It is supposed to show the stack trace of the error so that the user could use it to report the issue. But it seems like you are running **Flask** in `debug` mode so its showing the error html file that it shows during an error.

Can't really tell what the error is other than it being some sort of `KeyError`, as it works fine on my side using the dashboard.yaml you provided (I also deleted .cache).

daren-thomas:
@reyery I love the toggle legend thing! Is this something that was implemented in each plot? because, while it's now moved to the header, it still shows up as a dropdown on the plots?

reyery:
@daren-thomas the dropdown is embedded in the plot div script when it was created in Flask, to clear it you would need to recalculate the plot (delete if from cache).

daren-thomas:
amazing! cool! let's merge!

reyery:
@daren-thomas should we remove the old code for the dashboard?

daren-thomas:
@reyery yes please: let's not keep old code lying around :)

PR #2289, opened 2019-09-03
2255 thermal network optimization refactoring

### LINKED ISSUE #2255: thermal-network-optimization does not run from cli / dashboard / workflow

daren-thomas:
This PR contains all the work I did on refactoring `thermal-network-optimization` to fix #2255. There are some issues (and bugs) still in the code. But for whole-year simulations (or [start-t, stop-t] bigger than number-of-plants) it should run through.

To test the (current) bug, set `start-t = 8759, stop-t = 8760`. This probably never really worked.

Also, I'm not confident that multithreading works for this. Due to the way files are read / written - it's hard to reason about.

But the main part of the issue, it not running through, should now be fixed.


JIMENOFONSECA:
@shanshanhsieh any progress wiht this PR?

shanshanhsieh:
- tested `thermal-network` without multi-processing from cea console
- tested `thermal-network` with multi-processing from cea dashboard
- compared results with&without multi-processing, as well as results from master

@daren-thomas , i noticed that the computational time is 1.3 times longer than master (single-process), do you perhaps know where does this come from?

PR #2290, opened 2019-09-04
changing workflows structure

JIMENOFONSECA:


daren-thomas:
This works as in it the refactoring is now working. The heating case workflow doesn't run through, though. I'll file a separate issue for that.

PR #2292, opened 2019-09-05
steps for creating a release updated

JIMENOFONSECA:


docs/communication.rst
```diff
@@ -7,39 +7,39 @@ CEA website
 ------------
 
 Host: squarespace.com
-Credentials: Sign-in with CEA gmail account uesm.fcl@gmail.com in `here <https://docs.google.com/document/d/19rHXlRvjB3dfZHjWvZ8GMd2tZ0aeGgznsiyd8FChGIg/edit?usp=sharing>`__.
+Credentials in `here <https://docs.google.com/document/d/19rHXlRvjB3dfZHjWvZ8GMd2tZ0aeGgznsiyd8FChGIg/edit?usp=sharing>`__.
```
daren-thomas: @JIMENOFONSECA I think it's great that we removed the email address - but... keeping passwords in a google document that we then post on the internet for all to find - even if there is authorization involved - seems like a security risk. We could just change it to "Ask Product Owner for access"?

docs/how-to-create-a-new-release.rst
```diff
@@ -157,3 +157,22 @@ Uploading to PyPI
...
 
 .. _twine: https://pypi.python.org/pypi/twine
+
+
+Updating Link in www.cityenergyanalyst.com/tryit
+--------------------------------------------------
+
+- Go to www.cityenergyanalyst.com
+- press Esc and login into squarespace (the credentials are here_)
+- go to Pages/Try CEA  (It is the last page in the list
```
daren-thomas: small typo

docs/communication.rst
JIMENOFONSECA: good point? maybe we can create one link in a long temr repo? maybe ETH zurich / A/S?

daren-thomas: I'm not really sure what a "link in a long temr repo" is... We could just specify that the Product Owner manages the passwords (maybe in a tool like [KeePass](https://keepass.info/)? but the point being that's totally up to the Product Owner) and anyone who needs access can ask him for the user/password or to create one for him... What do you think?

JIMENOFONSECA: I meant the WIKI of A/S, I believe that has more long-term duration than the Product Owner.

daren-thomas: I think the product owner is responsible for the passwords. They shouldn't be hosted anywhere. The idea being, there is always a product owner and he/she passes the collection of passwords on to the next one? Also, note that in the version of `communication.rst` above, the same document is being referenced, even though the access groups are not the same...

JIMENOFONSECA: What is the PO is missing? I believe at least the Scrum Master should have access to. Why not wiki? we used to store passwords there too...

daren-thomas: Well, the Scrum Master gets all the passwords from the PO upon taking on that role. So yes, he/she would have the passwords too. And can act as a vice password manager.

The thing about keeping them in a wiki is that it's not secure. We shouldn't really store passwords in plain text anywhere. You should have something like KeePass or LastPass etc. for storing passwords in. Yes. We used to store passwords in a wiki. That doesn't mean it's a good idea...

JIMENOFONSECA: got it, What about now? :-)

Issue #2192, opened 2019-07-22
Clicking on a plot's name takes me to a single view of the plot


JIMENOFONSECA:
So when we click on the plot's name it takes me to a bigger version of the plot in the same window. This makes difficult to work and come back to where i was working.

The functionality whould be different:

If the user clicks on the plot's name, it should open a new window and take her to a single "big" view of the plot. (it can be the html version of it).  It is not necessary that the plot opens inside teh CEA dashboard...

daren-thomas:
That's not the only solution to this usability bug: We could also display a link "back to dashboard {dashboard name}".

What problems are you trying to solve _exactly_? I'm not a big fan of opening a new window. But I _am_ open to convincing arguments.

The display necessarily _has_ to happen as part of the "dashboard" in the sense that it has to be a URL passing through our flask server, as the "html" version of the plots don't really exist - they're created on-the-fly in the flask server.

JIMENOFONSECA:
got it.

So the main argument:

1. for the user it is just a hassle to come back and reload the entire CEA once she goes into the plot
2. the user wants to see the plot BIG (the entire webpage). like we used to see them before the dashboard.
3. a BIG and separate window leaves more room for exploration. the user can open several and see them in a BIG screen or save them.

would that work?


On 22 Jul 2019, 5:45 PM +0800, Daren Thomas <notifications@github.com>, wrote:

That's not the only solution to this usability bug: We could also display a link "back to dashboard {dashboard name}".

What problems are you trying to solve exactly? I'm not a big fan of opening a new window. But I am open to convincing arguments.

The display necessarily has to happen as part of the "dashboard" in the sense that it has to be a URL passing through our flask server, as the "html" version of the plots don't really exist - they're created on-the-fly in the flask server.

reyery:
would not be issue if moved to electron

daren-thomas:
@reyery how _exactly_ would the move to electron address this issue?

daren-thomas:
Status update: #2288 (FEATURE: New React Dashboard) removed the link in the plot title.

PR #2299, opened 2019-09-11
2192 single plot view

daren-thomas:
This PR implements #2192 - users can click on a plot title in the new (react) dashboard and a new tab opens up with an uncluttered view of the plot.

To test:
- open the dashboard for a project/scenario
- click on the title of a plot
- you should see a new tab with the plot
- try different dashboard styles (row / grid layout)


daren-thomas:
@JIMENOFONSECA assigning to you for now as you opened the issue. Feel free to delegate.

reyery:
It does not work with the supply system map plot, still trying to figure out why. Could be because of how the scripts are ordered/executed when mounting it in the body?

daren-thomas:
@reyery: I figured it out... the map _is_ being plotted. it's just that the container div has no fixed height... to test, open the single plot version of the supply system map plot. use the inspector to navigate to the div with id={hash}. set height="500px" (just an example) - voilà! there's your map.

Now... I'd like some help with you: Could you figure out what we should be using for the "height" attribute? how did you specify it for the react stuff? care to lend a hand? thank you!

reyery:
yeah. I have noticed that too, regarding the height of the div. But I still get something like this when I try to change it to 3D. Still not sure why. I noticed that it will try to download the deckgl script (dist.min.js) even though it's already downloaded. Something to do will jquery `get` caching set to `off` since there are some kind of timestamp behind the js file

image
image

Anyway, regarding the height I am thinking of making it 50% of view port with a min-height of 450, that would mean we would need to resize the plots accordingly with a window resize event listener, will push a commit for this. For React, the div container has a height style property probably that's why it worked.

daren-thomas:
your update works for me (I was nearly about to push something similar). I can't do 3D plots toggling either. BUT: i think we should not bother with this right now, but instead merge and create an issue, since we'll be using the react containers anyway soon - it would just be a waste of resources.

@reyery what do you think?

reyery:
Created issue for supply system map. #2312

Issue #2235, opened 2019-08-08
Change folder structure of saving and loading schedules


gabriel-happle:
With PR #2230 schedule files are read from `input/building-properties` and saved to `output/data/demand`. This change is necessary so that changes in occupancy or other building properties will be applied to the schedule-maker. See #2229 

In the future, a more intuitive folder structure should be used.
This is related to #2185, #2184 and #2183

JIMENOFONSECA:
@gabriel-happle  can you assign a milestone, label, and developer? thanks

daren-thomas:
@martin-mosteiro @gabriel-happle how is this connected to the PR #2302 (Add script to create occupant schedules from MATSim population)?

martin-mosteiro:
I did some minor folder reorganization as part of the MATSim schedule implementation. Might solve this issue or not – up to Gabriel (the creator of this issue and reviewer assigned to PR #2302) to decide.

PR #2302, opened 2019-09-12
Add script to create occupant schedules from MATSim population

martin-mosteiro:
I added a script to read a MATSim population file and create building schedules from it. The schedules are then exported as csv files as done by @gabriel-happle in `occupancy_model.py`.

A few additional things this PR also does:
1. Move occupancy models to a new subfolder `cea.demand.occupancy`. That way all occupancy-related scripts are in one directory.
2. Move schedule inputs folder to `inputs.building-schedules` and schedule outputs folder to `outputs.data.schedules` (#2235) in order to avoid schedule files piling up in the building properties and energy demand folders.
3. Refactor a few parts of `occupancy_model.py` to make data structure somewhat more intuitive – and to make it possible to use it in the new `occupancy_from_transportation.py`.


daren-thomas:
@martin-mosteiro this PR does not include instructions on how to test it. I would also appreciate documentation (an *.rst file in the `docs` folder) on how (and why) to use it.

martin-mosteiro:
@daren-thomas Thanks for your feedback, I'll create the corresponding documentation.

Not sure what the `scripts.yml` file is for. There are many scripts that are not listed in that file. What's the criteria for a script being listed there?

martin-mosteiro:
@daren-thomas Added the files required. Not sure how to test, though, as you need a MATSim case study to be able to use this tool. Which is why when we first discussed this tool I had created I described it as a "niche tool" other people might not be interested in.

daren-thomas:
The `scripts.yml` file sets up a script so that it can be run from the various interfaces (e.g. dashboard or CEA Console). It defines the parameters the file needs as well as what input files need to be present for the tool to run (defined as input locator method names).

So, the term "script" is a bit ambiguous - I see it as a defined set of functionality of the CEA. Anything that isn't accessible this way is a candidate for removal, since you can't really "use" it, right? (unless by running the python module directly either with `python.exe` or inside PyCharm)

In my view, declaring a script in `scripts.yml` makes it officially a part of the CEA. You want your script to be official right?

martin-mosteiro:
@daren-thomas I have now added the following:
1. a "fake" MATSim model for the CEA reference-case-open, comprising a set of MATSim facilities (`facilities.csv`) and a dummy population (`population.xml`) as inputs to the CEA model;
2. a unit test for the CEA schedule maker based on transportation data (`cea.tests.test_schedules.TestTransportationScheduleCreation`) to make sure the results produced by this script remain consistent.
You can test the script by simply running the normal unit test for occupancy schedules, `cea.tests.test_schedules`.

martin-mosteiro:
@daren-thomas Fixed the issue with the testing. Will @gabriel-happle review this? Or @daren-thomas do you wanna check it out again?

JIMENOFONSECA:
@martin-mosteiro @gabriel-happle , the schedules should be saved in an output folder (e.g., outputs/data/schedules/. There should be a "use-matsim-schedules" boolean parameter in config.demand that tells CEA to use those schedules instead of default ones in the next iteration.

OPTIONALY, you can create the Schedule-maker tool. The tool should do :

1. Get default schedules based on CEA databases.
OR
2. Get Matsim schedules based on "extra-input files"

The outputs:

1. Occupancy schedules per building in csv...stored in outputs/data/occupancy_schedules

After that, we could add the schedule maker tool as part of the workflow of CEA and avoid the use of any boolean variable "use-matsim-schedules".

The tool seems not ready to be implemented a it is.

martin-mosteiro:
@JIMENOFONSECA Thanks for your feedback! Some comments:

> @martin-mosteiro @gabriel-happle , the schedules should be saved in an output folder (e.g., outputs/data/schedules/. There should be a "use-matsim-schedules" boolean parameter in config.demand that tells CEA to use those schedules instead of default ones in the next iteration.

Why should the schedules be in the output folder? The option of using schedules from the `inputs` folder already exists, so why add a boolean to get schedules from the outputs instead of the inputs?

I see the point of automatically-generated scripts being placed in an output folder in order to avoid users mistakenly using unwanted schedules (which the normal schedule maker in CEA now does). But if users are running a dedicated tool to create schedules, it seems safe to say they know what they are doing, and it doesn't make sense to me to make them then select a boolean or copy the schedules that they just created to the inputs.

To me this tool works like data-helper tools, it's a preprocessing step that users can use and whose results go in the inputs (where CEA needs them), not the outputs.

> OPTIONALY, you can create the Schedule-maker tool. The tool should do :

Well, that is the long term plan. But I think that is beyond scope here: this issue was not about the schedule-maker tool, it was about the MATSim schedules tool.

> After that, we could add the schedule maker tool as part of the workflow of CEA and avoid the use of any boolean variable "use-matsim-schedules".

Agreed, the schedule maker tool should eventually be part of the CEA workflow.

> The tool seems not ready to be implemented a it is.

Could you please explain what is the issue with this specific tool? I honestly don't understand. The purpose of this tool is for people who have access to MATSim data to generate schedules based on that data. Eventually, it should become one of a suite of schedule creation tools (i.e., deterministic, stochastic, transportation), but since these tools don't yet exist I don't see why this one cannot work as a standalone tool.

JIMENOFONSECA:
Sure. Well i think that if this is going to be a tool of CEA it should be part of the workflow.

I see that a schedule maker tool can as well be this tool you are proposing. All we need to do is that this tool gives the option to create default schedules or matsim schedules (both options). That is all.

Once that is ready, we will include it into the workflow.

The results should be saved in the output folder. That makes the schedules easy to find and examine. Using files in the input folder should be considerd a temporary fix today.

Why i think it is not ready to be a tool? Without becoming a schedule maker, i trend to feel that the code can only be used by one user. I feel that no one else can use it unless they have access to all matsim data (quite unlikely). I just ask myself. Why affectting the workflow, adding extra steps to all users, if the tools can be barely use (number of times * expexted users)

Will this tool be used after spacergy, or muses? Are we expecting a recurrent use of this? If the answer to these questions is negative, we might want to ask ourselves: Should this be a research tool that can go into the legacy folder?

I recently started to ask my elf this question with other tools. We took the mpc controllers tools (2 tools) and the neural network to legacy based on this reasoning. It was the right call, since it avoids unnecessary maintenance for tools that will not be used too often.

My call is to either make the effort now and make this tool the schedule maker, OR put this feature in the legacy of CEA woth documentation.

On 27 Sep 2019, 22:37 +0800, Martín Mosteiro-Romero <notifications@github.com>, wrote:

@JIMENOFONSECA<https://github.com/JIMENOFONSECA> Thanks for your feedback! Some comments:

@martin-mosteiro<https://github.com/martin-mosteiro> @gabriel-happle<https://github.com/gabriel-happle> , the schedules should be saved in an output folder (e.g., outputs/data/schedules/. There should be a "use-matsim-schedules" boolean parameter in config.demand that tells CEA to use those schedules instead of default ones in the next iteration.

Why should the schedules be in the output folder? The option of using schedules from the inputs folder already exists, so why add a boolean to get schedules from the outputs instead of the inputs?

I see the point of automatically-generated scripts being placed in an output folder in order to avoid users mistakenly using unwanted schedules (which the normal schedule maker in CEA now does). But if users are running a dedicated tool to create schedules, it seems safe to say they know what they are doing, and it doesn't make sense to me to make them then select a boolean or copy the schedules that they just created to the inputs.

To me this tool works like data-helper tools, it's a preprocessing step that users can use and whose results go in the inputs (where CEA needs them), not the outputs.

OPTIONALY, you can create the Schedule-maker tool. The tool should do :

Well, that is the long term plan. But I think that is beyond scope here: this issue was not about the schedule-maker tool, it was about the MATSim schedules tool.

After that, we could add the schedule maker tool as part of the workflow of CEA and avoid the use of any boolean variable "use-matsim-schedules".

Agreed, the schedule maker tool should eventually be part of the CEA workflow.

The tool seems not ready to be implemented a it is.

Could you please explain what is the issue with this specific tool? I honestly don't understand. The purpose of this tool is for people who have access to MATSim data to generate schedules based on that data. Eventually, it should become one of a suite of schedule creation tools (i.e., deterministic, stochastic, transportation), but since these tools don't yet exist I don't see why this one cannot work as a standalone tool.

martin-mosteiro:
> Will this tool be used after spacergy, or muses? Are we expecting a recurrent use of this? If the answer to these questions is negative, we might want to ask ourselves: Should this be a research tool that can go into the legacy folder?
Agreed, I also wondered this (and posited it in issue #2302). However, I saw my issue as tangential to the schedule maker. Based on your feedback, I'll leave the MATSim occupancy scheduler as blocked by the issue about the schedule maker.

> I see that a schedule maker tool can as well be this tool you are proposing. All we need to do is that this tool gives the option to create default schedules or matsim schedules (both options). That is all.
> The results should be saved in the output folder. That makes the schedules easy to find and examine. Using files in the input folder should be considerd a temporary fix today.

Okay, so your suggestion would be the following workflow, correct?
0. Run data helper (if necessary)
1. Run radiation (outputs go to `outputs\data\solar-radiation`)
2. Run schedule maker (outputs go to `outputs\data\schedules`); edit schedules in outputs if desired
3. Run demand model

martin-mosteiro:
Until the schedule maker tool gets created, I have now made this model an option that the user can toggle on or off by selecting the boolean `use-transportation-schedules` as suggested by @JIMENOFONSECA:
> There should be a "use-matsim-schedules" boolean parameter in config.demand that tells CEA to use those schedules instead of default ones in the next iteration.

daren-thomas:
Just did a quick review on this and... it includes some nice cleaning up of the existing code too. It would be a shame to waste the whole thing... I vote to not close this just yet until we figure out the best way to include these changes into the master branch. I can imagine merging most of the cleaned up code but extracting the actual MATSIM specific stuff into a small blog post. I think that would provide the most value from what we have.

@martin-mosteiro if I wrote such a post - would you mind fact-checking it?

martin-mosteiro:
@daren-thomas Definitely!

JIMENOFONSECA:
So we gave it a thought with @reyery @shizhongming @gabriel-happle and we will give it a go and include it into a new schedule maker tool. I wonder if you would like to wait until that to do the Blog Post? It will be a team effort and it could be out as soon as next week. @gabriel-happle and I start tomorrow. The interface can go for this sprint too.

The tool could as of now:

1. Have a dropdown with the next options (matsim schedules, SIA schedules)
2. Have a dropdown with the names of buildings to create the schedules for (default, all)
3. Have advanced options (two variables indicating where to locate the matsim files (there are two necessary). The matsim schedules will only run if these files are indicated.

What do you guys think?

daren-thomas:
@JIMENOFONSECA that can work too!

daren-thomas:
decision: assigning to daren who will add it to legacy or jupyter notebook

JIMENOFONSECA:
@daren-thomas this is a bit outdated. I wonder if we can just close this PR.

daren-thomas:
hmmm... it _does_ seem it won't be merged anytime soon. it's a pity, but maybe that's just how it is.

PR #2305, opened 2019-09-16
2279 show baseline network in input editor

### LINKED ISSUE #2279: Render supply networks in input editor based on their connectivity
Currently the input editor shows supply networks with nodes that are all connected to each other, when in fact the connectivity is supposed to be determined by the type of network indicated in the input database `supply-systems`.

reyery:
This PR resolves #2279.

The network layout shown in the input editor is now based on the type of network indicated in the input database `supply-systems`, similar to what is done in the `supply-systems-map` plot.

When starting the input editor, it will load the map first before trying to load the network layout (which can be seen on the top left corner _'Fetching ... network'_) as it might take some time.

Editing and saving the `supply-systems` database would also cause the map to fetch the network layout again.

Bugfix
-
- Fixed a bug where doing successive edits and saves on the same table will only save the first change.
  - Caused by the regeneration of the `data` property of the `inputstore` object after saving while the table is still using a previous pointer of the `data` property. Solved by recreating the table based on new `data` pointer.
https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/533cc14dd00787b2d6d487494772f6036118c017/cea/interfaces/dashboard/inputs/static/js/store.js#L141-L145


reyery:
From how it is implemented now, the call to create the network layout sometimes take quite a while and it blocks Flask from receive any other request until it has finished (e.g. trying to navigate away from the page). Maybe we need to move the calculation to a separate thread? @daren-thomas

daren-thomas:
I'm not sure how this is supposed to work...
I just tried this with the "CASE-EXECUTIVE-COURSE/baseline" scenario:

image

Easy! It seems I need to run Network Layout first, right? Nope. That image is from _after_ running network layout for both DH and DC...

daren-thomas:
Also, I'm not sure how the logic is supposed to happen from the description of the issue. @JIMENOFONSECA is there documentation somewhere on how the networks are defined in the inputs etc.?

JIMENOFONSECA:
@daren-thomas no documentation. No need to run network layout. This PR basically creates a network if the supply_system Th_sys or Tc_sys contains one type of system with district heating or cooling. i.e., T12, T14 etc...

The network layout tool we have (which creates some input files) is just wrong. It should be an output of a potential network. Maybe we change that in this PR?

reyery:
Testing
-
- Open input editor and edit one of the `supply-system` data to include supply systems that have a `scale` of **DISTRICT** (check `.\inputs\technology\lifecycle` for the supply system codes) and save.
  - The map will try to fetch the new network layout and there should be no network shown as it needs to have at least 2 connected buildings.
- Repeat the previous step by modifying another row.
  - The map will try to fetch the new network layout and should show 2 buildings that are connected.

daren-thomas:
wait... the supply system codes are defined in a file called "LCA_infrastructure.xlsx"? I can see now why jack & the other students are trying to push for a change here... is there a way to make this more intuitive?

(imagine you'd never used the cea before - how would you find the system codes?)

I created an issue for this: #2310

Also, @JIMENOFONSECA, regarding the network layout tool - I don't quite understand. How it wrong? Could you please create an issue explaining the problem and a possible solution?

cea/interfaces/dashboard/inputs/routes.py
```diff
@@ -76,6 +76,8 @@ def route_get_building_properties():
     import cea.plots
     import cea.glossary
 
+    div = request.args.get('div', default=False)
```
daren-thomas: this looks like you're doing a backward-compatibility thing for the electron interface. I suggest we move all-or-nothing. and use a separate route - so it's easier to delete the old stuff when we're done with the transition.

daren-thomas:
As far as I can tell, this can be merged.

PR #2319, opened 2019-09-18
Bugfix: Dashboard set scenario function not working

### LINKED ISSUE #2314: Dashboard "set-scenario" does not work in master
@reyery maybe we can check this? to replicate just press the 'set-scenario' button in the dashboard

image

reyery:
This PR fixes #2314.

Bugfix
-
- Fix wrong function parameter when using set scenario function in dashboard


daren-thomas:
Works. thank you :)

PR #2335, opened 2019-09-24
Bugfix: Edit selection in input editor does not work if some old values is equal to the new value

### LINKED ISSUE #2328: Edit Selection does not work if some rows have the same value as the new value
Hi,

I am trying to edit inputs for multiple buildings so I am using the edit selection button. The problem is: if at least one of the buildings already has one of the fields the same as what I am trying to edit, it doesn't let me press ok and apply the changes for the others.
I get that it doesn't make sense to edit a building to change to the same properties, but sometimes is the fastest way to edit multiple buildings (e.g. I want to change all buildings to retail (below) but because some of them are already as retail, I can't apply the changes).

image

Not sure if that was clear, please let me know if you have questions about it

reyery:
The PR resolves #2328 

This bug is caused by accessing a nested object property where one of the higher level properties do not exist, thus causing a `TypeError`. This error stops the loop and prevents the edit selection modal from closing after clicking **OK**.

### DIFF
--- a/cea/interfaces/dashboard/inputs/static/js/store.js
+++ b/cea/interfaces/dashboard/inputs/static/js/store.js
@@ -54,9 +54,11 @@ class InputStore {
         if (method === 'update') {
             //Check if update is the same as default
             if (this.tables[table][building][column] === value) {
-                delete this.changes[method][table][building][column];
-                if (!Object.keys(this.changes[method][table][building]).length) {
-                    delete this.changes[method][table][building];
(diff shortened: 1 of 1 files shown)


reyery:
Testing
-
- Select multiple rows in the table and click edit selection
- Change one of the column values to something that is the same as the value of one of the rows 
(e.g. one of the rows has `MULTI_RES` value of 1, enter 1 to `MULTI_RES` input and click **OK**)
  - The edit selection modal should close and the selected rows have the new value

Issue #2270, opened 2019-08-21
network-layout tool: UnboundLocalError: local variable 'field' referenced before assignment


daren-thomas:
This is an issue posted by @kerembeygo (transferred from #2259)

![image.png](https://images.zenhubusercontent.com/57b40957e40e5714b16d7331/6f1fbe32-d268-478d-b9fa-85df012d21d4)

![image.png](https://images.zenhubusercontent.com/57b40957e40e5714b16d7331/0a17c5df-3025-4966-9e1a-45b6defc02bf)

JIMENOFONSECA:
I can check this...

JIMENOFONSECA:
EL should not be part of the inputs. We do not support that functionality for now. PR #2337 solves this issue

PR #2337, opened 2019-09-24
fix to network layout script (EL is not implemented yet)

JIMENOFONSECA:
This PR solves #2270 in fact other networks besides DH and DC are not fully supported for now

### DIFF
--- a/cea/technologies/network_layout/main.py
+++ b/cea/technologies/network_layout/main.py
@@ -53,9 +53,10 @@ def layout_network(network_layout, locator, plant_building_names=[], input_path_
     output_nodes = locator.get_network_layout_nodes_shapefile(type_network, output_name_network)
     output_network_folder = locator.get_input_network_folder(type_network, output_name_network)
 
-    if connected_buildings != []: #if not blank.
+    if connected_buildings:
         disconnected_building_names = [x for x in connected_buildings if x not in total_demand_location['Name']]
     else:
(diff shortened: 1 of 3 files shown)


cea/technologies/network_layout/main.py
```diff
@@ -53,11 +53,14 @@ def layout_network(network_layout, locator, plant_building_names=[], input_path_
     output_nodes = locator.get_network_layout_nodes_shapefile(type_network, output_name_network)
     output_network_folder = locator.get_input_network_folder(type_network, output_name_network)
 
+    if connected_buildings != []: #if not blank.
```
daren-thomas: the pythonic way to write this is actually `if connected_buildings`.

cea/technologies/network_layout/steiner_spanning_tree.py
```diff
@@ -253,6 +253,9 @@ def calc_coord_anchor(total_demand_location, nodes_df, type_network):
         field = "QH_sys_MWhyr"
     elif type_network == "DC":
         field = "QC_sys_MWhyr"
+    elif type_network == "EL":
```
daren-thomas: else? there should be no `elif` clause without an `else` clause... mainly because it makes me super nervous. defensive programming would suggest to raise a `ValueError` or something with the text `"Invalid value for type_network ({type_network})".format(**locals())`

Issue #1974, opened 2019-05-17
What happened to the concept module?


daren-thomas:
It seems it's referenced in `cea/optimization/flexibility_model/mpc_district/planning_and_operation_batch_run.py`. It's probably referenced in other places as well.

Was it removed?

JIMENOFONSECA:
if it si not there, then it is removed...

daren-thomas:
@JIMENOFONSECA, maybe let me refrain the question a bit: What is the rationale? Why was it removed? Whoever removed it, didn't do a proper search / replace.

![image](https://user-images.githubusercontent.com/2969564/58234652-52f96b00-7d3f-11e9-8313-f4e90437281b.png)

This code is now non-functional. So I'd really like to know who moved it, where to, and what to do about the stuff that's now broken. Do we delete it?

JIMENOFONSECA:
yep, I believe it is not part of the tools MPC-district and MPC-building anymore, so we can delete it…

_______________________

Dr. Jimeno A. Fonseca
Senior researcher and project coordinator
Future cities Laboratory
Singapore ETH-Centre

Chair of Architecure and Building Systems
ETH Zurich

On 23 May 2019, 3:47 PM +0800, Daren Thomas <notifications@github.com>, wrote:

@JIMENOFONSECA<https://github.com/JIMENOFONSECA>, maybe let me refrain the question a bit: What is the rationale? Why was it removed? Whoever removed it, didn't do a proper search / replace.

[image]<https://user-images.githubusercontent.com/2969564/58234652-52f96b00-7d3f-11e9-8313-f4e90437281b.png>

This code is now non-functional. So I'd really like to know who moved it, where to, and what to do about the stuff that's now broken. Do we delete it?

daren-thomas:
ok. assigning to you - you are probably the most knowledgeable person on the team regarding this matter. we'll prioritize it next week.

PR #2338, opened 2019-09-24
Moving flexibility (concept project) to legacy

JIMENOFONSECA:
Since this feature will not be used anymore (due to drastic changes in the optimization) and lack of people to maintain it, we move it to legacy

this PR solves issues #1974


daren-thomas:
@JIMENOFONSECA I'm going to merge this as-is, but would like to take this chance to point out something I figured out the other day: Instead of having a legacy folder for old, rotting code that will never ever work again, because the environment it was embedded in has changed, we could do this instead:

Use tags. On GitHub, we can tag versions (think: snapshots) of the repository. We do that for releases already.

I propose, instead moving something to legacy in the future, we do this instead:

- find the last revision that the code actually worked
- create a tag (in this case: "legacy-flexibility-concept-project")
- add some documentation (a `legacy.rst` in the docs?) that specifies
  - _ what_ the code was about
  - _why_ it was moved to legacy

What do you think?

PR #2353, opened 2019-10-01
Project api for electron application

reyery:
This PR add additional functions to the Flask server to enable the use of the upcoming electron version of the CEA.

The additions mainly include endpoints to create and manage project settings.

The pre-release version of the electron app can be downloaded here for testing. 
https://we.tl/t-5H3yJKhY08

### DIFF
--- a/cea/interfaces/dashboard/api/project.py
+++ b/cea/interfaces/dashboard/api/project.py
@@ -1,11 +1,12 @@
 import os
+import shutil
 
 import geopandas
+from flask import current_app
 from flask_restplus import Namespace, Resource, fields, abort
-import shutil
 from staticmap import StaticMap, Polygon
 
-import cea.config
+# import cea.config
 import cea.inputlocator
 from cea.utilities.standardize_coordinates import get_geographic_coordinate_system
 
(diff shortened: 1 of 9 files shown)


daren-thomas:
@reyery what exactly am I supposed to be testing here?

reyery:
@daren-thomas mainly project management i.e. creating new project/scenario (especially the new scenario form), editing/deleting scenarios.

daren-thomas:
I'm able to create new projects and scenarios. I think it's good as-is.

daren-thomas:
Regarding the electron interface: if we add the link to the manual and issues (the documentation part), then we'd be good to go, no?

daren-thomas:
I also tested with the browser dashboard. That still works. (it only has to work until we have an electron installer for windows, after that we can stop supporting it i guess)

cea/interfaces/dashboard/api/project.py
```diff
@@ -20,18 +28,30 @@
...
+        if not os.path.exists(config.scenario):
+            config.scenario_name = ''
+            config.save()
+        return {'name': name, 'path': config.project, 'scenario': config.scenario_name,
+                'scenarios': config.get_parameter('general:scenario-name')._choices}
 
     @api.doc(body=PROJECT_PATH_MODEL, responses={200: 'Success', 400: 'Invalid Path given'})
-    def post(self):
-        config = cea.config.Configuration()
+    def put(self):
```
daren-thomas: are you sure this isn't the wrong way round? `PUT` should be used to create a new one, `POST` to update an existing project?

reyery: yeah, `POST` is to **create** and `PUT` is to **update/replace**

daren-thomas: @reyery you're right! i had this wrong in my head all these years! thank you :)

Issue #2204, opened 2019-07-24
Schedule of temperature set-points for CEA


martin-mosteiro:
The heating and cooling set points are controlled as follows:
```
def get_heating_system_set_point(people, t, bpr, weekday):
    """
    :param people:
    :param t: hour of the year, simulation time step [0...HOURS_IN_YEAR]
    :type t: int
    :param bpr: BuildingPropertiesRow
    :type bpr: cea.demand.building_properties.BuildingPropertiesRow
    :param weekday:
    :return: heating system set point temperature [°C]
    :rtype: double
    """

    if is_heating_season(t, bpr):

        if people == 0:
            if 5 <= weekday <= 6:  # system is off on the weekend
                return np.nan  # huge so the system will be off
            else:
                return bpr.comfort['Ths_setb_C']
        else:
            return bpr.comfort['Ths_set_C']
    else:
        return np.nan  # huge so the system will be off
```

This means that if a building is occupied 7 days a week, the heating/cooling systems will be turned off when people aren't there on Saturdays and Sundays. This doesn't seem very realistic – why would a building that is occupied 7 days a week treat a Sunday differently from a Monday?

This leads to results like these:
<img width="940" alt="Screen Shot 2019-07-24 at 11 46 14" src="https://user-images.githubusercontent.com/18548065/61783889-f0891d80-ae08-11e9-9161-d837e98b5767.png">
(the golden line is measured data, the other lines are different simulations using CEA)

Me and @gabriel-happle discussed this and we think whether a system is on or not on weekends should be an input parameter – either a simple input (something like `type_cs`) or a temperature setpoint schedule – I believe @BhargavaKrishnaSreepathi worked on this at one point.

gabriel-happle:
I think temperature set-point schedules is a good idea. We should add it to the schedules file so that the user has more control.

Actually, I'm reviewing a paper on this topic right now.
It states ISO 13790 standard as source and gives a schedule like this for residential:
07:00 - 17:00  :  16°C
17:00 - 23:00  :  21°C
23:00 - 07:00  :  18°C

If we want to allow this kind of input, we should define 24-h schedules with temperatures directly.
Not only set-point and set-back times.

JIMENOFONSECA:
just include a new variable in the inputs/indoor-comfort, saying whether the weekend operation is allowed or not and then update the code accordingly....

daren-thomas:
[...]

PR #2363, opened 2019-10-07
I2204 temperature setpoints

gabriel-happle:
this PR contains the following changes:
- Temperature set point / set back schedules are moved to the occupancy database. These schedules consist of temperature values or `OFF`.
- Temperatures are therefore removed from the `comfort.dbf`
- The occupancy model adds these temperature schedules to the schedule files according to the main use of the buildings.
- These schedules are directly used to control heating and cooling systems. The only other parameter controlling heating and cooling operation is the heating season and cooling season.

How to test:
- check if all schedules in the database have temperature values
- run the data-helper
- run the demand (for different cases)
- check the schedules output files
- check the demand output files


JIMENOFONSECA:
@martin-mosteiro you included a way to calculate the main use of a building right? I remember we had one. Maybe @gabriel-happle could use that one instead of creating a new input variable in CEA?

It just avoids unintended behavior. Users, forgetting to run datahelper and changing the main use.
What do you think?

check line 203 of the data_helper.py

image

gabriel-happle:
@JIMENOFONSECA I am using that one. The only thing I did was to save it. It was not saved before, so scripts executed after the data-helper did not have access to it.

martin-mosteiro:
1. What is the source for the system schedules? The new schedules change the system behavior compared to the old code. In the new version, office building's heating systems are off at night on weekdays.
2. I'm not sure this is a good idea:
> Temperatures are therefore removed from the comfort.dbf

I think this makes it a little bit harder for users to adjust setpoint and setback temperatures, which the model is quite sensitive to. (If two office buildings have different set point temperatures, you have to adjust the schedule produced in CEA). Maybe the schedules should not be "SETPOINT"/"SETBACK"/"OFF" instead of "21"/"18"/"OFF"?
Screen Shot 2019-10-07 at 11 01 30
Or is that too contrived?

JIMENOFONSECA:
so as for

1. We thought of a more sensible way to set the control of the system. Changes are welcome. We will have to do this for now base on technical assumptions.

2. This will become more easy to do now that we create an interface in CEA to deal with schedules building by building and district by district. This is coming up as an update to this PR soon. The user will not interact with this excel file anymore, but with a schedule defined at the building level. The excel files are being converted into CSV files so we are able to track changes more easily.

As of 1, maybe we can discuss schedule by schedule how it should operate?


_______________________

Dr. Jimeno A. Fonseca
Senior researcher and project coordinator
Future cities Laboratory
Singapore ETH-Centre

Chair of Architecure and Building Systems
ETH Zurich

On 7 Oct 2019, 5:24 PM +0800, Martín Mosteiro-Romero <notifications@github.com>, wrote:

  1.  What is the source for the system schedules? The new schedules change the system behavior compared to the old code. In the new version, office building's heating systems are off at night on weekdays.
  2.  I'm not sure this is a good idea:

Temperatures are therefore removed from the comfort.dbf

I think this makes it a little bit harder for users to adjust setpoint and setback temperatures, which the model is quite sensitive to. (If two office buildings have different set point temperatures, you have to adjust the schedule produced in CEA). Maybe the schedules should not be "SETPOINT"/"SETBACK"/"OFF" instead of "21"/"18"/"OFF"?
[Screen Shot 2019-10-07 at 11 01 30]<https://user-images.githubusercontent.com/18548065/66298437-08dde400-e8f2-11e9-99e0-f8c445d62550.png>
Or is that too contrived?

gabriel-happle:
@martin-mosteiro is it sufficient to have only two temperatures?
In the related issue I mention a schedule with 3 temperatures:
https://github.com/architecture-building-systems/cityenergyanalyst/issues/2204#issuecomment-514574299

What do you think about that? Have you seen this before?

gabriel-happle:
@martin-mosteiro your point 2 is a good one!
I didn't think of that.
You are right, it is not easy to control the setpoint of different buildings of the same use.

martin-mosteiro:
@JIMENOFONSECA 
> 1. We thought of a more sensible way to set the control of the system. Changes are welcome. We will have to do this for now base on technical assumptions.

Maybe I could've been more clear: buildings such as offices and retail now have no setback temperatures and are only conditioned during working hours. Is this common practice?

JIMENOFONSECA:
so this is one example of EPlus:

https://designbuilder.co.uk/helpv1/Content/Schedules_Defined_Using_Profiles.htm

we can determine the best practices we decide for the context. Still, these databases are just databases. We are creating a schedule-helper tool, which will translate these databases into schedules per building that can be edited via the CEA dashboard interface

JIMENOFONSECA:
@gabriel-happle yes, what I mean is that we do not need to save it. We can recalculate it as needed. I know it adds some extra weight in the background. But it helps to not create anymore input variables in CEA (those that appear in the input databases)

JIMENOFONSECA:
@gabriel-happle @martin-mosteiro another alternative (this is what Eplus does) is that we create setpoint schedules in the same fashion as we do for occupancy. They could be normalized between 0 and 1. 

Then we will need to indicate one temperature setpoint. The schedule will then multiply to the temperature setpoint every hour. We will need to find the right number that gives use the 18 degrees of setback temperature etc.. 

But again, our plan is to make the data helper create schedules per building basis. THe CEA interface will allow us to easily modify schedules of groups of buildings or just one building. That will solve your problem number 2 right martin? @martin-mosteiro

gabriel-happle:
> 
> 
> @martin-mosteiro you included a way to calculate the main use of a building right? I remember we had one. Maybe @gabriel-happle could use that one instead of creating a new input variable in CEA?
> 
> It just avoids unintended behavior. Users, forgetting to run datahelper and changing the main use.
> What do you think?
> 
> check line 203 of the data_helper.py
> 
> image

this is changed now.

gabriel-happle:
> 
> 
>     1. What is the source for the system schedules? The new schedules change the system behavior compared to the old code. In the new version, office building's heating systems are off at night on weekdays.
> 
>     2. I'm not sure this is a good idea:
> 
> 
> > Temperatures are therefore removed from the comfort.dbf
> 
> I think this makes it a little bit harder for users to adjust setpoint and setback temperatures, which the model is quite sensitive to. (If, for example, different office buildings have different set point temperatures, you would have to adjust each schedule produced in CEA). Maybe the schedules should not be "SETPOINT"/"SETBACK"/"OFF" instead of "21"/"18"/"OFF"?
> Screen Shot 2019-10-07 at 11 01 30
> Or is that too contrived?

@martin-mosteiro this is implemented now!

martin-mosteiro:
@JIMENOFONSECA a quick question about the system schedules for certain occupancy types (e.g., OFFICE): in the old implementation in CEA, OFFICE buildings had a set point and set back temperatures, and additionally the systems were turned off during the weekend. In the new versions, systems are simply off when there are no occupants (no setback temperature). Is this desired? Is this more realistic system operation?

JIMENOFONSECA:
@martin-mosteiro, I just got what you meant. Well nowadays, many prefer to reduce the setpoint temperature over the weekend and at night (but not turning the equipment off) so we avoid the early peak on monday morning. 

Many others cool down the building a bit earlier in the morning one hour at least before the building gets occupied so people are comfortable when they arrive. I think this depends on the case, and what is there is an assumption. 

We can take another one. Actually it would be great if we can ask in Zurich how they do it in some offices buildings. I imagine even they operate on Saturdays...

We implemented the version you mentioned. with setbacks. We believe it works better. let me know your opinion.

Maybe we can open an issue to update these 'control' considerations for CH and Singapore?

PR #2372, opened 2019-10-08
2360 create installer for electron interface

daren-thomas:
This PR creates an installer for the CEA Electron interface and updates the notes on creating a new release to include providing the CEA Electron interface - which, for now, lives in its own repository and will be developed in lock-step with the rest of the CEA code.

A test-version of the installer can be found here: https://www.dropbox.com/s/nukficzmvq64hyf/Setup_CityEnergyAnalyst_2.23.exe?dl=0

After installation, use the "CEA Dashboard" link from the start menu to open the Electron interface.

### DIFF
--- a/cea/interfaces/dashboard/dashboard.py
+++ b/cea/interfaces/dashboard/dashboard.py
@@ -141,12 +141,10 @@ def main(config):
     # the protocol for the Connection messages is tuples ('stdout'|'stderr', str)
     app.workers = {}  # script-name -> (Process, Connection)
 
-    print("starting webbrowser timer")
-    threading.Timer(0.5, lambda: webbrowser.open('http://localhost:5050')).start()
-    print("started webbrowser timer")
     print("start socketio.run")
     socketio.run(app, host='localhost', port=5050)
     print("done socketio.run")
+
 
 if __name__ == '__main__':
     main(cea.config.Configuration())
--- a/docs/how-to-create-a-new-release.rst
+++ b/docs/how-to-create-a-new-release.rst
@@ -160,11 +160,26 @@ Uploading to PyPI
 
 .. _twine: https://pypi.python.org/pypi/twine
 
+Updating the CEA Electron interface
+-----------------------------------
+
+For the installer to be able to pick up the newest version of the CEA Electron interface, you need to build it,
+create a release and attach a ``win-unpacked.7z`` to the release. Here are the steps:
+
+- pull the newest version of the ``cea-electron`` repository
+- open CEA Console, navigate to the GitHub repo of the ``cea-electron`` repository
+- type ``yarn dist:dir``, wait for the command to complete
+- the subfolder ``dist`` will contain a folder ``win-unpacked``
+- compress the folder ``win-unpacked`` to ``win-unpacked.7z``
+- open https://github.com/architecture-building-systems/cea-electron/releases in the browser
+- Draft a new release, use a tag corresponding to the CEA version number (e.g. ``v2.23``)
+- Attach the ``win-unpacked.7z`` file to the release (the installer will automatically use the latest release on installation)
+
 
 Updating Link in www.cityenergyanalyst.com/tryit
 --------------------------------------------------
 
-- go to www.cityenergyanalyst.com
+- go to http://www.cityenergyanalyst.com
 - press Esc and try logging into squarespace (the credentials are here_)
 - go to Pages/Try CEA  (it is the last page in the list)
 - go to edit 'Page content'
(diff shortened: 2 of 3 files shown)


cea/interfaces/dashboard/dashboard.py
```diff
@@ -141,12 +141,10 @@ def join_paths(path_list, loop_index):
     # the protocol for the Connection messages is tuples ('stdout'|'stderr', str)
     app.workers = {}  # script-name -> (Process, Connection)
 
-    print("starting webbrowser timer")
-    threading.Timer(0.5, lambda: webbrowser.open('http://localhost:5050')).start()
-    print("started webbrowser timer")
     print("start socketio.run")
     socketio.run(app, host='localhost', port=5050)
     print("done socketio.run")
```
reyery: Does the 'done' here mean `socketio.run` "has started running" or "has closed / stopped running", as it will not be printed until `socketio.run` ends since it's blocking

docs/how-to-create-a-new-release.rst
```diff
@@ -160,11 +160,26 @@ Uploading to PyPI
...
+For the installer to be able to pick up the newest version of the CEA Electron interface, you need to build it,
+create a release and attach a ``win-unpacked.7z`` to the release. Here are the steps:
+
+- pull the newest version of the ``cea-electron`` repository
+- open CEA Console, navigate to the GitHub repo of the ``cea-electron`` repository
+- type ``yarn dist:dir``, wait for the command to complete
+- the subfolder ``dist`` will contain a folder ``win-unpacked``
+- compress the folder ``win-unpacked`` to ``win-unpacked.7z``
+- open https://github.com/architecture-building-systems/cea-electron/releases in the browser
+- Draft a new release, use a tag corresponding to the CEA version number (e.g. ``v2.23``)
```
reyery: Hmmm. I feel that the version number for the interface should not follow CEA since it would not make sense if the minor version suddenly jumps if it has not changed for a while compared to CEA, of course that would be more troublesome. I dont know, what do you think?

reyery:
I don't think we can merge this first right? Since it would mean that the browser will not be launched when in master.

cea/interfaces/dashboard/dashboard.py
daren-thomas: `socketio.run` starts the Flask webserver. So yes, it's only printed when the application ends.

docs/how-to-create-a-new-release.rst
daren-thomas: I think the version numbers correspond to the milestones / sprints. `cea-electron` doesn't have it's own release cycle, right? In my mind, it's closely linked to the CEA lifecycle (the cea-electron repository can't really be used outside of the CEA context at all anyway). So, I don't really see much of a problem.

If it really bothers you, can you recommend a way to name the tags, so that they're not cea-electron versions, but denote that they _refer_ to a CEA release?

I really think it's easier if the releases of cea-electron are based on the CEA release cycle as the two are so closely linked. _theoretically_ the situation could arise where `cea-electron` is not updated during a sprint. _practically_, I think the opposite is more likely given the current priorities.

daren-thomas:
@reyery if we merge this now, then all is good: The last installer uses a version that loads the web browser and after merging, the next installer will use the cea-electron interface, effectively deprecating the original reactless interface.

docs/how-to-create-a-new-release.rst
JIMENOFONSECA: In my opinion, the releases of cea-electron (now called CityEnergyAnalyst-dashboard) can be separate. The reason is that everytime an update of the interface exists (after solving a bug) we need to make a build and publish it. At least that is how it works now.

We are facing this now with Reynold, where I checked this PR, found several bugs that needed to be fixed and tested. But to do that. We need to make another build, I have to download the build, replace files and test. that is just a long process.

Ideally, we create a way to automatically update the interface once a new build is available. Still we will need that build to be labeled somewhere.

Maybe I am missing something? Mind a chat so we make this clear? It would save some time of bouncing the ball back..

JIMENOFONSECA:
@daren-thomas @reyery this is not ready to be merged. I already asked directly to reyeyery about some bugs/clitches that need to be solved before merging. One of those is the "shortcut" not working, and the automatic start of the old dashboard.

docs/how-to-create-a-new-release.rst
daren-thomas: I'm on vacation all next week, so a chat won't be possible until after that. Here some thoughts:

- CityEnergyAnalyst-dashboard... we missed a chance to re-brand what we call the whole interface as opposed to the plotting stuff. How about CityEnergyAnalyst-GUI?
- The CEA installer picks up the _latest_ published release of the CityEnergyAnalyst-dashboard from GitHub. So you can easily publish bugfixes by doing `v2.24.1`, `v2.24.2` etc.
- Even during bug fixing, it might be prudent to keep official releases on the same release schedule as the CEA (bi-weekly) - unless a hotfix is necessary. This creates some stability.
- an auto-update feature for the interface should be an issue: this is a great idea!
- regarding testing: that should work similar to how the development process of CEA works: `git pull` on the `CityEnergyAnalyst-dashboard` repo, type `yarn dev` to start the electron interface. Am I missing something?

daren-thomas:
@JIMENOFONSECA regarding the automatic start of the old dashboard: We can probably easily fix that by adding a parameter to `cea dashboard --auto-open-browse` defaulting to  `on` and starting it from electron with the `off` setting.

I was hoping, by publishing the `CityEnergyAnalyst-dashboard` interface, we can remove the old dashboard. That will make maintenance a lot easier! And we can go ahead then and delete a lot of old code (including all those javascript libraries currently in the CityEnergyAnalyst repo!)

PR #2374, opened 2019-10-09
2356 Update environment.ubuntu.yml

strusoftsawen:
Some of the dependencies of Ubuntu CEA have been changed to a `broken` state at `conda-forge`. This change means CEA can be installed, but the dependencies should really be updated.

Additionally, `flask` requires `jinja2=2.10.1` to work. I have not checked what the implications of this are.

### DIFF
--- a/environment.ubuntu.yml
+++ b/environment.ubuntu.yml
@@ -4,6 +4,8 @@ channels:
   - dlr-sc
   - https://mirrors.tuna.tsinghua.edu.cn/anaconda/pkgs/main
   - conda-forge
+  - conda-forge/label/gcc7
+  - conda-forge/label/broken
   - defaults
 dependencies:
   - attrs=19.1.0=py_0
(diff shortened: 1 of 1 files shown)


daren-thomas:

JIMENOFONSECA:
@daren-thomas all good, but you might want to verify that the environment is up to date with the environment of windows (i.e., libraries it requires)

daren-thomas:
@strusoftsawen thank you for this!

In order to add you to the CREDITS.md file (Contributors section), would you prefer to use your full name (specify) or your GitHub handle?

strusoftsawen:
@daren-thomas Great, glad I could help. Here's just a reminder that using the `broken` dependencies is probably a really bad idea and should be replaced when there are more resources for ubuntu development. Happy to help out with that if I get to do more testing in the ubuntu environment.

Feel free to use my full name for the credits: Toivo Säwén.

Cheers,
Toivo

PR #2378, opened 2019-10-10
I2204b temperature setpoints finish

martin-mosteiro:
PR #2363 was merged before I had a chance to make a (minor) change.

I have adjusted the heating system schedules so they match the previous behavior in CEA, i.e., in non-residential buildings systems (i) operate at the set-point temperature when people are present, (ii) operate at the setback temperature when people are out on days where the building is occupied (i.e., on weekdays and possibly Saturdays), and (iii) do not operate at all on days when the building is unoccupied (i.e., Sundays and possibly Saturdays).

Given that the assumption that systems are turned off on weekends is already optimistic based on my measured data, it seems reasonable to assume that not to be the case on weekdays.

I was originally planning to do a little deep dive into the subject for this PR, but due to the lack of time I was not able to. For now, I'm keeping the heating systems as was done up until now in CEA and might do the deep dive later on.

The cooling system schedules are unchanged from PR#2363 since it does seem like a reasonable assumption that systems are simply turned off when the building is unoccupied.


gabriel-happle:
The schedules are changed exactly as @martin-mosteiro described.
As he is an expert in heating schedules, I suggest merging.

PR #2382, opened 2019-10-14
New Schedules Maker / Model

JIMENOFONSECA:
This PR creates a new element in the workflow of CEA called the occupancy model.

We did the next changes:
1. We changed the occupancy_archetypes.xls to csv files. This allows us to track changes more easiliy.
2. We included one more option to the data helper. the user can now create occupancy schedules per building. An interface is being prepared by @reyery for this.
3. We included a new TOOL called "Building_schedules" This tool creates either a deterministic or stochaistic profile of occupancy and internal loads.

Now, before running the energy demand, you need to run the schedule-naker model first. 
the results are stored in outputs/data/occupancy.

The energy demand script uses these schedule_maker model files as input now.

to test:
1. run data helper with the new default configuration. This configuration includes an option to create archetypical daily schedules for every building (these are inputs of the schedule maker)
2. run the Building schedules tool (it is in the interface) under demand forecasting.
3. run the energy demand too. 

To verify, all should be passing the tests.


JIMENOFONSECA:
@martin-mosteiro I would need your help to match the unit tests for the schedules. I find strong difficulty on understanding the tests.  Would you have anytime soon for this?

martin-mosteiro:
Just started. I'm not sure I understand the purpose of `TestSavingLoadingSchedules`, which just seems to run the model twice and check the results are the same for both. git appears to blame @daren-thomas, maybe he has some ideas about this?

martin-mosteiro:
@JIMENOFONSECA I have updated the schedule tests and now they run correctly again.

JIMENOFONSECA:
@martin-mosteiro should we merge?

martin-mosteiro:
I suggest @gabriel-happle reviews this first. This is quite central for his work too so it's important that he agrees with the changes.

JIMENOFONSECA:
@gabriel-happle mind letting us know if this ok to go from your side?

JIMENOFONSECA:
In the data-helper a list of buildings can be provided as advanced input. What is the idea here?
—so the idea is to make the data helper eventually work in individual buildings…so the user can replace only one input if required. Now it works only for the schedules. You can create the schedule for a selection of buildings or for everyone.
—do you think we should not have this functionality yet?




On 16 Oct 2019, 11:31 AM +0800, architecture-building-systems/CityEnergyAnalyst <reply@reply.github.com>, wrote:

In the data-helper a list of buildings can be provided as advanced input. What is the idea here?

gabriel-happle:
> 
> In the data-helper a list of buildings can be provided as advanced input. What is the idea here? —so the idea is to make the data helper eventually work in individual buildings…so the user can replace only one input if required. Now it works only for the schedules. You can create the schedule for a selection of buildings or for everyone. —do you think we should not have this functionality yet? On 16 Oct 2019, 11:31 AM +0800, architecture-building-systems/CityEnergyAnalyst <reply@reply.github.com>, wrote: In the data-helper a list of buildings can be provided as advanced input. What is the idea here?

OK @JIMENOFONSECA, got it. I think that's fine.

martin-mosteiro:
Not to be a stickler, but one point about the units changed in this PR:
- _pax_ is not a very common abbreviation for "persons", apparently it's only common [informally in the hospitality industry](https://en.wiktionary.org/wiki/pax) (_1. (informal, usually in the plural) passenger; passengers; 2. (informal, usually in the plural, by extension, hospitality industry) guest (at an event or function)_)
- the unit _lpd_ stands for **l**iters per **p**erson per **d**ay, so _lpdpax_ sounds redundant to me (I guess it could be interpreted as **l**iters **p**er **d**ay per **pax**...?). same with _lpspax_.

Issue #2352, opened 2019-09-30
void decks (odd results)


lguilhermers:
I was discussing some results with @JIMENOFONSECA and showed him some results with simulations of void decks.

I added 0.05 (5%) of void decks, which should increase ventilation and reduce the energy consumption. Furthermore, as I understood, it is reducing the area of apartments in this building (I imagine as the ground floor of a condominium or any building with pilotis for example.

By this logic, the cooling load/ electricity consumption should decrease, but instead it is slightly increasing. Am I missing any point here in understanding this parameter?

PR #2384, opened 2019-10-15
fix void-deck issues

JIMENOFONSECA:
This fixes issues with the void-deck

We redefine what void-deck parameter means. It is the number of floors from the ground up that are void in the building.

issue #2352 

To test,
run data helper, 
run demand,


JIMENOFONSECA:
@lguilhermers does this work?

lguilhermers:
@JIMENOFONSECA 

A few points on that, based on tests I did:
1) there is no check to put a number of void decks higher than the number of floors. I was even able to simulate the buildings in this case (not sure how CEA interprets that).
2) both the EUI and total energy supply is basically the same when using 0, 2 or 10 for void decks in a building with 20 floors, so not sure if the void decks are acting accordingly.
3) the equivalent simulation in master branch caused a small increase in EUI, so this point seems to be fixed by this branch (void decks no longer increase energy consumption).

Issue #816, opened 2017-09-26
Reduce support for bayesian calibration tool of individual building


JIMENOFONSECA:

JIMENOFONSECA:
This is blocked by a publication submitted to IBIPSA 2019. should it be accepted (January 2019) we publish the tool

JIMENOFONSECA:
@adam-rysanek any news on the publication?

JIMENOFONSECA:
We will discontinue support on this tool and move it to the legacy folder. It is just consuming unnecessary resources at the moment. Idealy we implement a calibration tool based on GA (which could be significantly faster and appliacable to urban scale problems. Bayesian calibration is an overkill in here. However  I am a fan of using the approch for research purposes.

PR #2390, opened 2019-10-16
I816 stop bayesian calibration support

JIMENOFONSECA:
This solves #816.

It moves calibration to the legacy folder. More in the issue's comments @martin-mosteiro unless you need this, we end support right now. let me know


martin-mosteiro:
Well, in my own experience the tool never worked all the way through on my computer, so as far as I can tell it isn't really usable at its current status. Since nobody is working on this at the moment, it makes sense to move it to legacy.

Issue #2399, opened 2019-10-22
Energy Demand calculation failing with KeyError: 'rhum_max_pc'


daren-thomas:
When running `cea test`, the pydoit task `run_demand:open` fails with the following output:

```
TaskError - taskid:run_demand:open
PythonAction Error
Traceback (most recent call last):
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\doit\action.py", line 383, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\tests\dodo.py", line 200, in run_demand
    cea.demand.demand_main.main(config)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\demand\demand_main.py", line 158, in main
    demand_calculation(locator=locator, config=config)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\demand\demand_main.py", line 129, in demand_calculation
    repeat(debug, n))
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\utilities\parallel.py", line 95, in wrapper
    result = map_result.get()
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\multiprocessing\pool.py", line 572, in get
    raise self._value
KeyError: 'rhum_max_pc'
```

I'm guessing there is some step that needs to be added to the tests.

daren-thomas:
Nope. Just checked: Running the data-helper and the schedule-maker does not solve the problem.

daren-thomas:
Running with `multiprocessing=False` yields a more accurate description:

```
Traceback (most recent call last):
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\worker.py", line 146, in worker
    run_job(config, job, server)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\worker.py", line 108, in run_job
    script(config=config, **parameters)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\api.py", line 52, in __call__
    self._runner.__call__(*args, **kwargs)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\api.py", line 32, in script_runner
    script_module.main(config)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\demand\demand_main.py", line 158, in main
[...]

PR #2401, opened 2019-10-22
Rename calls to inexistent variable `RH_max_pc`

martin-mosteiro:
Someone selectively renamed `rhum_max_pc` and `rhum_min_pc` to `RH_max_pc` and `RH_min_pc` but didn't rename every call to that variable and didn't rename the variables in the databases, so *thermal_loads.py* was broken.

I have reverted both variables to the old naming convention since it's not clear to me that this was an intentional change. Now thermal_loads works correctly again.


martin-mosteiro:
Just noticed @daren-thomas had a different pull request (#2399) for this, so I'm closing this one.

Issue #2408, opened 2019-10-24
Shutting down does not kill all python processes


JIMENOFONSECA:
1. run cea-GUI
2. run demand tool.
3. close cea after the tool starts
4. try to relaunch cea

It does not work. Some processes remain in the background of CEA and CEA cannot open again.

![image](https://user-images.githubusercontent.com/8973186/67478081-f56e9f00-f68d-11e9-8467-211d9ec07f8e.png)

daren-thomas:
I'm able to reproduce this. A quick googling shows some ideas here: https://stackoverflow.com/questions/23434842/python-how-to-kill-child-processes-when-parent-dies/23587108

Since we know _when_ the process is shutting down (`/server/shutdown`) I think it could be as easy as collecting those `POpen` objects and `terminate()`ing them. I'll give it a shot.

PR #2414, opened 2019-10-25
Terminate running cea worker processes when closing the GUI

daren-thomas:
This PR solves #2408 - when closing the CEA GUI, running cea worker processes continued running. With the PR, they are terminated on shutdown of the cea flask server.

To test:

- run the CEA GUI
- open project / scenario
- run a bunch of long-running processes (like `demand` or `thermal-network`)
- observe them in the Task Manager (look for python processes)
- close the CEA GUI
- note the python processes are terminated

### DIFF
--- a/cea/interfaces/dashboard/server/__init__.py
+++ b/cea/interfaces/dashboard/server/__init__.py
@@ -9,6 +9,7 @@ from flask import Blueprint, current_app
 from flask_restplus import Api, Resource
 from .jobs import api as jobs
 from .streams import api as streams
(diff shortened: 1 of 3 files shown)


JIMENOFONSECA:
@daren-thomas  It does not seem to work. CEA remains in the background process. 

It happens when one closes CEA while the demand script is running in several cores. 

image

daren-thomas:
@JIMENOFONSECA I was able to reproduce the behavior. You'll need to `pip install -e .` to get the `psutil` library which I then use to recursively kill all sub-processes of the cea workers.

Issue #2397, opened 2019-10-21
Radiation results do not account for adjacent buildings


martin-mosteiro:
I am running a case study that includes the building shown in yellow below:
![image](https://user-images.githubusercontent.com/18548065/67223640-ec809200-f42f-11e9-8c4d-ebf614dcce7a.png)
![image](https://user-images.githubusercontent.com/18548065/67223644-f0141900-f42f-11e9-9416-cb4c0a706ca2.png)

As you can see, the building is surrounded on three sides by a taller building. This means that three out of the four sides of the building should have an irradiation of 0. However, the results show the following:
![Screen Shot 2019-10-21 at 18 25 16](https://user-images.githubusercontent.com/18548065/67223763-2782c580-f430-11e9-961c-a408493f36a2.png)

Sure, the western facade (the only one that does not touch the neighboring building) gets most of the irradiation, followed by the roof, as expected. However, 22% of the irradiation falls on the three facades that shouldn't have any irradiation at all.

My first assumption was that the building might not directly touch the neighboring building on the shapefile (which I doubt, given that I snapped the lines when drawing the shapefile on CAD), but even if there was a small gap between the building and the neighbor 22% of all irradiation on those surfaces sounds very high to me.

JIMENOFONSECA:
@martin-mosteiro mind sending me the case study?
thnaks!

PR #2416, opened 2019-10-25
Radiation now account for adjacent buildings

JIMENOFONSECA:
@martin-mosteiro this should fix it #2397


martin-mosteiro:
@JIMENOFONSECA line 407 in **geometry_generator.py** (`instersection = calculate.point_in_solid(sensor_cord, solid)`) appears to print a message that didn't use to happen as far as I can tell:
```
('State', 1)
```
(prints out once for every `solid` in `building_solid_list`)

Does this happen for you too?

martin-mosteiro:
In addition to the above, the geometry generator doesn't seem to work on my case study anymore. I left it running for ~2 hours and it stayed stuck in the loop that contains the line mentioned above.

This was _not_ the case for the `reference-case-open`. Then again, there's no adjacent buildings in the reference case, so I wonder if that's the difference?

JIMENOFONSECA:
Weird. It should work for any case. Can u try again?. The messagge is normal...use multiprocessing
On 25 Oct 2019, 22:18 +0800, Martín Mosteiro-Romero <notifications@github.com>, wrote:

In addition to the above, the geometry generator doesn't seem to work on my case study anymore. I left it running for ~2 hours and it stayed stuck in the loop that contains the line mentioned above.

This was not the case for the reference-case-open. Then again, there's no adjacent buildings in the reference case, so I wonder if that's the difference?

martin-mosteiro:
@JIMENOFONSECA Unfortunately not. I have now tested on two computers and the `geometry_generator` works on neither, although the master branch works on both. I've been using multiprocessing in all cases.

martin-mosteiro:
@JIMENOFONSECA Indeed, it worked now. The results are still strange, though. While we now have no windows in adjacent walls (good), for a test building 4% of the insolation still occurred on adjacent walls. While 4% is not all that much, is it also not easy to filter the surfaces by insolation, either, as the adjacent walls get a similar insolation per m2 as certain exposed surfaces.

JIMENOFONSECA:
@martin-mosteiro sorry I made a mistake in there. Does this work now? if so, mind creating the test files ? thanks!!

JIMENOFONSECA:
@martin-mosteiro did it work now?

martin-mosteiro:
Successfully tested on two Hochschulquartier scenarios and on the reference-case-open. The tests are not passing because the data in the reference case is outdated (no `intersection` column). After updating this data I would assume it'll work.

Issue #1633, opened 2018-10-02
Heat gains of whole GFA and whole envelope are allocated to conditioned space


gabriel-happle:
Some smaller buildings in the new TPW case study are overheating and resulting in very unrealistic conditions.

I think the reason is that their air-conditioned area is relatively small compared to the GFA (small `Hs` value).

If all buildings are defined as mixed use with a high residential fraction in Singapore, typically `0.25 < Hs < 0.65`.

This means that all solar heat gains and all appliance and person heat and humidity gains are concentrated on 25% of the space, leading to very high cooling loads and in extreme cases the code of the RC-model to break down.

I think a solution would be to only account for a fraction of the heat gains in such buildings.
However, I am not sure yet. I'm still investigating the situation.
As I remember there was already a similar issue raised by @martin-mosteiro earlier.

@JIMENOFONSECA, @BhargavaKrishnaSreepathi, @shizhongming, @shanshanhsieh I think we should address this before we run simulations for our upcoming publications. What could be the action plan?

JIMENOFONSECA:
Yes I think we made a previous work around that, and it somehow broke Martin’s case study.

I imagine creating again a new factor for the amount of electrified area (responsible of internal loads). this should be different to Hs.


On 2 Oct 2018, at 5:12 PM, Gabriel Happle <notifications@github.com<mailto:notifications@github.com>> wrote:


Some smaller buildings in the new TPW case study are overheating and resulting in very unrealistic conditions.

I think the reason is that their air-conditioned area is relatively small compared to the GFA (small Hs value).

If all buildings are defined as mixed use with a high residential fraction in Singapore, typically 0.25 < Hs < 0.65.

This means that all solar heat gains and all appliance and person heat and humidity gains are concentrated on 25% of the space, leading to very high cooling loads and in extreme cases the code of the RC-model to break down.

I think a solution would be to only account for a fraction of the heat gains in such buildings.
However, I am not sure yet. I'm still investigating the situation.
As I remember there was already a similar issue raised by @martin-mosteiro<https://github.com/martin-mosteiro> earlier.

@JIMENOFONSECA<https://github.com/JIMENOFONSECA>, @BhargavaKrishnaSreepathi<https://github.com/BhargavaKrishnaSreepathi>, @shizhongming<https://github.com/shizhongming>, @shanshanhsieh<https://github.com/shanshanhsieh> I think we should address this before we run simulations for our upcoming publications. What could be the action plan?

[...]

PR #2420, opened 2019-10-29
I1633 gains and losses proportional to conditioned area

gabriel-happle:
This implements changes discussed in #1633 
It is expected to fail the unit-tests, because the results change.
We first have to update the databases accordingly, before we update the results.

Major changes:
- construction_properties.xlsx[ARCHITECTURE] contains now `Hs_ag` and `Hs_bg`: fraction of conditioned floor area above ground and fraction of conditioned floor area below ground.
- heat transfer properties in the RC for roofs, walls, and windows above ground are calculated with `Hs_ag` and `sqrt(Hs_ag)`, respectively
- heat transfer properties in the RC model for walls below ground are calculated with `Hs_bg` and `sqrt(Hs_bg)`, respectively
- solar heat gains in the RC model take into account a factor `sqrt(Hs_ag)`, this is a simplification
- internal gains from appliances and lights take into account a factor `min(Af/Aef, 1.0)`

To do:
- update `Hs_ag` and `Hs_bg` in `CH` and `SG` databases with realistic values
- update all example files databases and `architecture.dbf`
- update unit-tests

@JIMENOFONSECA, who can take care of the databases and who can do the testing?


JIMENOFONSECA:
@gabriel-happle I think @lguilhermers and I can help to test and update the databases. thanks for this PR

martin-mosteiro:
@gabriel-happle I adjusted the ZH case study to have a consistent `Hs_ag` and `Hs_bg` for each building. Will do some testing and comment as necessary.

JIMENOFONSECA:
@martin-mosteiro mind giving me a hand with the unittests of schedules? All the rest in the dodo pass. But I am not sure how to recreate and fix those values

daren-thomas:
status update: according to @martin-mosteiro this needs to be tested still with the CH cases.

JIMENOFONSECA:
An update on behalf of luis, It works for Singapore case quite well

JIMENOFONSECA:
@martin-mosteiro we merge the solar radiation one (which accounts for shading surfaces) now there is a conflict in the thermal loads and the reference case.

martin-mosteiro:
So, I adjusted the ETH case study for testing purposes. I considered the following cases:
1. if 0.0 < `Hs` ≤ `floors_ag / (floors_ag + floors_bg)`: `Hs_ag = Hs * (floors_ag + floors_bg) / floors_ag` (i.e., all heated surfaces are above ground)
2. if `floors_ag / (floors_ag + floors_bg)` < `Hs`: `Hs_ag = 1` and `Hs_bg = [Hs * (floors_ag + floors_bg) - Hs_ag * floors_ag] / floors_bg` (i.e., the entire floor area above ground is conditioned, plus some parts below ground).
3. for a number of buildings (the ones where no information from the building owners were available and in which the CEA archetypes had been implemented directly) I also considered the case where `Hs_ag = Hs` and `Hs_bg = 0`, as the current CEA archetypes indicate.

Here is a quick overview of what I found for these three cases:
Hs  = floors_ag : (floors_ag + floors_bg)
Hs   floors_ag : (floors_ag + floors_bg)
Hs_ag = Hs, Hs_bg = 0

In all cases, the general trend is a quite noticeable decrease in heating demand and increase in cooling demand. This is due to the fact that even though the overall heated area for a given building is maintained, the solar gains per heated m2 change with respect to the old calculation (i.e., for a building with `Hs=0.9` as most of the old archetypes indicated, if the solar gains used to be 100%, now they are 95%).

Does this make sense? Well, I don't know, to be honest. In part because I'm not sure where the archetypal `Hs` values (0.82 for residential and 0.9 for almost everything else) came from in the first place. Was the 0.82 meant to be 82% of all above-ground surfaces? Or was the assumption that the basement + any unheated areas (say, the staircase) added up to 18%?

The adjustment seems reasonable on the one hand, but it also seems like we're forcing CEA to behave a certain way in order to be able to model one very specific typology in Singapore. Without having the time to do a detailed analysis of the demand model before and after this change compared to measured data, I can't say the current model is better than the old one. Our claim so far was that the CEA model was able to accurately model the case study in Zug as published in 2015. I'm not sure this will still be the case after this change.

JIMENOFONSECA:
Hi Martin,

I wonder if in a next stage we can calibrate the values of hs_ag and hs_bg to match measured data. Again we will need some measure data to at least calibrate the Archetypes.

I agree that the values of 0.92 and 0.9 were defined for the specific case of Zug. Yet this is relative, and I believe that CEA may not be still giving the same results (we did a lot of implementations in the past 4 years. What I can do however Is to create the case of Zug as our new open case (suing google street maps) and then manually calibrate the archetypes for now.

So if you agree, we merge and continue with this direction? we will need to update the unitttests though
On 5 Nov 2019, 1:06 AM +0800, Martín Mosteiro-Romero <notifications@github.com>, wrote:

So, I adjusted the ETH case study for testing purposes. I considered the following cases:

  1.  if 0.0 < Hs ≤ floors_ag / (floors_ag + floors_bg): Hs_ag = Hs * (floors_ag + floors_bg) / floors_ag (i.e., all heated surfaces are above ground)
  2.  if floors_ag / (floors_ag + floors_bg) < Hs: Hs_ag = 1 and Hs_bg = [Hs * (floors_ag + floors_bg) - Hs_ag * floors_ag] / floors_bg (i.e., the entire floor area above ground is conditioned, plus some parts below ground).
  3.  for a number of buildings (the ones where no information from the building owners were available and in which the CEA archetypes had been implemented directly) I also considered the case where Hs_ag = Hs and Hs_bg = 0, as the current CEA archetypes indicate.

Here is a quick overview of what I found for these three cases:
[Hs = floors_ag : (floors_ag + floors_bg)]<https://user-images.githubusercontent.com/18548065/68140898-dd6c0a80-ff2c-11e9-8f07-0143baf45d19.png>
[Hs floors_ag : (floors_ag + floors_bg)]<https://user-images.githubusercontent.com/18548065/68140903-e066fb00-ff2c-11e9-8bf3-f2471f454736.png>
[Hs_ag = Hs, Hs_bg = 0]<https://user-images.githubusercontent.com/18548065/68140908-e2c95500-ff2c-11e9-860a-e10fd4b65616.png>

In all cases, the general trend is a quite noticeable decrease in heating demand and increase in cooling demand. This is due to the fact that even though the overall heated area for a given building is maintained, the solar gains per heated m2 change with respect to the old calculation (i.e., for a building with Hs=0.9 as most of the old archetypes indicated, if the solar gains used to be 100%, now they are 95%).

Does this make sense? Well, I don't know, to be honest. In part because I'm not sure where the archetypal Hs values (0.82 for residential and 0.9 for almost everything else) came from in the first place. Was the 0.82 meant to be 82% of all above-ground surfaces? Or was the assumption that the basement + any unheated areas (say, the staircase) added up to 18%?

The adjustment seems reasonable on the one hand, but it also seems like we're forcing CEA to behave a certain way in order to be able to model one very specific typology in Singapore. Without having the time to do a detailed analysis of the demand model before and after this change compared to measured data, I can't say the current model is better than the old one. Our claim so far was that the CEA model was able to accurately model the case study in Zug as published in 2015. I'm not sure this will still be the case after this change.

martin-mosteiro:
> Yet this is relative, and I believe that CEA may not be still giving the same results (we did a lot of implementations in the past 4 years.

That is a very good point. With so many changes to CEA, whatever the accuracy of the results for Zug were ~4 years ago is not necessarily the current accuracy.

The split into `Hs_ag` and `Hs_bg` makes a lot of sense to me because so far we were actually assuming underground floors were heated – which they are usually not. What I'm not so sure about is the scaling of the gains because it's not based on anything in particular, just some simplifying assumption. A building with an `Hs_ag` of 50% could still be getting 100% of the solar gains if the unheated spaces were in the middle of the floor area, far from the windows, or they could get none at all.

To put it a different way, my mental justification for using a single thermal zone model was always that if, say, a residential building has an unheated staircase, then the staircase will still be passively heated through the uninsulated walls of the apartments. So, to me, one insulated perimeter would mean one thermal zone accounting for all gains and losses through the envelope. In order to model a building with different heated and unheated spaces would require a multi-thermal zone model.

However, I don't have the time to develop such a model at the moment, and nobody else in our group does. So this simplifying assumption might make sense but should be verified. Since I do not have the time at the moment to do so either, I can't really confirm or deny that the model is better than it was. Intuitively, given the boundary conditions we are setting, the changes make sense.

PR #2431, opened 2019-11-01
Update scripts.yml

### LINKED ISSUE #2430: some CEA scripts were missing access to a multiprocessing parameter
I ran into this error when trying out the workflow script from pycharm.
The parameter that is missing is  `general:number-of-cpus-to-keep-free`.

gabriel-happle:
- add  'general:number-of-cpus-to-keep-free' to radiation, optimization and sensitivity scripts

This fixes #2430 and makes the workflow run from pycharm


gabriel-happle:
@daren-thomas I can also report that the cooling workflow runs smoothly once the region and weather parameters are set to Singapore.

PR #2448, opened 2019-11-07
Ventilation to avoid unrealistically high temperatures during heating season

martin-mosteiro:
This PR implements the following changes:

**1-Make heat recovery bypass available:**
So far we assumed this was not possible. Based on discussions with the Zurich team, this appears to be not only possible but common. Otherwise ventilation would have little effect on the indoor temperatures since the air being circulated in the building would be preheated by heat recovery.

**2-Make night flushing available year-round** (so far it was only available during the cooling season)
This seems like a reasonable assumption given that certain building occupancy types are overheating during the day according to the archetypal loads (in particular RETAIL). The effect on the reference case open is quite minor, though (see graphs below).

**3-Make economizers available year-round** (so far they were only available during the cooling season)
Again, this seems like a reasonable assumption: if the building is overheating during the day, it would make sense to increase ventilation to cool it down.

**4-Make economizers standard in new constructions**
The archetypes so far assumed most building occupancy types had a ventilation system type 'T1', meaning mechanical ventilation with heat recovery and night flushing, but no economizer. This seems to be insufficient for RETAIL buildings given the loads defined by SIA standards. Making ventilation available when the indoor temperature gets too hot intuitively does not seem like such a complicated approach. So I assumed all buildings built or renovated after 2006 would have ventilation systems type 'T2' (i.e., same as 'T1' but with economizers).
**5-Make future residential buildings have ventilation systems**
Since Minergie buildings require mechanical ventilation, it seems like a reasonable assumption that residential buildings built after 2021 would have mechanical ventilation. So I added this to the archetypes database.

I tested the aforementioned changes on `reference-case-open` and found the following.

**B01 - 75% FOODSTORE, 25% COOLROOM**
The internal temperature in intermediate seasons reaches ~28ºC when running the current CEA simulation (no night flushing or economizers during the heating season) as well as when night flushing is indeed available year-round. When an economizer is added, the peak temperature in the intermediate seasons decreases by about 1ºC.
B01_T_int

The same is true about the heating and cooling demands throughout the year. The availability of an economizer reduces the peak cooling demand by about 20%.
B01_Qhs_Qcs

**B06 - 50% RETAIL, 50% RESTAURANT**
The peak temperature in autumn reaches 29.8ºC. The addition of night flushing in spring reduces the peaks somewhat, but not in autumn. The addition of an economizer keeps indoor temperatures below 27ºC year-round.
B06_T_int

The heating demands are unchanged by the change in ventilation controls, whereas cooling demand is decreased significantly by the economizer. The peak demand also decreases by about 30%.
B06_Qhs_Qcs

Looking at the yearly demands of all buildings, the heating demand is practically unchanged by night flushing, whereas the heating demand changes by 0-2% with the economizer (average: +0.1%). The cooling demand on the other hand goes up very slightly due to night flushing (+0.2% on average) whereas the cooling demand goes down significantly when the economizer is available (-27%-+3%, average -12.5%). The electricity demand is unchanged by night flushing and goes down by 1.2% due to the economizer.
B01-B09_Qhs_Qcs B01-B09_Qhs%_Qcs%


martin-mosteiro:
```
GitHub pull request #2448 of commit bf5884a297e3a452220c3777f5ecb7205e5e11ab, no merge conflicts.
Setting status of bf5884a297e3a452220c3777f5ecb7205e5e11ab to PENDING with url https://ceajenkins.ngrok.io/job/run%20cea%20test%20for%20pull%20requests/2110/ and message: 'Build started for merge commit.'
Building in workspace C:\Program Files (x86)\Jenkins\workspace\run cea test for pull requests
[WS-CLEANUP] Deleting project workspace...
[WS-CLEANUP] Deferred wipeout is used...
ERROR: [WS-CLEANUP] Cannot delete workspace: Unable to delete 'C:\Program Files (x86)\Jenkins\workspace\run cea test for pull requests'. Tried 3 times (of a maximum of 3) waiting 0.1 sec between attempts.
ERROR: Cannot delete workspace: Unable to delete 'C:\Program Files (x86)\Jenkins\workspace\run cea test for pull requests'. Tried 3 times (of a maximum of 3) waiting 0.1 sec between attempts.
Setting status of bf5884a297e3a452220c3777f5ecb7205e5e11ab to FAILURE with url https://ceajenkins.ngrok.io/job/run%20cea%20test%20for%20pull%20requests/2110/ and message: 'Build finished. '
Finished: FAILURE
```

Not sure what the issue is on the Jenkins. @daren-thomas maybe you have an idea?

Issue #1777, opened 2019-01-31
Add multiprocessing capabilities to the decentralized script


daren-thomas:
As a Researcher, I want the decentralized script (`cea decentralized`) to run on multiple cores (`-- general:multiprocessing on`) so the script runs faster.

daren-thomas:
Contact @shanshanhsieh for more information about this script / testing

JIMENOFONSECA:
@shanshanhsieh @daren-thomas could we add this to an epic? thanks!

daren-thomas:
Reference run: (to be compared to during performance optimization)

```
λ cea decentralized
City Energy Analyst version 2.26.0
Running `cea decentralized` with the following parameters:
- general:scenario = C:\Users\darthoma\Documents\CityEnergyAnalyst\projects\reference-case-cooling\baseline
  (default: {general:project}\{general:scenario-name})
- decentralized:detailed-electricity-pricing = False
  (default: False)
Running decentralized model for buildings with scenario = C:\Users\darthoma\Documents\CityEnergyAnalyst\projects\reference-case-cooling\baseline
B001  decentralized cooling supply system simulations...
[...]
Cost calculation...
2928.0433042 seconds process time for the decentralized Building Routine

Run decentralized model for buildings
Execution time: 2931.58s
```
2931.58s = *48.86min*

daren-thomas:
By using sympy to solve the system of equations for the variable `q_hw_KW`, I was able to reduce the above to:

```
C:\Users\darthoma\Documents\GitHub\CityEnergyAnalyst (1777-optimize-decentralized-script-for-speed -> origin)
λ cea decentralized
City Energy Analyst version 2.26.0
Running `cea decentralized` with the following parameters:
- general:scenario = C:\Users\darthoma\Documents\CityEnergyAnalyst\projects\reference-case-cooling\baseline
  (default: {general:project}\{general:scenario-name})
- decentralized:detailed-electricity-pricing = False
  (default: False)
Running decentralized model for buildings with scenario = C:\Users\darthoma\Documents\CityEnergyAnalyst\projects\reference-case-cooling\baseline
B001  decentralized cooling supply system simulations...
[...]
Cost calculation...
222.8987958 seconds process time for the decentralized Building Routine

Run decentralized model for buildings
Execution time: 226.28s
```
226.28s = **3.77min** which is a little over 10x faster...

PR #2467, opened 2019-11-21
1777 optimize decentralized script for speed

daren-thomas:
This PR addresses #1777 but actually goes a bit farther. The calculation in `cea.technologies.chiller_absorption.calc_operating_conditions` was replaced - the `sympy.linsolve` method call is not needed if you already have a symbolic solution to the equation. See the docstring for that method on how this was done. The end product is a really long and hairy expression but _so much faster_.

Execution speed went from:
- original: 2931.58s
- using expression: 226.28s
- smarter use of AbsorptionChiller (less pandas __getitem__ calls): ~180s
- added multiprocessing, 7 cores:  96.41s

I only worked on the cooling case. Do we need to do the same for heating?


JIMENOFONSECA:
@daren-thomas awesome!!, yes, the same for heating, please!

shanshanhsieh:
@daren-thomas , the implementation works well!
i will wait until PR #2478 is merged, so i can run the workflows and tests before merging this branch.

shanshanhsieh:
@daren-thomas could you please resolve the conflict? thanks

daren-thomas:
@shanshanhsieh: no conflicts found :) let's merge already!

Issue #2474, opened 2019-11-26
Choosing an incorrect shading system causes all architecture parameters to be imported incorrectly


martin-mosteiro:
When choosing the type of shading, choosing an incorrect shading type (e.g., 'T4') causes all other architecture parameters (e.g., the window-to-wall ratios) to be imported as NaN. There should probably be an error warning instead.

martin-mosteiro:
Solved by #2475.

Issue #2473, opened 2019-11-26
Building property `bpr.architecture.win_wall` is outdated


martin-mosteiro:
BuildingPropertiesRow includes a parameter called `win_wall`, which seems to be a remnant from the old days when we had only one window-to-wall ratio per building for all façades. At the moment, it is set equal to the following: `self.win_wall = envelope['wwr_south']`

i.e., the window-to-wall ratio of the south-facing façade is taken as the window-to-wall ratio for the entire building, which is evidently wrong. Nevertheless, this parameter appears to be used only in one method in the entire CEA, namely `cea.demand.space_emission_systems.get_delta_theta_e_sol`, so this is not an urgent issue. However, the buildings WWR should be calculated as an average of the WWR in all façades rather than taken from only one façade.

martin-mosteiro:
Solved by #2475.

PR #2475, opened 2019-11-26
Window to wall ratio and envelope properties

martin-mosteiro:
This PR solves issues #2473 and #2474, namely:
1-The wrong window to wall ratio was being used in one calculation in CEA. This has been changed from `bpr.architecture.win_wall=wwr_south` to `bpr.architecture.win_wall=A_win / (A_win + A_wall)`
2-Specifying an incorrect material in the architecture inputs (e.g., `type_shade` 'T4') made buildings be skipped entirely, such that if all buildings had such an incorrect parameter, no architecture parameters would be passed to the demand calculation at all. This has been solved by adding an error message that warns the user which buildings have incorrect input parameters of which type.


daren-thomas:
@martin-mosteiro regarding 2: is an error message enough? I think the script should just crash!

gabriel-happle:
I tested this with a wrong roof type in reference case open.
The code works.
Do you think it should display a warning or crash @daren-thomas ?
I can modify the `print` statement to raise an Error if you want.

martin-mosteiro:
> @martin-mosteiro regarding 2: is an error message enough? I think the script should just crash!

@daren-thomas Done!

Issue #2310, opened 2019-09-16
Input Editor: Show the list of valid supply systems in dropdowns


daren-thomas:
![image.png](https://images.zenhubusercontent.com/57b40957e40e5714b16d7331/ae803229-0860-4a4b-83c9-7205706995f3)

When editing the supply-systems, the user needs to know the system id to use. These are rather cryptic (all starting with `"Tx"`) and need to be looked up in "{general:scenario}\inputs\technology\lifecycle\LCA_infrastructure.xlsx".

Instead, let's let the computer do the work for us. When editing the supply systems, each cell in the columns `type_cs`, `type_dhw`, `type_hs`, `type_el` should show a dropdown that let's the user choose one of the valid systems. The dropdown should provide information from LCA_infrastructure.xslx, including "code", "source_hs", "scale_hs", "eff_hs", "reference" (or similar - basically, the whole worksheet...)

Issue #2141, opened 2019-07-05
Adding input validation for input editor in dashboard


reyery:
Currently, the input editor in the dashboard only does basic input validation i.e. `string or number`. This is not enough as there are some constraints on inputs such as those in `occupancy` where the sum of the columns must be equal to 1 since each column is the fraction of the gross floor area.

Instead of writing such constraints one by one for each table in javascript, the `schema.yml` should contain such information and javascript should just parse it from there.

daren-thomas:
We could model this similar to how SQL databases specify this information. Maybe check sqlalchemy which a) is already a dependency of the CEA and b) includes ways to define this kind of stuff.

PR #2479, opened 2019-11-28
Data validation for inputs in front end

reyery:
This PR provides a way to implement input validation for the front end, which helps with #2310 and #2141

It introduces a new property **_constraints_** for fields in _input.yaml_, which could be used to specify the range of the field i.e. **_max_**.

A new field type **_choice_** is also added to restrict users to only select values found in a list. 
A property **_location_** is used to extract this list of values from a database.

Other Changes:
Update `technical_systems` to `air_conditioning_systems`


daren-thomas:
@reyery I wonder if we should be using `schemas.yml` instead of `input.yml`? That way, we'd have one source of truth for our data?

JIMENOFONSECA:
@daren-thomas maybe let's open a new issue to move inputs.yml to schemas

daren-thomas:
@reyery @JIMENOFONSECA as far as I can tell, this can be merged - it's what I presented at the demo.

PR #2492, opened 2019-12-10
Fixing plots nyc

JIMENOFONSECA:
@daren-thomas @shanshanhsieh as part of user feedbak in NYC i updated the demand/load_curve plots and the solar potentital/photovoltaic potential plots. 

They now can work with a normalization and a timeframe, which is pretty cool.

@shanshanhsieh mind testing if this works for you?
@daren-thomas mind helping me to speed up the plots (specially the data reading and chache thing. I tried my best, but the plots run pretty slow. Specially that one of load curve. thanks!


daren-thomas:
@JIMENOFONSECA I took a quick look, but I'm really not sure what you mean. What is "pretty slow"? Did you measure? How did you measure? As far as I can tell, from using the optimizer on the `load_curve.py` file itself (so without cache), the main time spent is in loading plotly (this only needs to be done once, so not a problem for the GUI) and in calculating hourly loads (which is cached anyway) - and that necessarily requires reading each csv file. The actual plot code _we_ wrote doesn't take any time at all in comparison.

I've pushed an update to use the `MemoryPlotCache` which should also minimize reading data from the file system, but apart from that I'm really not sure where the problem is. Can you show me exactly what needs optimizing?

Issue #2395, opened 2019-10-21
Missing input files should create an error, not a success in the jobs list


daren-thomas:
Steps to reproduce:

- create new scenario (chose all the helpers)
- forget to run `data-helper`
- run `radiation`

Expected: Error status on job, detailing the missing files

Observed behavior: Job is completed as "success", but the output of the script shows "ERROR: Missing input files..."

I think the code that checks for these files should end with an error code (`sys.exit(1)`)

PR #2498, opened 2020-01-06
Missing input files should create an error, not a success in the jobs list

daren-thomas:
This PR addresses #2395 - Missing input files should create an error and the job info in the CEA GUI should reflect that:

image

To test, follow steps from the issue (basically, just remove all data-helper output and run radiation)

### DIFF
--- a/cea/api.py
+++ b/cea/api.py
@@ -26,7 +26,7 @@ def register_scripts():
             cea_script.print_script_configuration(config)
             if list(cea_script.missing_input_files(config)):
                 cea_script.print_missing_input_files(config)
-                return
+                raise cea.MissingInputDataException()
             t0 = datetime.datetime.now()
             # run the script
             script_module.main(config)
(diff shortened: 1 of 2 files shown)


reyery:
Works great!

However, the error after running **radiation** does not show any `print_missing_input_files`
 (_just a minor issue we can solve later_).
Seems like `print_missing_input_files` is getting `script_suggestions` from **schemas.yml**
(_which we should probably update soon_). 
I was just thinking, since there is a `input-files` property in **scripts.yml**, we could also add a `output-files` property later so we could trace the `script_suggestions` from there instead.
Was playing around, we could create a DAG with that information and maybe use it for some kind of task queuing. What do you think?

image

daren-thomas:
@reyery that is kind of what the `trace-inputlocator` script is trying to accomplish - collecting this data. I think we should spend some time to update this from it's current (a bit hacky) state. Maybe make it launch a `workflow.yml` file or something?

reyery:
@daren-thomas alright, will try to look at `trace-inputlocator` again. But from what I remember, it tries to get the information but running each script right?

daren-thomas:
@reyery exactly. of course we can also hand-curate the `schemas.yml`. It was intended as a starting point.

Issue #2447, opened 2019-11-07
Sometimes the path to the Daysim binaries gets overwritten


daren-thomas:
- find out when
- find out why
- fix this

possible solutions:
- radiation script could check for existence of these binaries before running
- add special case check for standard windows installation and override to Depencies\Daysim

daren-thomas:
other possible solution: add a Daysim folder to CityEnergyAnalyst repo with subfolders: windows, unix, macos.

PR #2499, opened 2020-01-07
Check for daysim binaries

daren-thomas:
This branch addresses #2447 by checking for the existence of the Radiance/Daysim binaries _before_ running the radiation script.

It uses the value in the config file as a starting point and then checks some other obvious places (Daysim subdirectory of Dependenecies folder as installed by installer, `C:\Daysim\bin`, `%RAYPATH%`).

If the binaries can't be found, an error is raised, otherwise, the simulation is run with the path found.

### DIFF
--- a/cea/resources/radiation_daysim/radiation_main.py
+++ b/cea/resources/radiation_daysim/radiation_main.py
@@ -5,6 +5,7 @@ from __future__ import print_function
 from __future__ import division
 
 import os
+import sys
 import pandas as pd
 import time
 from cea.utilities import epwreader
@@ -33,14 +34,14 @@ def create_radiance_srf(occface, srfname, srfmat, rad):
 
 
 def calc_transmissivity(G_value):
-    '''
+    """
     Calculate window transmissivity from its transmittance using an empirical equation from Radiance.
 
     :param G_value: Solar energy transmittance of windows (dimensionless)
     :return: Transmissivity
 
     [RADIANCE, 2010] The Radiance 4.0 Synthetic Imaging System. Lawrence Berkeley National Laboratory.
-    '''
+    """
     return (math.sqrt(0.8402528435 + 0.0072522239 * G_value * G_value) - 0.9166530661) / 0.0036261119 / G_value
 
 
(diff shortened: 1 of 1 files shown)


daren-thomas:
@JIMENOFONSECA can you please check the PR?

JIMENOFONSECA:
It is surprising that we did not add the binaries to the folder of cea as agreed over the phone last year. I am for that plan instead of this PR.  Maybe I am missing something, @daren-thomas  mind confirming the reasons of this change?

see in issue #2447 
 "...other possible solution: add a Daysim folder to CityEnergyAnalyst repo with subfolders: windows, unix, macos...."

daren-thomas:
@JIMENOFONSECA the main installation method does not check out the repo, so the binaries would not be available then. Also, we currently don't have other binaries (macos, unix), so I chose the implementation that is more robust and works with what we have. Adding the other binaries to the repo can still be done, and this method will work when we create installers for those platforms. It doesn't really look like we'll be doing that in near future though.

JIMENOFONSECA:
@daren-thomas got it. so since macos and linux support is not in the pipeline. I suggest to remove it once and for all and ocus on windows and getting it to work seamlessly. Daysim binaries can be added to the cea source code with MIT license agreement. that should work right?

I just discussed with Arno, since the project finishes in march 31 (this take at least), it will not be realistic to have more support to euler or other things that use macOS or Linux. 

In that order of ideas we can just eliminate macOS or Linux support from the documentation (which to tell the truth never existed 100%).

So what about we proceed to add the binaries and end support of macOS and Linux?

daren-thomas:
@Jimeno Fonseca <fonseca@arch.ethz.ch>
we already ship the binaries for Windows. This PR points to those binaries
if it can't find them from the location pointed at in the config file.

I suggest leaving the documentation as hints for future users to figure out
how to get Linux and MacOS working by themselves - Just not invest any more
effort into it. I _do_ suggest merging the PR for the linux
environment.ubuntu.yml though. It's work already done and not merging it
does not make a lot of sense.

We can officially state that we won't add support for these operating
systems but welcome contributions from the community, what do you think?

Issue #2491, opened 2019-12-09
New Scenario on fresh install does not end up in project folder


daren-thomas:
Master / 2.27.0a0 - Steps to reproduce:

- fresh CEA installation (no `cea.config`, clean virtual machine)
- start CEA GUI
- create new project:
  - name: "Kleinalbis"
  - location: "C:\Users\User\Desktop"
- create new scenario
  - name: "baseline"
  - use the open street maps workflow
- open Windows Explorer, navigate to Desktop

Expected: A folder called "Kleinalbis" with a subfolder "baseline" with the scenario contents

Buggy behaviour: two folders, one (empty) called "Kleinalbis" and one called "baseline".

Assumption: The default values setting the project might be based on the default value for scenarios and that might be "C:\" or something on a fresh installation. This could be tested with a pre-configured system.

Idea for fix: Installer could set the project root to something like "Desktop" - user can still change it, but the default values would then work better?

Issue #2483, opened 2019-12-03
error when create a scenario importing input files


lguilhermers:
I was not able to import a .shp file to make those as my "zone" file. I even tried with CEA's own files created in previous projects but the error persists.

When I click in "Create", a new window saying "Creating Scenario" shows up, but goes away right after and the import is not successful. 

![image](https://user-images.githubusercontent.com/50981039/70022927-d1797400-15d0-11ea-8588-4ece5cd465ed.png)

daren-thomas:
@lguilhermers can you provide a test case?
@reyery should the interface maybe show more information if something went wrong?

lguilhermers:
@daren-thomas I tested in the new version (2.27) with CEA shp from other projects and it is working (before it wasn't).
I will have to correct the projection of the shp file I want to import because I think this might be the error source. I update here once I fix it and then try importing it again.

daren-thomas:
@lguilhermers this will be fixed with the next release (2.28) expected tomorrow or Monday.

lguilhermers:
@daren-thomas thanks for the update! I also discussed with Jimeno yesterday and maybe I was missing some steps on my side (e.g. adding some columns in the attribute table of the .shp file). I will try next week in 2.28 and with those additional steps and update you. thanks!

daren-thomas:
@lguilhermers the new version is out, please try it out and report back on whether it solves your problem :)

lguilhermers:
Sorry, I missed that. Yes, everything seems to be running smoothly importing external shp, given that the attribute tables and georeferences are set up correctly. If there is a higher demand for user- imported zones, perhaps a post in the blog with some tips would be useful to not miss anything that is key for CEA. I can help by tracing my steps and issues I faced. :) Thanks, @daren-thomas!

PR #2503, opened 2020-01-08
Fix for scenario creation

reyery:
This PR tries to fix some of issues when creating a new scenario such as new scenarios not being created in the project folder (#2491) and no error messages if the script fails (#2483)

Fixes:
-
- Use a temp config when creating new scenario since changing scenario name and then running `config.save()` does not give the new scenario path sometimes. (not sure if it solves #2491 as I was able to replication this issue)
- Wrap create scenario functions in `try...except` to catch exceptions and send a `500` code and the error trace back to the front end.
- Allow generation of scenario specific plot parameters such as `BuildingParameter` when scenario changes


reyery:
Testing:
- Try to create a new scenario that will throw an error 
  - (e.g. select an area and use `surrounding-helper` when there are not surrounding buildings nearby, or just add a `throw Exception` line to any of the data management tool scripts)
  - an Error message should appear in the modal and show the error message and the stack trace
- Create 2 scenarios with different number of buildings then try to create a new plot in the dashboard (the default PV Potential plot should do)
- Change the scenario parameter and check the buildings parameter. It should change based on the buildings of the scenario. Selecting buildings that do not exist in the selected scenario will show an error message

daren-thomas:
Testing:
- OK: exception happens from zone helper, error message shown in GUI as predicted
- BUG: tried to create second scenario using copy method ("Copy input folder from another scenario in the project")
  - scenario folder created,
  - doesn't contain input folder... what happened?
- OK: Copied scenario from reference-case-open/baseline, deleted B01
  - created plot, selected B01 in baseline, changed to 2nd scenario, error message shows up
  - lists also change

for the BUG referenced above, I'm creating a new issue: #2508

cea/interfaces/dashboard/api/project.py
```diff
@@ -46,7 +48,7 @@ def get(self):
             config.scenario_name = ''
             config.save()
         return {'name': name, 'path': config.project, 'scenario': config.scenario_name,
-                'scenarios': config.get_parameter('general:scenario-name')._choices}
+                'scenarios': get_scenarios_from_project(config)}
```
daren-thomas: I like this! This could also be made into a Configuration method, but for now, let's keep it like this. OR: moved to the `cea.config` namespace? That would be even better!

cea/interfaces/dashboard/api/project.py
```diff
@@ -102,94 +104,104 @@ class Scenarios(Resource):
     def post(self):
         """Create new scenario"""
         config = current_app.cea_config
+        temp_config = cea.config.Configuration()
```
daren-thomas: why does this need to be a temporary configuration object? you need to add a comment here to explain.

cea/interfaces/dashboard/api/project.py
```diff
@@ -102,94 +104,104 @@ class Scenarios(Resource):
...
+                    trace = traceback.format_exc()
+                    return {'message': e.message, 'trace': trace}, 500
 
         elif payload['input-data'] == 'copy':
-            source_scenario = os.path.join(config.project, payload['copy-scenario'])
-            shutil.copytree(cea.inputlocator.InputLocator(source_scenario).get_input_folder(),
-                            locator.get_input_folder())
+            try:
+                os.makedirs(scenario_path)
+                source_scenario = os.path.join(temp_config.project, payload['copy-scenario'])
```
daren-thomas: this might be where #2508 got introduced...

cea/interfaces/dashboard/api/project.py
```diff
@@ -298,3 +311,7 @@ def get(self, scenario):
         else:
             abort(400, 'Scenario does not exist', choices=choices)
 
+
+def get_scenarios_from_project(config):
```
daren-thomas: This could be moved to either a method on config (`scenarios`) or a module-level function in `cea.config` - though... first check if it's used anywhere else in the code: If not, then we just leave it here...

daren-thomas: should probably be called "list_scenario_names_for_project" (as we tend to use the name "scenario" to refer to the whole path most of the time.

cea/interfaces/dashboard/api/project.py
reyery: Fixed in bca24d8. Probably something to do with `os.makedirs(scenario_path)`. Did not have this issue when trying to copy a new scenario, only when copying a scenario that already has some outputs from scripts. weird.

cea/interfaces/dashboard/api/dashboard.py
```diff
@@ -23,13 +24,30 @@ def dashboard_to_dict(dashboard):
...
+    print(scenario)
     parameters = []
-    for pname, fqname in plot_class.expected_parameters.items():
+    plot_parameters = sorted(plot_class.expected_parameters.items(), key=lambda x: x[1])
+    # Make sure to set scenario name to config first
+    if 'scenario-name' in [parameter[0] for parameter in plot_parameters]:
+        if scenario:
+            config.scenario_name = scenario
+        elif hasattr(plot_class, 'parameters') and 'scenario-name' in plot_class.parameters:
+            config.scenario_name = plot_class.parameters['scenario-name']
```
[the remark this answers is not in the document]
reyery: Some plots have `scenario-name` as a parameter to show scenario specific plots. That is why if the `scenario-name` parameter is not set first in the `plot_class`, fetching scenario dependent parameters like `BuildingParameter` will still be based on the old `scenario-name`

cea/interfaces/dashboard/api/dashboard.py
```diff
@@ -167,9 +184,12 @@ def put(self, dashboard_index, plot_index):
         # Set parameters if included in form and plot exists
         if 'parameters' in form:
             plot = dashboard.plots[plot_index]
-            print('expected_parameters: {}'.format(plot.expected_parameters.items()))
-            for pname, fqname in plot.expected_parameters.items():
-                parameter = config.get_parameter(fqname)
+            plot_parameters = plot.expected_parameters.items()
+            if 'scenario-name' in [parameter[0] for parameter in plot_parameters]:
```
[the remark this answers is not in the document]
reyery: Ah yes, why did I not see that I could just iterate through the dict.

cea/interfaces/dashboard/api/project.py
reyery: After reading through it again, I don't think there is a need to do that here.

cea/interfaces/dashboard/api/dashboard.py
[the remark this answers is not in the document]
daren-thomas: my point is it's `plot_class`, not a `plot_object`... so the `scenario-name` can't be correct? I just checked: The name of the argument to `get_plot_parameters` should be `plot` - then my confusion goes away. But I see it's also used in another place, so we'll need to check this carefully.

cea/interfaces/dashboard/api/dashboard.py
```diff
@@ -157,6 +173,7 @@ def put(self, dashboard_index, plot_index):
         """
         form = api.payload
         config = current_app.cea_config
+        temp_config = cea.config.Configuration()
```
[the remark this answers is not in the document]
daren-thomas: What are the main pain points with `config`? What could we do to fix that? (when we started, `config` was used to store parameters and info about parameter types for scripts to run. with the web interface, the concept starts showing it's age - but I'm not sure how to fix it easily without breaking command line usage of the CEA - ideas?)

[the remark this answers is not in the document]
daren-thomas: I added a comment to explain - I get it now.

daren-thomas:
Works as advertised! :)

PR #2513, opened 2020-01-10
fix nominal cap

JIMENOFONSECA:


gabriel-happle:
@JIMENOFONSECA looks good.

PR #2516, opened 2020-01-10
fixing small bugs i2512 and i2509

gabriel-happle:
- change chiller type in district cooling cost to centralized chiller 'CH1'
- fix inconsistent if statement in chiller cost calculation


gabriel-happle:
How to test bug i2509 in pycharm python console:
```
import cea.inputlocator
locator = cea.inputlocator.InputLocator(scenario="C:/reference-case-cooling/baseline")
from cea.technologies.chiller_vapor_compression import calc_Cinv_VCC
calc_Cinv_VCC(3500000, locator, 'CH3')
Out[5]: (89556.82063313726, 63110.44327573253, 1262208.8655146505)
calc_Cinv_VCC(0, locator, 'CH3')
Out[6]: (0, 0, 0)
calc_Cinv_VCC(3500001, locator, 'CH3')
Out[7]: (89556.84622080032, 63110.461307287755, 1262209.226145755)
```

JIMENOFONSECA:
@gabriel-happle for the type of system, it is better if you use the constant value. I mean, you import the variable from constants and use it in the code. In that way we do not hardcode 'CH1' etc..

gabriel-happle:
> 
> 
> @gabriel-happle for the type of system, it is better if you use the constant value. I mean, you import the variable from constants and use it in the code. In that way we do not hardcode 'CH1' etc..

Done.

Issue #2525, opened 2020-01-20
District cooling network temperature error


martin-mosteiro:
`substation_main_cooling` calculates the network temperatures as follows (line 111):

```
T_DCN_supply_to_cs_ref = np.vectorize(calc_DC_supply)(T_DC_supply_to_cs_ref, T_DCN_supply_to_cs_ref)
```
where (from `calc_DC_supply`, line 863):
```
    if t_0 <= 0.0 and t_1 <= 0.0:
        t_0 = 0.0
        t_1 = 0.0
    elif t_0 <= 0.0:
        t_0 = 1E6
    elif t_1 <= 0.0:
        t_1 = 1E6

    tmin = min(t_0, t_1)
```

Since COOLROOMs have nominal operating temperatures of 1°C/5°C (from the CEA database), `T_DC_supply_to_cs_ref = -1` (as calculated by `calc_temp_this_building_cooling`, line 183) for buildings with COOLROOMS. Thus, `T_DCN_supply_to_cs_ref = 0`. So far so good.

Now imagine the next building has SERVERROOMs but no COOLROOMs. These have a nominal operating temperature of 7°C/15°C, such that `T_DC_supply_to_cs_ref = 5`. When recalculating the network temperature, since `T_DCN_supply_to_cs_ref = 0` and `T_DC_supply_to_cs_ref = 5`, `T_DCN_supply_to_cs_ref = 5`.

So later when `calc_substation_cooling` gets called in `substation_model_cooling` (line 279) the nominal conditions for the building with COOLROOMs are `thi_0 = Tcdata_sys_re_C = 5` and `tci_0 = T_DCN_supply_to_cs_ref_data_C = 5`, and since `cc_0` is proportional to `1/(thi_0 - tci_0)`, you get `cc_0 = inf`, `UA_cooling = inf` and eventually a ZeroDivisionError in `calc_HEX_cooling` (line 585).

The solution seems quite simple: network temperatures below zero should not be assigned a value of 0, but rather something very small, i.e.:
```
    if t_0 < 0.0:
        t_0 = 1E-6
    if t_1 < 0.0:
        t_1 = 1E-6
    if isclose(t_0, 0.0) and isclose(t_1, 0.0):
        t_0 = 0.0
        t_1 = 0.0
    elif isclose(t_0, 0.0):
        t_0 = 1E6
    elif isclose(t_1, 0.0):
        t_1 = 1E6

    tmin = min(t_0, t_1)
```

PR #2526, opened 2020-01-20
Fix COOLROOM bug in thermal networks

martin-mosteiro:
As described in #2525, the way `calc_DC_supply` was written created an error in buildings with COOLROOMs if there were also SERVERROOMs present in the case study.

@JIMENOFONSECA perhaps you have some thoughts on the change made on `calc_DC_supply`?


JIMENOFONSECA:
@martin-mosteiro i made a change so it works for all cases. if it is ok for you, then go ahead an merge. It works for me

martin-mosteiro:
@JIMENOFONSECA Thanks for generalizing this. However, I don't think the code did the same thing as before. Since 0 values are used to indicate the temperatures of systems which are not operated, the current version of the script will give 1E-6 any time one of the systems is not operated – even though it should give the supply temperature of the systems that _are_ operated.

So I'm reverting the changes and refactoring the code to generalize it as you did but while keeping `t0 = t1 = 0.0` as a special case.

Ideally, the temperatures of the systems that are not operated should not be 0... but I that would require time to refactor the code, which we don't have at the moment.

martin-mosteiro:
Okay, I have now updated the following:
1. fixed `calc_DC_supply` following @JIMENOFONSECA's solution, but removing a bug by which supply temperatures would've always been 1E-6 or lower (solves #2525).
2. fixed another temperature bug (#2529) by which disconnected buildings were also used to set the operating temperature of the thermal network
3. fixed `COOLROOM` process schedules - I just saw that these would've led to 0 kWh refrigeration for all cases!

PR #2546, opened 2020-01-31
Glossary.csv - synchronized with newest schemas.yml

daren-thomas:
This PR is about as far as we can go to get the `glossary.csv` up-to-date _without_ actually manually editing all the "TODO"s - this still needs to happen and I will be creating mini-issues for each file I can't figure out myself and hope to assign those to more knowledgable CEA devs.

The synchronizing was done with a Jupyter notebook (`notebooks/update_glossary.ipynb`).

Testing hints: Try and search for glossary items in the interface.


JIMENOFONSECA:
@daren-thomas I spent some hours filling in the glossary manually. The new version overwrites those changes several times with "TODO". I wonder if this PR should first check the past glossary and only update the terms that have TODO.  Maybe I am missing something..

daren-thomas:
@JIMENOFONSECA this is _exactly_ why we need code reviews! Good catch! Of _course_ nothing should be overwritten. I'll check again to see if I can find your changes. (This is probably due to filenames changing and similar issues).

daren-thomas:
@JIMENOFONSECA I've just checked (see latest commit) and I can't find any non-TODO items in the master version of `glossary.csv` that are overwritten by this PR. Can you please point out the specifics?

My check depends on the locator_method and variable name being the same. Maybe your changes are with out-dated locator_method and variable names?

daren-thomas:
ok. i found some (few) rows that were missed, mainly because the filename and locator_method had changed and updated those, see latest commits.

Issue #2534, opened 2020-01-23
UX : Optimisation should select what technologies to run and which to not


JIMENOFONSECA:
Taken from #2469
- Technologies a well as buildings.

Issue #2373, opened 2019-10-09
District cooling plant plot does not run without optimizing for cooling


JIMENOFONSECA:
This issue was identified as part of PR #2295 

so in essence that plot cannot run without optimizing for district cooling. How would you like to handle that error? @daren-thomas in the front or the back-end. if so how? I am opening a separate issue for this.

daren-thomas:
@JIMENOFONSECA ideally, we use the `input_files` array in the plot class: This approach assumes there is a file that is only present when the optimization for district cooling has been run. The `input_files` array is a list of tuples `(locator_method, list_of_arguments_to_locator_method)` and is checked by the plotting system - the system will output an error (plus a hint of which script to run) if a file is not found.

So, to fix this, all you need to do is add a new input file description to the plot class' `__init__` that can identify if optimization cooling has run.

daren-thomas:
@JIMENOFONSECA are we talking about the plot "Dispatch curve for cooling plant in system"? (I'm not quite sure which plot to check as I couldn't find a plot called "District cooling plant")

JIMENOFONSECA:
Hi Daren, yes, that one. The problem is that either way (ridding for district heating, or running for district cooling) we are generating exactly the same files. What changes are the values of the files. But that is not known apriori. So, the only way would be to create a “dummy file” that describes that the problem was run only for district heating or cooling, right?
On 21 Oct 2019, 6:40 PM +0800, Daren Thomas <notifications@github.com>, wrote:

@JIMENOFONSECA<https://github.com/JIMENOFONSECA> are we talking about the plot "Dispatch curve for cooling plant in system"? (I'm not quite sure which plot to check as I couldn't find a plot called "District cooling plant")

daren-thomas:
@JIMENOFONSECA what I noticed was that the file `outputs\data\optimization\slave\gen_1\ind_0_Cooling_Activation_Pattern.csv` doesn't have any columns except for the `DATE`:

```
DATE
2005-01-01 00:00:00
2005-01-01 01:00:00
2005-01-01 02:00:00
2005-01-01 03:00:00
```

The heating version of the file has the following headers:

```
[...]

Issue #2325, opened 2019-09-19
Refactor all the databases


daren-thomas:
Check
- naming (files, tables, columns)
- what get's used by what script? (we have some hints from the TraceInputLocator)
- create a plan on how to move forward

JIMENOFONSECA:
I made a major refactoring...more to come

PR #2550, opened 2020-02-05
I2534 opt technologies

JIMENOFONSECA:
This solves issues #2534 and #2373, adding great functionality for our optimization users.

To test, just run the optimization algorithm for any case study and for a couple of generations, 10 individuals will do. Then create a few supply systems' plots.


JIMENOFONSECA:
@daren-thomas any chance to get this tested? I am waiting for this to pass so I can update a PR based on #2325

Issue #2553, opened 2020-02-06
Restructuring and Reducing the CEA database (up for discussion)


shanshanhsieh:
Here are some points we discussed in the CEA meeting (6th Feb), we still need to make the final calls in the coming days.
@martin-mosteiro @daren-thomas @shizhongming any thought is welcome!

- [x] **1.** Remove Primary Energy everywhere (the behavior is linear to emission)
- [x] This information is currently linked to: optimization, life-cycle analysis, feedstocks, mobility, embodied energy.
- [x] @martin-mosteiro , are you using PE as a performance indicator?
- [x] **2.** Move LIFECYCLE into ARCHETYPES as an EMISSION FACTOR tab
- [x] **3.** Add a few tabs to the **input database** that allows the user to modify easily. 
- [x] - New tab: "EMISSION INTENSITY" (see point 2)
- [ ] **4.** Clean up the columns inside the system database
- [ ] - USD-2015
- [ ] - remove Type (in the "PV" tab)
- [x] - changes in these tab involve many changes in the scripts
- [x] **5.** Add information about cost calculation to "supply_systems" tab
- [ ] **6.** Add a yaml file to specify in which tab a new row can be added.
- [ ] - cannot: "cooling" and "heating" tabs under air_conditioning_systems
- [ ] **7.** Validate input values in each cell 
- [ ] - some columns can only contain unique value
- [ ] - some cells can only take a limited number of values (e.g. T1, T2, T3)
- [x] **8.** where too keep:  "FEEDSTOCK",  "ALL_IN_ON_SYSTEM"?
- [x] move feedstocks to a separate spreadsheet and combine it with detailed electricity costs. 24 hour input so the optimisation can use different hourly tariffs during the day. Emission factors can also be an opportunity.

martin-mosteiro:
1. What do you mean by PEN being linear to emission?
Here's what the CH feedstocks in CEA look like:
![image](https://user-images.githubusercontent.com/18548065/73940344-960a6700-48eb-11ea-8b13-096a6f82ca31.png)
The Swiss grid is a huge outlier because it's low emission but highly PEN intensive (due to nuclear).

Here's what it looks like for SG:
![image](https://user-images.githubusercontent.com/18548065/73940673-2648ac00-48ec-11ea-949e-90f89ca12a10.png)

Bottom line: I do use PEN as an indicator, but I'm not really using CEA for any of these calculations, so it doesn't matter to me much.

[...]

PR #2557, opened 2020-02-10
I2553 restructure database

JIMENOFONSECA:
Solving all the points of #2553 in here.

To test, download the latest interface in CEA.
Assign a new database to your project.
Check the databases.
Run any tool in CEA.


gabriel-happle:
@JIMENOFONSECA 
I think the database editor looks great.
I don't think there needs to be changes for the `SYSTEMS`.

The main issue that I see is how (or if) the user is allowed to create new `building-use`/`occupancy` types in the `ARCHETYPES` databases. Because if yes, that has consequences not only on the database but also on the scenario input files.
Here is what I mean:
-> Let's assume the user creates a new line in `Construction_properties/architecture` and calls it `SUPERMARKET`
-> information on `SUPERMARKET` has then to be created for all other `construction_properties` table
-> BUT also a new schedule for `SUPERMARKET` has to be created.
-> *BUT!!!* also the all EXISTING and NEW `occupancy.dbf` files should have an additional column called `SUPERMARKET`.

I will think a bit more about this. Maybe the `occupancy` can be in a more flexible form instead of this matrix-like structure. It's anyways not so nice to have a file where most fields are `0`.
Or maybe `building construction type` and `building use-type` should be separated. I remember that was an issue we discussed a long time ago.

Other minor ideas:
- schedules for different `building-use` could be selected from a drop-down list instead of so many tabs
- `create new schedule based on...`  and `create new building use-type based on...` would be cool functionality
- a description text could be added to each database
- `Feedstocks` could be separated in `Heating` and `Electricity` and `Cooling`
- why does the `Feedstock` now have 24 values? It seems to be related to time, but why? Except for electricity, I don't think anything gets traded hourly


@shanshanhsieh @martin-mosteiro I'm tagging you here because you commented on the issue #2553

JIMENOFONSECA:
@gabriel-happle thanks for the comments.

@reyery mind adding validation to the databases of **ISSUE 1**? A warning message to the user could be sufficient for now. Let's open and make it a separate issue, so we avoid this PR becoming bigger and bigger.

@gabriel-happle **ISSUE 6 and ISSUE 7** the reason behind is that we needed to merge detailed costs of electricity with the feedstocks database. The compromise was 24 hour inputs. These inputs allow us to explore multiple tariffs during the optimisation routine. It also will allow to explore multiple supply mix (in terms of emissions) in the near future. Should this behaviour not be needed, the user can just put the same value everywhere. CEA nonetheless does the costs and emissions calculation in an hourly basis.

@gabriel-happle **ISSUE 2** mind writing a few more words, I am not sure if I understand what you have in mind.

gabriel-happle:
@JIMENOFONSECA more details to *ISSUE2* above:

the `building construction type` is now inferred as the main-use of occupancy/building use. It could be defined by the user.
the `building use type` input is currently restricted to the predefined columns in `occupancy.dbf`. If we want to allow the user to have more control over the databases according to *ISSUE1*. The inputs could look something like this:

`building-construction.dbf` (now called `age.dbf`)
| name | construction-type | built | roof | hvac | ...|
| B01 | RESTAURANT | 2000 | 2000 | 2010 | ... |

`building-use.dbf` (now called `occupancy.dbf`)
| name | use |
| B01 | RESTAURANT:0.4, OFFICE:0.5, GYM:0.1 |
| ... | (this is a string input) |

In this way, the user could create new construction and occupancy categories without restrictions.

PR #2559, opened 2020-02-12
Update how-are-schedules-defined.rst

AlexJew:
Correction of a typing mistake


daren-thomas:

JIMENOFONSECA:
@AlexJew thanks! good catch

PR #2563, opened 2020-02-14
2227 negative values in ct

### LINKED ISSUE #2227: Negative values in CT calculation
when running `calc_CT` at close to full load, the electricity demand `wpartload` is negative
need to check the reference for the range of operation

shanshanhsieh:
this PR resolves #2227 

after testing, there is no instance that the electricity usage is negative
but i made unittest functions for cooling towers
@daren-thomas please check if i did them right, thanks!


JIMENOFONSECA:
@daren-thomas mind taking this review over, thanks!, important is that things do not get slower, neither in testing nor in the workflow

PR #2571, opened 2020-02-18
fix skytemp

gabriel-happle:
- fix equation
- use opaque skycover `opaqskycvr_tenths` from weather file directly for calculation
- add documentation


gabriel-happle:
I'm running the tests to see how much this influences the results.

gabriel-happle:
Since the equation should be correct now, I think we have to accept the changes.
I'm therefore updating the tests.

gabriel-happle:
I don't know why the tests fail. They are passing on my machine.

daren-thomas:
@gabriel-happle it's something with the Jenkins server, I'll check it tomorrow - I can't fix it remotely with TeamViewer for very boring reasons (I deleted the software on my local computer and now can't access it anymore...)

cea/utilities/epwreader.py
```diff
@@ -33,15 +33,26 @@ def epw_reader(weather_path):
...
+    or:
+    https://bigladdersoftware.com/epx/docs/8-6/engineering-reference/climate-calculations.html
+
+    :param Tdrybulb: Dry bulb temperature [C]
+    :param Tdewpoint: Wet bulb temperature [C]
+    :param N: opaque skycover in [tenths], minimum is 0, maximum is 10 see: http://glossary.ametsoc.org/wiki/Sky_cover
+    :return: sky temperature [C]
+    """
+
+    sky_e = (0.787 + 0.764 * math.log((Tdewpoint + 273) / 273)) * (1 + 0.0224 * N - 0.0035 * N ** 2 + 0.00028 * N ** 3)
```
daren-thomas: I noticed you're dividing by an integer. Let's get in the habit of using floats when we mean floats. I also checked and this file is missing the `from __future__ import division`.

gabriel-happle: OK. Thanks.

Issue #2595, opened 2020-02-21
Reference cooling case is not updated


daren-thomas:
Have we tested the SG case yet with the 2.31.0 databases? I'm getting the following error:

```
Workflow step: script=archetypes-mapper
Running archetypes_mapper with args {'input_databases': ['comfort', 'architecture', 'air-conditioning', 'internal-loads', 'supply', 'schedules'], 'buildings': [], 'scenario': 'C:\\Users\\darthoma\\Documents\\CityEnergyAnalyst\\projects\\reference-case-cooling\\baseline'}
City Energy Analyst version 2.31.0
Running `cea archetypes-mapper` with the following parameters:
- general:scenario = C:\Users\darthoma\Documents\CityEnergyAnalyst\projects\reference-case-cooling\baseline
  (default: {general:project}\{general:scenario-name})
- archetypes-mapper:input-databases = ['comfort', 'architecture', 'air-conditioning', 'internal-loads', 'supply', 'schedules']
  (default: ['comfort', 'architecture', 'air-conditioning', 'internal-loads', 'supply', 'schedules'])
- archetypes-mapper:buildings = []
  (default: [])

---------------------------
ERROR: Missing input files:
Traceback (most recent call last):
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\Scripts\cea-script.py", line 11, in <module>

    load_entry_point('cityenergyanalyst', 'console_scripts', 'cea')()
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\interfaces\cli\cli.py", line 66, in main
    script_module.main(config)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\workflows\workflow.py", line 56, in main
    do_script_step(config, step)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\workflows\workflow.py", line 138, in do_script_step
    run(config, py_script, **py_parameters)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\workflows\workflow.py", line 29, in run
    f(config=config, **kwargs)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\api.py", line 52, in __call__
    self._runner.__call__(*args, **kwargs)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\api.py", line 28, in script_runner
    cea_script.print_missing_input_files(config)
[...]

Issue #2594, opened 2020-02-21
reference-case-open is not updated


daren-thomas:
I'm trying to run the workflow (to be able to test all the new plots!) and ran into this problem:

```
C:\Users\darthoma\Documents\GitHub\CityEnergyAnalyst (2541-documentation-refactoring -> origin)
λ cea radiation
City Energy Analyst version 2.31.0
Running `cea radiation` with the following parameters:
- general:scenario = C:\Users\darthoma\Documents\CityEnergyAnalyst\projects\reference-case-open\baseline
  (default: {general:project}\{general:scenario-name})
- general:multiprocessing = False
  (default: True)
- general:number-of-cpus-to-keep-free = 1
  (default: 1)
- general:debug = False
  (default: False)
[...]
verifying geometry files
C:\Users\darthoma\Documents\CityEnergyAnalyst\projects\reference-case-open\baseline\inputs\building-geometry\zone.shp INFO:numexpr.utils:NumExpr defaulting to 8 threads.
getting geometry materials
creating 3D geometry and surfaces
Reading terrain geometry
Creating 3D building surfaces
Traceback (most recent call last):
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\Scripts\cea-script.py", line 11, in <module>

    load_entry_point('cityenergyanalyst', 'console_scripts', 'cea')()
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\interfaces\cli\cli.py", line 66, in main
    script_module.main(config)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\resources\radiation_daysim\radiation_main.py", line 464, in main
    config)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\resources\radiation_daysim\geometry_generator.py", line 510, in geometry_main
    height_col='height_ag', nfloor_col="floors_ag")
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\resources\radiation_daysim\geometry_generator.py", line 240, in building_2d_to_3d
    data_preprocessed = BuildingData(locator, settings, geometry_terrain, height_col, nfloor_col)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\resources\radiation_daysim\geometry_generator.py", line 145, in __init__
    self.zone_building_solid_list = np.vectorize(self.calc_zone_building_solids)(self.zone_building_names)
[...]

PR #2604, opened 2020-02-25
2595 2594 update reference cases

JIMENOFONSECA:
This solves #2595 and #2594.

We introduce a new reference case study, where we have measurements. It should work for both cooling and heating. It is diverse enough to test some key features in CEA.

Since we have measurements, in the future, we should check against the demand simulation to pass the tests. That remains part of future work. However the measured data can be found already cleaned in inputs/building-measurements.

To test:
1. extract the new reference case and run solar radiation. cea/examples/reference-case-open
2. run also demand.

@daren-thomas would you mind testing it? thanks!


daren-thomas:
I'm nearly ready to merge - just doing some tests. The move to the workflows allows us to do a shorter test for pull requests and a more exhaustive (read: run all the scripts, run all the unittests) for merging to master.

As before, you can test your changes locally by running:

- bin\ceatest.bat
- bin\ceatestall.bat

which are equivalent to pip installing and doing `cea test --workflow quick` and `cea test --workflow slow` respectively. Note that you can also do `cea test --workflow unittests` to just run the unittests.

PR #2620, opened 2020-03-02
Emissions plot missing

### LINKED ISSUE #2619: Emissions plot missing
ups. we erased this in the previous release.

JIMENOFONSECA:
This PR fixes #2619 

to test:
1. run lifecycle/emissions
2. run plots/scenario_comparisons/Anual emissions from the dashboard


reyery:
Currently the parameter `scenarios-and-systems` accepts an empty value which causes an error when plotting the plot. I'm not sure how we should handle this. I thought of "selecting all" values when it is empty but that would mean a long list to plot.

JIMENOFONSECA:
It hink we can make the scenarios-and-systems parameter accept as default in general:Scenario_today_

Issue #2584, opened 2020-02-20
Add migration script for users using older input databases


reyery:

daren-thomas:
@JIMENOFONSECA how far should we go with this?

One solution would be to:

- use database exported with database editor (or CH/SG)
- merge `age.dbf` and `occupancy.dbf` into a `typology.dbf`
- let the user select STANDARD (edit the database to include the STANDARD?)
- run archetypes-mapper

Problem: This undoes any manual changes the user made to the input tables. So:

- update the other `building-properties` tables
- this requires updating the database with new entries (for assemblies / feedstocks etc.)
- (archetypes would not need to be updated by the migration, as the user isn't expected to run archetypes-mapper again).

Anyway. I'm exploring this stuff currently in a branch (`2584-migration-script`) in a notebook. I'm unsure how much effort we should put into this. What do you think?

daren-thomas:
@JIMENOFONSECA can we just assume that databases in v2.29.0 were left untouched? That way the migration script could map T0 -> HVAC_COOLING_AS0 etc. That would make so much stuff so much easier, but if a user had already changed the databases... well... then things would get really tricky to migrate. 

Unless they just have to select a standard and then run the `archetypes-mapper`. I kind of like that idea... (i managed to migrate to the `typology.dbf`)

JIMENOFONSECA:
I think we can leave it to that point you mentioned.but the steps could be something like this. Mind the order.

1. We: The database is reinizialized to the new one
2. We: Occupancy and age file are transformed to typology. While at it, the standard is automatically selected based on the built age (check zone helper for how to do this) and the newly initialized database/archetypes/construction/standardefinition
3. The user: will have to run archetypes mapper.

I think that is what we should offer right now. We cannot offer more because there is not antmore a straight mapping between standards and the archetypes. We redid this completely.


On 4 Mar 2020, 22:57 +0800, Daren Thomas <notifications@github.com>, wrote:

@JIMENOFONSECA<https://github.com/JIMENOFONSECA> can we just assume that databases in v2.29.0 were left untouched? That way the migration script could map T0 -> HVAC_COOLING_AS0 etc. That would make so much stuff so much easier, but if a user had already changed the databases... well... then things would get really tricky to migrate.

Unless they just have to select a standard and then run the archetypes-mapper. I kind of like that idea... (i managed to migrate to the typology.dbf)

daren-thomas:
[...]

PR #2623, opened 2020-03-05
2584 migration script

daren-thomas:
This PR implements #2584  - a database migration script for migrating from v2.29.0 to v2.31.1.

The principle it works on is an ordered list of migrations to run in sequence. Each migration has a test and a migrator. The test returns True, if the scenario matches the start of the migration and False if this migration is not necessary. This setup can be used to implement further automatic migrations if necessary using the same framework.

The steps to use are:

- run data-initializer
- run database-migrator
- run archetypes-mapper

NOTE: The GUI could check the result of `cea.datamanagement.database_migrator.find_migrators` in the project overview (or when opening the input tables?) and direct the user to the migrator tool. The check would look like this:

```
if list(find_migrators(scenario)):
   print("you'll need to migrate this scenario."
   print("maybe make a copy first as this can't be undone")
```

@reyery do you think you could implement that into the GUI? I have attached a project/scenario for testing.

[2.29.0.zip](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/4293044/2.29.0.zip)


cea/datamanagement/database_migrator.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def find_migrators(scenario):
+    """
+    Add new migrations here as they become necessary
+    the database-migrator will run these in sequence starting from the first migrator found
+    (NOTE: I've added a dummy migration - 2.31 - 2.31.1 - to show how the principle works)
+    """
+    migrations = collections.OrderedDict()
+    migrations["v2.29.0 - v2.31.0"] = (is_2_29, migrate_2_29_to_2_31)
```
reyery: I'm still not too sure how this works. So supposedly we change the structure again down the road, do we just add a new migration entry to the dictionary? e.g. `v2.31.0 - v3.0.0`. Then if the user has `v2.29.0` input databases, this script would run the first migration function i.e. `v2.29.0 - v2.31.0` (since it would be `true` for the first migration function) then `v2.31.0 - v3.0.0`?

I'm thinking it might not work in the long run since the shipped databases for `v2.31.0` might not be the same as the ones shipped in the new version (e.g. `v3.0.0`) so the migration function that is trying to reference older shipped databases to do its job could cause some problems (since the installed version would only contain their own version of shipped databases, unless we fetch it from somewhere maybe?).

So instead it could be `v2.29.0 - {latest}`, `v2.31.0 - {latest}`, etc, for the different database structures. Unless users _really_ want to be able to switch between different versions. But this way would also mean we have to update each migration function to see how each version would map to the latest version.

Or maybe I'm missing something

cea/datamanagement/database_migrator.py
```diff
@@ -0,0 +1,147 @@
...
+    if os.path.exists(os.path.join(scenario, "inputs", "building-properties", "typology.dbf")):
+        # avoid migrating multiple times
+        return False
+    return True
+
+
+def migrate_2_29_to_2_31(scenario):
+    def lookup_standard(year, standards_df):
+        # find first standard that is similar to the year
+        standard = standards_df[(standards_df.YEAR_START < year) & (year < standards_df.YEAR_END)].iloc[0]
```
reyery: I'm getting `IndexError: single positional indexer is out-of-bounds` when `year` is 1971. I guess it should be `<=` instead of  `<` and handle the exception when the dataframe is empty

cea/datamanagement/database_migrator.py
```diff
@@ -0,0 +1,147 @@
...
+        return typology_dbf
+
+    age_dbf_path = os.path.join(scenario, "inputs", "building-properties", "age.dbf")
+    occupancy_dbf_path = os.path.join(scenario, "inputs", "building-properties", "occupancy.dbf")
+
+    age_df = dbf_to_dataframe(age_dbf_path)
+    occupancy_df = dbf_to_dataframe(occupancy_dbf_path)
+
+    locator = cea.inputlocator.InputLocator(scenario=scenario)
+    standards_df = pd.read_excel(locator.get_database_construction_standards(), "STANDARD_DEFINITION")
```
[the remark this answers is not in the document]
reyery: Ah right I forgot to run data-initializer first.

daren-thomas:
test this please

reyery:
Tested and works as intended with the scenario you provided. However it gives an error when trying another scenario, when the building year was equal to the terminal years of the range. Should be fixed in 616f9cb. I think we can merge this.

As for the GUI, I suppose we could run the function in input editor before it tries to read the files for now. Should we also do a check of the input databases first? Since it requires the user to have the latest databases to be able to run `database_migrator` successfully.

cea/datamanagement/database_migrator.py
daren-thomas: yes exactly. the migrations would be run in sequence. but that's a really good point you mentioned. I hadn't thought of that...

I guess that means that in the migrations scripts we can't really depend on the databases being a certain way. hm...

daren-thomas:
@reyery I don't understand your last question? 

> Should we also do a check of the input databases first? Since it requires the user to have the latest databases to be able to run database_migrator successfully.

Maybe you're referring to `get_database_construction_standards`? That is set up as an `input_file`in `scripts.yml`. My thinking was that this would be sufficient for running the migrator script?

Also. I'm not sure how much effort we want to spend on this right now - I'm assuming when we next write a migration we can revisit some of these trickier questions, but by then it might not really be as important to support 2.29 anymore and the script could just stop then with an error message.

Good job testing edge cases btw!

reyery:
Yea, it does give an error when `get_database_construction_standards` is missing. I was just thinking in terms of user experience on the GUI side of things (The user is directed to migration script from input editor, migration script fails because old databases, go to data initializer script, go back to migration script).

But besides that, the migration script works for now and I guess we could just merge this first for users who need to use it.

Issue #2621, opened 2020-03-03
AttributeError: 'numpy.ndarray' object has no attribute 'values


kerembeygo:
**Describe the bug**
"AttributeError: 'numpy.ndarray' object has no attribute 'values'"
[Job Output for 3 optimization.docx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/4281034/Job.Output.for.3.optimization.docx)

**To Reproduce**
1. Optimization
2. Supply System Part 2 (centralized)

**Expected behaviour**

**Screenshots**
![1](https://user-images.githubusercontent.com/29708612/75771863-d8dd2480-5d5b-11ea-89eb-3dbb79e60b09.jpg)

**Hardware (please complete the following information):**
 - OS and version: Windows 8.1
 - Version of CEA v. 2.31.1

daren-thomas:
Daren thinks the expression `district_electricity_demands['E_hs_ww_req_disconnected_W'].values` is probably supposed to be a series and somehow got set to an `ndarray` - that's where I'd start to check (maybe a little-used code path set this and it only shows up in specific configurations).

@JIMENOFONSECA I'm assigning this to you for now - is that ok?

PR #2639, opened 2020-03-13
2621 optimization fix

JIMENOFONSECA:
This solves #2621 

The problem had to do with passing series and sometimes passing arrays. Now it will always pass arrays. The problem should be solved now.

to test:
- run a few of the optimization routine. 
@daren-thomas mind taking a look at this PR? thanks!

### DIFF
--- a/cea/optimization/slave/electricity_main.py
+++ b/cea/optimization/slave/electricity_main.py
@@ -358,10 +358,10 @@ def extract_electricity_demand_buildings(master_to_slave_vars, building_names, l
     for name in building_names:  # adding the electricity demand of
         building_demand = pd.read_csv(locator.get_demand_results_file(name))
         # end-use electrical demands
-        Eal_req_W += building_demand['Eal_kWh'] * 1000
-        Edata_req_W += building_demand['Edata_kWh'] * 1000
-        Epro_req_W += building_demand['Epro_kWh'] * 1000
-        Eaux_req_W += building_demand['Eaux_kWh'] * 1000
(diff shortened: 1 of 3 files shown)


cea/optimization/slave/electricity_main.py
```diff
@@ -358,84 +358,80 @@ def extract_electricity_demand_buildings(master_to_slave_vars, building_names, l
...
             if name in buildings_connected_to_district_heating and name in buildings_connected_to_district_cooling:
                 # if connected to the heating network
-                E_hs_ww_req_W += 0.0
-                E_cs_cre_cdata_req_W += 0.0
+                E_hs_ww_req_W += np.zeros(HOURS_IN_YEAR)
+                E_cs_cre_cdata_req_W += np.zeros(HOURS_IN_YEAR)
             elif name in buildings_connected_to_district_heating:
                 # if disconnected from the heating network
-                E_hs_ww_req_W += 0.0
+                E_hs_ww_req_W += np.zeros(HOURS_IN_YEAR)
```
daren-thomas: nice! while the effect is the same, this communicates the "arrayness" of the variable :)

daren-thomas:
weird: `district-heating-systems` runs through, but `district-cooling-systems` fails in the `radiation` stage. It seems the `archetypes-mapper` shortens the list of buildings for the `architecture`, `air-conditioning-systems` and `supply-systems` input tables. I'm guessing it might be because the scenario is using the SG region and the `typology` doesn't map perfectly. I'll investigate this before merging.

PR #2641, opened 2020-03-16
Database validation class

reyery:
This PR adds an initial **InputFileValidator** class to validate input files based on the `schemas.yml`.

Now the schema allows columns and constraints information for each 'sheet'/'table' of excel data files.
Currently constraints are only row-based (making use of `pandas.eval()`)

e.g.
```
get_database_construction_standards:
  created_by: [data-initializer]
  file_path: inputs/technology/archetypes/CONSTRUCTION_STANDARDS.xlsx
  file_type: xlsx
  schema:
    STANDARD_DEFINITION:
      columns:
        STANDARD:
          type: string
          primary: true
        Description:
          type: string
        YEAR_START:
          type: int
          min: 0
        YEAR_END:
          type: int
          min: 0
      constraints:
        YEAR:
          YEAR_START < YEAR_END
```


reyery:
@daren-thomas this might conflict with what you are working on in this branch https://github.com/architecture-building-systems/CityEnergyAnalyst/tree/2637-single-point-of-reference-for-variables, with `schemas.yml`. Just letting you know what I have so far with this. I have only tested and changed values in `schemas.yml` for input databases.

cea/databases/__init__.py
```diff
@@ -0,0 +1,122 @@
+import os
+from collections import OrderedDict
+
+FILE_EXTENSIONS = ['.xlsx', '.xls', '.csv']
```
daren-thomas: what about `.dbf` files?

daren-thomas:
@reyery how do i test this? are you ready to merge?

cea/databases/__init__.py
reyery: This is only for constructing the file structure tree for input databases in `cea.databases`. I used it to get the database categories and the files that are part of that category. e.g. assemblies has ENVELOPE, HVAC and SUPPLY.

reyery:
There isn't much changes to the GUI (in terms of input validation), just how it interprets the new format of  `schemas.yml` (without hard coding anymore). So it should still work the same the previous version.

As for the **InputFileValidator** class, the main method for `cea.datamanagement.database_verification` actually validates database files of the current scenario. To test, insert invalid values into some of the databases and it should show up.
I have tried the validator with other input files and it seems to work. But since most of them still only have the `type` property, it will only check for the type. Adding more properties for these files would make it more useful.

daren-thomas:
tested. works.

PR #2643, opened 2020-03-17
I2021 pumping energy

shanshanhsieh:
the pumping energy calculation has been updated, only the pressure losses in the critical path is being considered now.
to test: (please test one DH and one DC case)
1. run `cea thermal-network` 
2. check all the thermal network plots from the interface


daren-thomas:
@shanshanhsieh who is supposed to be testing this?

martin-mosteiro:
@daren-thomas I'm testing as we speak

Issue #2597, opened 2020-02-24
GDAL and DAYSIM environment variables are not set for users running scripts from PyCharm


reyery:
Problem
-
Variables `GDAL_DATA`, `PROJ_LIB`, `RAYPATH` are not set when running scripts from PyCharm and this causes some unusual behavior such as in `radiation` script where it runs really fast

Solution
-
We could find a way to set these variables like how it done when running it from console. Or find some way to warn and teach users to include these variables in PyCharm using _configuration_, before running them.

daren-thomas:
similar to checking for Daysim paths, we could also check for GDAL paths and add those to `os.environ` at the beginning of the script.

daren-thomas:
@reyery is only the radiation script affected?

Issue #2589, opened 2020-02-20
Function checking for DAYSIM binaries for radiation script does not pass binaries installed by DAYSIM installer


reyery:
Currently CEA is checking for a list of files to be present to ensure that the radiation script can run.

https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/160cc8c1b2415e8784374fb98eeadc805273de16/cea/resources/radiation_daysim/radiation_main.py#L205-L215

However, it seems that it is only passing the directory of the DAYSIM binaries that we ship with the CEA installation and not the one installed using the original DAYSIM installer.

It is looking for 2 files `isotrop_sky` and `rayinit` that is not found in the `bin` folder but in the `lib` folder of the original DAYSIM directory.
```
c:\daysim\bin [('ds_illum', True), ('epw2wea', True), ('gen_dc', True), ('isotrop_sky', False), ('oconv', True), ('radfiles2daysim', True), ('rayinit', False), ('rtrace_dc', True)]
c:\daysim - copy\bin [('ds_illum', True), ('epw2wea', True), ('gen_dc', True), ('isotrop_sky', False), ('oconv', True), ('radfiles2daysim', True), ('rayinit', False), ('rtrace_dc', True)]
C:\Users\Admin\Documents\CityEnergyAnalyst\CityEnergyAnalyst\cea\resources\radiation_daysim [('ds_illum', False), ('epw2wea', False), ('gen_dc', False), ('isotrop_sky', False), ('oconv', False), ('radfiles2daysim', False), ('rayinit', False), ('rtrace_dc', False)]
c:\daysim\lib [('ds_illum', False), ('epw2wea', False), ('gen_dc', False), ('isotrop_sky', True), ('oconv', False), ('radfiles2daysim', False), ('rayinit', True), ('rtrace_dc', False)]
c:\users\admin\documents\cityenergyanalyst\dependencies\daysim [('ds_illum', True), ('epw2wea', True), ('gen_dc', True), ('isotrop_sky', True), ('oconv', True), ('radfiles2daysim', True), ('rayinit', True), ('rtrace_dc', True)]
```

We would need to investigate if `isotrop_sky` and `rayinit` are actually needed for our simulations.

daren-thomas:
@reyery if I remember correctly,`isotrop_sky` and `rayinit` are data files and they _are_ required - I believe I just ran the radiation script on an empty folder and kept adding the missing files until it ran through.

[...]

PR #2650, opened 2020-03-19
2597 gdal daysim variables from pycharm

daren-thomas:
This PR fixes 

- #2597 (GDAL and DAYSIM environment variables are not set for users running scripts from PyCharm) 
- #2589 (Function checking for DAYSIM binaries for radiation script does not pass binaries installed by DAYSIM installer)

I've tried a bunch of combinations, but please try to break this:

- remove the folder `C:\Daysim` (e.g. rename it to `C:\Daysim.bak`
- remove the Daysim installation from the CEA installer (check the `Dependencies` folder)

Running the scripts works from PyCharm as well as from the command line. I think we're done here.


reyery:
Tested and works as expected.

Issue #2638, opened 2020-03-13
Official end of support of Interfaces other than CEA Cli and CEA Dashboard


JIMENOFONSECA:
Picking up on the discussion of #2624, we would make official that CEA only supports our own interfaces. CEA dashboard and CEA Command line. This will help us to maintain focus on helping CEA users to have the best user experience possible. 

The steps to make this effective are:
1. Encourage users like @shizhongming to create their own repository (maybe copying what we already did) for new interfaces)
1. Move CEA Arcgis, CEA grasshopper to the Legacy folder of CEA.
2. Update documentation erasing traces of CEA Arcgis and CEA Grasshopper.

PR #2654, opened 2020-03-20
End of support for Grasshopper

JIMENOFONSECA:
This solves #2638 

No chance to build sphynx, daren mind giving me a hand? thank you!


daren-thomas:
ok, i'm on it!

PR #2673, opened 2020-03-27
2637 single point of reference for variables

daren-thomas:
This PR includes a _lot_ of work on the `schemas.yml` file as well as the `cea.inputlocator.InputLocator` class. The aim: To remove the `glossary.csv` file.

- `glossary.csv` file was removed and that information is now in `schemas.yml`
- new unit test file `cea/tests/test_schemas.py` checks integrity of `schemas.yml`:
  - all methods from `InputLocator` are described in `schemas.yml`
  - each column has a description
  - in future (after we do our little documentation effort) all the columns are tested for "TODO" (and the unit tests will fail if that's not the case`
  - all locator-methods (and folder-methods)  are unique - there was quite a bit of duplication going on!
  - all locator-methods are used - I removed any locator-methods that are used by legacy code. there's no point in dragging that along with us.
  - `cea-doc graphviz` now actually works and the outputs can be built by sphinx
  - `cea-doc glossary` now actually works and produces the expected output
  - `cea-doc html` runs through (with two minor warnings I hope to fix soon)
    - view the documentation here: https://city-energy-analyst.readthedocs.io/en/2637-single-point-of-reference-for-variables/index.html

I'm putting this out there now, so you know what I'm working on and maybe it's best to merge soon and then implement the other stuff next.

Stuff to work on next:

- remove `naming.csv` (I'll need to edit each and every plot)
- remove `inputs.yml` (the schemas are different, but compatible. i think. @reyery maybe you want to help here?)


daren-thomas:
NOTE: it's not ready to merge right now - I'm still testing it in the GUI. The Database Editor is currently failing.

daren-thomas:
@reyery would you mind testing this still works with the GUI? I've done some testing and I think this PR is good to go.

reyery:
@daren-thomas I have changed how the database editor filters the glossary due to the new format [here](https://github.com/architecture-building-systems/CityEnergyAnalyst-GUI/pull/23). Other than that, it looks good on the GUI side too.

daren-thomas:
@reyery does that mean we can merge? will merging the GUI be a problem?

reyery:
@daren-thomas yeah we could merge this. For the GUI, v3.0.2 GUI will still work with this PR, just missing the column glossary. The GUI PR just fixes that.

daren-thomas:
@reyery alright! please merge then :)

PR #2677, opened 2020-04-01
Plot data export

### LINKED ISSUE #2628: As a User, I would like to be able to download data from my plots on CEA as a .csv file
**Describe the feature you would like to have**
When creating a plot on CEA, I would like to be able to download the data from the plot..

**Describe the end purpose of this feature**
- Produce my own graph 
- Process the data to show relevant tendencies or important hidden correlations

**Describe the end-user of this feature**
- Better communication of the results
- Greater flexibility on the visualisation of results

**Provide Aditional information if possible**
No aditional information

reyery:
This PR resolves #2628 and add the ability to export plot data to .csv/.xlsx files for users.

This is done by parsing the data found in the plots that are encoded in the Plotly format to pandas dataframes and then exported. That means that this is actually an extra step since we are already calculating the data using dataframes before converting them to the Plotly format. If I'm not wrong, Plotly v3 actually has methods to plot straight from dataframes instead using Plotly graph objects in v2 that we are using now.

The current way it is implemented is more of a workaround than an actual solution. It would not exactly work on mostly plots but works best on plots that are purely made up of bar graph objects. However this should be sufficient for our needs now. I did not want to over-engineer this solution in the hopes that refactoring the plot classes and upgrading Plotly python library to be the wiser solution moving forward.

### DIFF
--- a/cea/interfaces/dashboard/api/dashboard.py
+++ b/cea/interfaces/dashboard/api/dashboard.py
@@ -246,4 +246,7 @@ class DashboardPlotInputFiles(Resource):
         dashboard = dashboards[dashboard_index]
         plot = dashboard.plots[plot_index]
 
-        return [locator_method(*args) for locator_method, args in plot.input_files]
+        data_path = plot.plot_data_to_file()
+        input_paths = [locator_method(*args) for locator_method, args in plot.input_files]
+
+        return {'inputs': input_paths, 'data': [data_path] if data_path else []}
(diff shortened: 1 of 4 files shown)


reyery:
Testing
-
After creating the required input files, try creating various plots and view the plot data files using the button on the top right of plots

image

daren-thomas:
image

This is exciting!

image

PR #2685, opened 2020-04-10
2661 update assembly database

emanuel030:
Cost and efficiency for cooling technologies was updated and expanded by AC Split-units in the SUPPLY database for SG. Further, the emission values for electricity in Singapore was updated in the FEEDSTOCKS file for SG.


daren-thomas:

JIMENOFONSECA:
@martin-mosteiro mind checking this?

martin-mosteiro:
@emanuel030 What is the source for the CO2 emissions in the FEEDSTOCKS file? The value changed but the source is still stated as before.

martin-mosteiro:
@emanuel030 Regarding the cooling technologies, are we sure these are the best references we can find?
- **Air/air chiller:** source is from Qatar (?), costs are 2X the original CAPEX in CEA
- **Water/air chiller:** costs 1.5X higher than before, no reference given
- **Lakewater/water district cooling:** source is for the Maldives (?), the costs are 11X higher than before and the efficiency is beyond the maximum COP allowed by CEA (hard-coded at around 8.5, if I'm not wrong)
- **Water/air district cooling:** finally a Singaporean source! :) However, it's hard to follow - eventually I found that you were citing the efficiency from an article where it was cited as "0.8 kW per RT-Hr" and converted it to find the number you used. I think it would be better if you write the formula in the Excel cell instead of writing a result of your calculation. For the costs part, it's not clear to me where you got the number you used, and the cost numbers in the Wikipedia article don't have any appropriate citation.
- **AC Split Units (Multi-split):** Marcel's paper appears reasonable for the efficiency, but what is "Online market"?

**TL;DR:** these references seem really hodgepodge and not necessarily representative (Qatar? Maldives?). It's hard to say which values are more reasonable since the old reference was just an "educated guess", but the values are _massively_ different to what was there before. So I'd suggest to get @shanshanhsieh's opinion on whether these values make sense for the Singapore context.

More fundamentally, can we somehow get a consistent source for all these values? Does, say, BCA have reference values (building energy benchmarking reports or something)?

JIMENOFONSECA:
@shanshanhsieh @emanuel030 this is kind of urgent, can you comment?

emanuel030:
Hi Martin, 

Thanks for checking. I think the previous source has no validity, since the numbers were quite random and it even stated the DCS to have a lower investment cost than AC-Multisplit units have. I would see them merely as place holders.

emanuel030:
Hi Martin,
Thanks for checking. I would look at the previous values as placeholders, since for example they stated the DCS investment cost with 100 USD/kW to be lower than that of Chillers with 200 USD/kW. From past experience we do know that this is not correct, since we are looking at aggregated costs, considering the whole DCS setup. In the past HGZZ project, where we manually calculated the cost of DCS and estimated costs of around 3500CHF/kW. Further, previous 'educated guess' stated 20 years lifetime of DCS, now I don't think its that 'educated' after all :)

**FEEDSTOCK**: For CO2 values of electricity Singapore I used IPCC 2014 values, but the previous source is also valid.

**SUPPLY**   
**Electricity**: The E/I&C costs are an interesting topic, since it touches on the scope of the CEA. Do we consider the installation of electrical equipment in buildings as part of the CEA scope? I would argue yes, since we also look at the embodied emissions of the building envelope, which is less related to energy systems then the E/I&C. Regarding this, the Swiss value of 1 USD/kW is not in the right ballpark. The source is not shareable.

 **Water/air chiller**: I added the source from literature (again foreign). I also contacted a firm in Singapore which sells chillers but they didn't get back to me, even after phone confirmation that they could help. No company is exactly eager to publish their cost data.

**Seawater/water district cooling**: I have found multiple sources in that price range. Our own calculations of the Lakewater DCS in Zurich were x3 in USD/kW. But Zurich was a special case due to long network from the lake to the HGZZ. So I consider the given value as very reasonable. I have adapted the efficiency of the DCS to account for the warm sea in Singapore and adapted it to 6.5, right in the range given by 'Bruelisauer et al. (2013)'. 

**Water/air district cooling**: Interesting one, since here we actually have a popular DCS in Singapore, but nowhere to find the costs data except for Wikipedia. I have added Cost for the plant and 60%for the 'Common Services tunnel', since there is no point of having the plant without the network, and in this case this network is multi-purpose.

**AC Multi-Split**: I have added the specific source, which is a Lazada.sg, a popular online shop in SG which happens to also sell AC's with installation.

New change: I have changed the Interest rate IR% to 1.3% for all systems according to Singapore long term interest rates https://www.ceicdata.com/en/indicator/singapore/long-term-interest-rate

I know the sources i have found are not ideal, but I believe that unless there is a industry or government (in SG case) partner it will be 'hodgepodged' :) I further think it should be updated more regularly, whenever there is new information found.

emanuel030:
After consultation with @JIMENOFONSECA, it is to clarify that the previous database values were dummy costs.

martin-mosteiro:
@emanuel030 Thanks for checking with @JIMENOFONSECA.

Actually, I'm a bit confused about what these values are used for, anyway (I haven't used CEA since the ASSEMBLIES were created). If there are CAPEX values for chillers (USD/kW), thermal networks (USD/m), etc. in the 'components' folder, why do we need a separate CAPEX value in USD/kW for district cooling?

In any case, I agree it's likely that there'll be a mix of source in the database. I'm just wondering if there's a better/more representative source than a presentation for a rail company in Qatar – Chillers are a very common technology in Singapore, I'm sure there must be some cost information somewhere.

emanuel030:
There is a new 'simplified' cost calculation, module system_costs, which calculates the aggregated costs. The components costs are used in the optimization process. For the current project we wanted to get the simplified costs thats why we update.

Prices for the chillers are always on request, I just phoned another company and wrote them an email with out request for cost data. Lets hope something comes around.

martin-mosteiro:
Good – now I understand why the lower costs for the district scale systems didn't make sense (I thought it was only chiller prices). In that case I agree – any value is better than a placeholder value. I'm just surprised there wouldn't be more local sources. Since @shanshanhsieh has more experience in getting system costs for SG, maybe she can chip in.

shanshanhsieh:
Thanks for your effort in collecting the data, @emanuel030 ! 
Please also make sure all the details are documented in your thesis. 
We will have to find better ways to store all the references in CEA.
Also, thank you @martin-mosteiro for going through these, i've never used this database myself either.

In SG, the efficiencies of systems are well-discussed, but the costs are indeed tricky.
The updated CAPEX for cooling systems seems okay to me, but you have to be aware that the CAPEX of DC is very case-specific. It might be worth conducting a sensitivity analysis on that one.

As for efficiencies of "Chiller-air-air" and "Chiller-water/air", have you looked into some data published from BCA? For example, the "BCA Building Energy Benchmarking Report" has values for chiller plants in commercial buildings. In 2016, they compared the average chiller plant efficiencies (Chart 31), the range is 1.1 - 0.64 kW/RT. It will be great if you can find the latest report and update the values from there.

Lastly, there seems to be a glitch on naming  "Chiller-air-air" and "Chiller-water/air", could you please change "Chiller-air-air" to "Chiller-air/air" in the database and the corresponding scripts?

emanuel030:
Thank you for the hint towards the BCA database. I have integrated he efficiency averages into the database. Chillers are renamed as well.

JIMENOFONSECA:
@emanuel030 Congrats! your first contrib. To CEA, a T-shirt is awaiting for you, so a place in the Website! 
C.C. @shanshanhsieh

emanuel030:
Thanks a lot!

Issue #2675, opened 2020-03-28
Solar radiation normalization fails


kerembeygo:
**Describe the bug**
Solar radiation normalization fails

**To Reproduce**
Demand forecasting -> Building solar radiation

Dashboard -> Add plot

Solar potentials -> Solar radiation

Next -> Normalization

Something went wrong!

**Expected behaviour**
Normalized solar radiation graph

**Screenshots**
![1](https://user-images.githubusercontent.com/29708612/77827935-ab508480-7129-11ea-85d9-dc16ca030f32.jpg)
![2](https://user-images.githubusercontent.com/29708612/77827938-aee40b80-7129-11ea-9944-12415ec6fcd1.jpg)
![3](https://user-images.githubusercontent.com/29708612/77827939-b1466580-7129-11ea-9d16-405624741681.jpg)
![4](https://user-images.githubusercontent.com/29708612/77827940-b3a8bf80-7129-11ea-91b3-36458f5fb851.jpg)

**Hardware (please complete the following information):**
 - OS and version: Windows 8.1
 - CEA 3.0.2

reyery:
@kerembeygo could you try opening CEA Console and run `cea dashboard` so we could see the output log of CEA. After seeing "**start socketio.run**" in the console, launch the Dasboard GUI. Try to reproduce the bug and see if there are any error messages in the console.

kerembeygo:
![1](https://user-images.githubusercontent.com/29708612/78424347-90c06300-7675-11ea-9762-ee39034d1533.jpg)

kerembeygo:
![2](https://user-images.githubusercontent.com/29708612/78424358-a170d900-7675-11ea-9b21-cfaa3488f251.jpg)

kerembeygo:
![3](https://user-images.githubusercontent.com/29708612/78424374-b3eb1280-7675-11ea-826c-873b566581ac.jpg)

kerembeygo:
![4](https://user-images.githubusercontent.com/29708612/78424380-c36a5b80-7675-11ea-9f9d-bee39bb660a4.jpg)

kerembeygo:
![5](https://user-images.githubusercontent.com/29708612/78424403-ef85dc80-7675-11ea-820e-29c3eec3a6dd.jpg)

kerembeygo:
[...]

Issue #2648, opened 2020-03-18
Installed Capacities plot fails for system = _sys_today


daren-thomas:
Steps to reproduce:

- run `cea workflow --workflow district-heating-system`
- open GUI at scenario (`reference-case-open/baseline`, as created in previous step)
- navigate to Dashboard
- add plot, choose Category="Supply System", Plot="Installed Capacities"
- for parameters, choose scenario-name="baseline", system="_sys_today" (the default)
- expected a plot, but instead get the following message:

```
This plot can't be displayed because some input files are missing:

C:\Users\darthoma\Documents\CityEnergyAnalyst\projects\reference-case-open\baseline\outputs\data\optimization\slave\gen_sys\ind_today_Electricity_Requirements_Pattern.csv
Try running these scripts first:

Supply System Part II (centralized)
```

Tried re-running the optimization script, but the file is never generated.

reyery:
I have tried it and can confirm that I have the same issue. 

It seems that `Supply System Part II (centralized)` does not run any tool/analysis for `_sys_today_` but only for those created during optimization as part of the script, therefore the file would not be created.

One way to solve it is to maybe try to run `_sys_today_` with `cea/optimization/master/evaluation.py` which helps create the data needed.

https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/d695eac0f151a8f8570981f60829b93663d74e59/cea/optimization/master/evaluation.py#L94-L129

shanshanhsieh:
we will not create the plots for `_sys_today_` for the moment, but it is an upcoming feature.
this feature will require a *supply system simulator* tool

daren-thomas:
@shanshanhsieh I can't get this plot to run for any system:

- Installed Capacities for sys_89_0 - baseline:

```
Traceback (most recent call last):
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask_restplus\api.py", line 584, in error_router
    return original_handler(e)
[...]

PR #2691, opened 2020-04-15
Block _sys_today_ supply plots

reyery:
As discussed in #2648, this PR prevents `_sys_today_` supply plots from being created for now until we are able to implement/extract the function to evaluate supply systems separately (instead of only in optimization). This PR also include some changes in plots with time frames as mentioned #2675.

Testing
-
_* Remember to clear `.cache` folder_

Opening any **supply system** plots with `_sys_today_` as the `system` parameter would show this message instead of an error message
image

Technology potential plots and solar radiation plots now indicate the time frame in the title which was missing previously. Monthly plots x-axis are now in _month_ _year_ format and Yearly plots x-axis only shows the _year_

image

image


daren-thomas:
This works as expected. Thank you!

Issue #2689, opened 2020-04-13
network-layout error, length mismatch


emanuel030:
When I run the network-layout script for one specific scenario, the following error is raised: 
```
`Traceback (most recent call last):
  File "C:/Users/emanuel/Documents/CityEnergyAnalyst_New/CityEnergyAnalyst/cea/technologies/network_layout/main.py", line 96, in <module>
    main(cea.config.Configuration())
  File "C:/Users/emanuel/Documents/CityEnergyAnalyst_New/CityEnergyAnalyst/cea/technologies/network_layout/main.py", line 92, in main
    layout_network(network_layout, locator)
  File "C:/Users/emanuel/Documents/CityEnergyAnalyst_New/CityEnergyAnalyst/cea/technologies/network_layout/main.py", line 65, in layout_network
    allow_looped_networks, optimization_flag, plant_building_names, disconnected_building_names)
  File "C:\Users\emanuel\Documents\CityEnergyAnalyst_New\CityEnergyAnalyst\cea\technologies\network_layout\steiner_spanning_tree.py", line 121, in calc_steiner_spanning_tree
    new_mst_nodes.loc[:, 'Name'] = names_temporary
  File "C:\Users\emanuel\Documents\CityEnergyAnalyst_New\Dependencies\Python\lib\site-packages\pandas\core\indexing.py", line 189, in __setitem__
    self._setitem_with_indexer(indexer, value)
  File "C:\Users\emanuel\Documents\CityEnergyAnalyst_New\Dependencies\Python\lib\site-packages\pandas\core\indexing.py", line 606, in _setitem_with_indexer
    raise ValueError('Must have equal len keys and value '
ValueError: Must have equal len keys and value when setting with an iterable`
```
The problem is, the whole file is too big to share through normal channels (14.7 GB),
Here the building geometry and building properties folders as well as the first outputs of the DC network before the script shoes an error: 
[scenario_network_error.zip](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/4469114/scenario_network_error.zip)

It seems that "nodes.shp" and the temp file "nodes_buildings.shp" have different length. 
this is the end output of the nodes.shp file: 
```
294  294.0   ...        NONE
295  295.0   ...        NONE
296  296.0   ...        NONE
297  297.0   ...        NONE
298    NaN   ...       B1089
```
and that of  "nodes_buildings.shp" :

[...]

PR #2692, opened 2020-04-15
I2689 network layout

JIMENOFONSECA:
this solves #2689 @emanuel030 kindly test and let me know so we can merge this in master


lguilhermers:
I still see an error when trying to run the first stage of thermal network in our case study:

ValueError: There was an error while populating the nodes fields. One or more buidlings could not be matched to nodes of the networkTry changing the constant SHAPEFILE_TOLERANCE in cea to fix this

Emanuel also confirmed also getting the same message

JIMENOFONSECA:
@luis I will investigate further

JIMENOFONSECA:
tested by @emanuel030 working now

Issue #2687, opened 2020-04-13
_ensure_folder method in InputLocator class may create race condition when used with multiprocessing


nasty-bos:
**Describe the bug**
While running "cea photovaltaic" within workflow for the districts containing multiple buildings of exactly same kind, simulation would often crash at the beginning of writing the results (BuildingName_PV.csv) into _outputs\\data\\potentials\\solar_ folder. 

**To Reproduce**
Difficult to reproduce as conditions require that the calculation for 2 buildings is finished simultaneously (which was often the case in my scenarios since the districts contain same buildings).

**Expected behaviour**
I would expect a folder to have been created before multiprocess is allowed to write into it i.e. 

```
    if not os.path.exists(locator.solar_potential_folder()):
        os.mkdir(locator.get_solar_radiation_folder())
```
Alternatively, one could catch the error in the "_ensure_folder" method and proceed with remaining execution after a short pause (100ms) and one final check for existence of the folder i.e. 

```
    @staticmethod
    def _ensure_folder(*components):
        """Return the `*components` joined together as a path to a folder and ensure that that folder exists on disc.
        If it doesn't exist yet, attempt to make it with `os.makedirs`."""
        folder = os.path.join(*components)
        if not os.path.exists(folder):
            try:
                os.makedirs(folder)
            except WindowsError as e:
                if "[Error 183] Cannot create a file when that file already exists" in e.__str__():
                    print("Error in creating [%s] - folder was created before I could create it")
                    time.sleep(0.1)
                    assert os.path.exists(folder), \
                        "Folder [%s] still not created or was deleted - aborting"
                else:
                    raise e
        return folder

```

**Screenshots**

```
Using 7 CPU's
calculating solar properties done
gathering properties of PV panel
calculating solar properties done
gathering properties of PV panel
calculating solar properties done
gathering properties of PV panel
calculating solar properties done
gathering properties of PV panel
calculating solar properties done
gathering properties of PV panel
calculating solar properties done
gathering properties of PV panel
calculating solar properties done
gathering properties of PV panel
filtering low potential sensor points done
calculating optimal tile angle and separation done
generating groups of sensor points done
filtering low potential sensor points done
calculating optimal tile angle and separation done
generating groups of sensor points done
filtering low potential sensor points done
calculating optimal tile angle and separation done
generating groups of sensor points done
Traceback (most recent call last):
[...]

PR #2698, opened 2020-04-16
added a fix for _ensure_folder method in /cea/inputlocator.py

nasty-bos:
As discussed in issue #2687 this fix attempts to catch the situation where multiple processes attempt to create the same folder simultaneously. The fix catches a generic OSError, forces a time.sleep of 500ms and checks if the folder was indeed created. 

Since the error is difficult to catch, the following test script can be used (script might need to be run several times in order to catch the error).

```
import os
import time
from multiprocessing import Pool

THIS_DIR = os.path.dirname(os.path.realpath(__file__))


##################################################
def return_folder_name(_):
    folder = os.path.join(THIS_DIR, "test_folder")
    if not os.path.exists(folder):
        try:
            print "making a folder"
            os.mkdir(folder)
        except OSError as e:
            print "failed to make a folder"
            time.sleep(0.5)
            if not os.path.exists(folder):
                raise e
    return folder


##################################################
def main():
    folder = "test_folder"
    if os.path.exists(folder):
        os.rmdir(folder)
    p = Pool(4)
    p.map(return_folder_name, [None for i in range(4)])
    p.close()


##################################################
if __name__ == "__main__":
    for i in range(250):
        print "main is at iter [%d]" %i
        main()

```

### DIFF
--- a/cea/inputlocator.py
+++ b/cea/inputlocator.py
@@ -4,8 +4,7 @@ inputlocator.py - locate input files by name based on the reference folder struc
 import os
 import shutil
 import tempfile
-
-import cea.config
+import time
 
 __author__ = "Daren Thomas"
 __copyright__ = "Copyright 2017, Architecture and Building Systems - ETH Zurich"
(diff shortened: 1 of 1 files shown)


daren-thomas:
@nasty-bos welcome to the CEA Collaborators team! I'll add you to the `CREDITS.md` file for Release 3.1.0 :)

daren-thomas:
@nasty-bos btw: you now have write access to our repo, so you don't _have_ to work on your own fork of the CEA if you don't want to.

nasty-bos:
@daren-thomas thanks for the warm welcome! :)

PR #2701, opened 2020-04-17
extended system efficiency calculation to all building scale cooling systems

emanuel030:
The elaborate heatpump model for the electricity load calculations of space cooling was replaced by a simplified based on the system efficiency from the database. This adds consistency when comparing against results for District cooling and allows to distinguish between different building scale cooling technologies.


daren-thomas:

PR #2705, opened 2020-04-20
update naming of database

emanuel030:
updated the naming of cooling technologies.
added a SG source (BCA benchmarking) for District cooling efficiency


daren-thomas:

PR #2706, opened 2020-04-22
Fix epwreader

reyery:
This fixes the bug with **epwreader** due to a missing variable `result`. This also adds normalization calculation to **Annual Cost** plot that was missing.

### DIFF
--- a/cea/utilities/epwreader.py
+++ b/cea/utilities/epwreader.py
@@ -95,7 +95,8 @@ def calc_wetbulb(Tdrybulb, RH):
 def main(config):
     locator = cea.inputlocator.InputLocator(scenario=config.scenario)
     # for the interface, the user should pick a file out of of those in ...DB/Weather/...
-    epw_reader(weather_path=(locator.get_weather_file()))
+    epw_data = epw_reader(weather_path=(locator.get_weather_file()))
+    print(epw_data)
 
 
 if __name__ == '__main__':
--- a/cea/utilities/epwreader.py
+++ b/cea/utilities/epwreader.py
@@ -8,7 +8,6 @@ import cea.inputlocator
 import numpy as np
 from cea.constants import BOLTZMANN, KELVIN_OFFSET, HOURS_IN_YEAR
 from calendar import isleap
-from cea.constants import HOURS_IN_YEAR
 
 __author__ = "Clayton Miller"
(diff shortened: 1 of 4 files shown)


reyery:
@JIMENOFONSECA could you help me test if this still works with .epw files with the leap day removed i.e. try to run `epwreader`? Or do you have an .epw file that I could test with?

cea/utilities/epwreader.py
```diff
@@ -37,29 +38,28 @@ def epw_reader(weather_path):
...
-                  "no need of further action")
-            date_range = date_range[~((date_range.month == 2) & (date_range.day == 29))]
-            epw_data['date'] = date_range
-            epw_data['dayofyear'] = date_range.dayofyear
-    else:
-        if len(epw_data) != HOURS_IN_YEAR:
-            raise Exception('Incorrect number of rows. Expected {}, got {}'.format(HOURS_IN_YEAR, len(epw_data)))
-        epw_data['date'] = date_range
-        epw_data['dayofyear'] = date_range.dayofyear
+        epw_data = epw_data[~((date_range.month == 2) & (date_range.day == 29))].reset_index()
```
JIMENOFONSECA, line 47: This line fails, since for a file with a leap year but without that day and month, it will fail. I sent you privately a file with that characteristics to test

reyery: It should work already since I generate the datetime index based on the data from the epw and not using intervals previously.

daren-thomas:
@reyery while I think it works, it's a bit weird to still have 2016 as the year after slicing out a day... would it not make more sense to change the year to... say... YEAR+1 in such cases? This shows up in the plots, e.g. Solar Radiation plot...

reyery:
@JIMENOFONSECA is the year of the weather data important? such as do users use that as a form of comparison (2016 vs 2017)? since right now they do show up in the plots. If not we could just take it out.

JIMENOFONSECA:
Better keep the year.  Otherwise it would lead to confusion.
On 23 Apr 2020, 08:10 +0800, Reynold Mok <notifications@github.com>, wrote:

@JIMENOFONSECA<https://github.com/JIMENOFONSECA> is the year of the weather data important? such as do users use that as a form of comparison (2016 vs 2017)? since right now they do show up in the plots. If not we could just take it out.

Issue #2667, opened 2020-03-25
As a User I would like "Unnamed: 0" columns to be renamed to better understand their meaning


daren-thomas:
There are 17 instances of columns named "Unnamed:0" (including one "Unnamed:0.1" in the `schemas.yml` file. This is what those columns are called in the corresponding csv files. I believe this is an artefact of how they are created (`df.to_csv`) and it could probably be easily fixed.

Often, a more meaningful name could be "hour" or "hour-of-year".

shanshanhsieh:
we should change those instances to `index=False`

daren-thomas:
@shanshanhsieh I believe most of these are handled in #2637 - we will do a check (grep `schemals.yml`) after merging.

daren-thomas:
I noticed some discrepancies with the `thermal-network` script outputs, depending on if "simplified" or "detailed" model is used:

- `get_thermal_network_node_types_csv_file`:
  - detailed columns: `[,Name,Type,Building,coordinates,Q_hex_h_ahu,Q_hex_h_aru,Q_hex_h_shu,Q_hex_h_ww,Q_hex_plant_kW]`
  - simplified columns: `[Building,Type]`
  - NOTE: I think this is connected with issue #2700
- `get_thermal_network_edge_list_file`: detailed version includes `Name` column
- `get_thermal_network_plant_heat_requirement_file`: the only column in this file has the header `NONE`. That's hardly better than "Unnamed: 0" and should probably be renamed to... idk... `plant_thermal_load_kW` or something?
- `get_thermal_demand_csv_file`: detailed version has extra (unnamed) column for the hours

There is a separate issue for this: #2710

NOTE: This also means we'll need to update the `schemas.yml` descriptions for the columns currently missing...

daren-thomas:
The multi-criteria-analysis script also needs updating: #2711

daren-thomas:
The decentralized script also needs updating: #2712

shanshanhsieh:
@daren-thomas please quickly run through the workflow and check what is remaining.

reyery:
Solved in #2748

PR #2714, opened 2020-04-28
Small performance optimization for detailed thermal-network script

daren-thomas:
While working on #2667, I made a small performance optimization to the detailed version of the thermal-network script: The results of the `find_loops` method can be cached for a specific input df. This saves a bit of time (maybe 20%). It's still very slow and as far as I can tell, most of that is performing a `np.isclose` call. Those are not really avoidable, _unless_ we can figure out that logically some are redundant. That's an exercise for another day!

### DIFF
--- a/cea/datamanagement/database_migrator.py
+++ b/cea/datamanagement/database_migrator.py
@@ -124,7 +124,7 @@ def migrate_2_29_to_2_31(scenario):
     os.remove(occupancy_dbf_path)
     print("- removing invalid input-tables (NOTE: run archetypes-mapper again)")
     for fname in {"supply_systems.dbf", "internal_loads.dbf", "indoor_comfort.dbf",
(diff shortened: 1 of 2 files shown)


daren-thomas:
@martin-mosteiro would you mind taking a quick look at this so we can include it in our Release tomorrow?

martin-mosteiro:
@daren-thomas I'll check it out today.

martin-mosteiro:
Tested doing the following:
1. Extracted the `reference-case-open`
2. Ran the data management tools.
3. Ran the demand tool.
4. Ran the thermal network tool for DH for 1000 time steps using the current branch.
5. Ran the thermal network tool for DH for 1000 time steps using the master branch.

The results were exactly the same, but this branch took 25% less time than the master branch. Good job, @daren-thomas!

PR #2721, opened 2020-05-04
fixes issue 2073

gabriel-happle:
fixes wrong `log` spotted by user @compostable


gabriel-happle:
@daren-thomas how to test:

Create a comfort chart for a random building. See if it works.

Issue #2693, opened 2020-04-15
System_costs module doesn't include CapEx of all cooling systems


emanuel030:
The system_costs.py module currently only considers Capital Expenditure of District Cooling and disregards Capex for other cooling technologies.  

The cooling capacity QC_sys0_kW should serve as a basis to calculate investment cost for all technologies.

daren-thomas:
@shanshanhsieh can you please perform a triage for this issue?

shanshanhsieh:
hi @emanuel030 , 
are you referring to line 92 in `system_costs.py`?
line 92:
```cooling[service + '_sys_total_capex_yr'] = cooling[service + '0_kW'] * cooling['CAPEX_USD2015kW']```
in which the services are defined as:  `cooling_services = ['DC_cs', 'DC_cdata', 'DC_cre']`

please also indicate what kind of assistance would you need to tackle this issue.
thanks!

emanuel030:
Yes exactly! I already modified the code to make the investment cost dependent on the thermal capacity. I will have to check if it still works with the dashboard plots of costs before creating the pull request.

PR #2725, opened 2020-05-08
#2693 capex all cooling systems

emanuel030:
Issue #2693 could be resolved by inverting the 'capacity' of the final service, i.e. 'GRID_cs_0kW' to the capacity of the underlying machine, using the respective system efficiency. Further, the code was adapted to annualize both the operational and capital expenditures using a present value of cost (PVC) annualized costs (EAC) formula.
image


daren-thomas:

JIMENOFONSECA:
@daren-thomas how do I test CEA locally now? I see the dod is not anymore there

JIMENOFONSECA:
@daren-thomas I got a few cryptic errors:, mind running the CEA test --- workflow slow in your local machine? that is what I did and I got this.

The error happens toward the end while running the plots unitests I think.

thanks!

image

Issue #2700, opened 2020-04-17
Heating Reset Curve plot does not work with reference-case-open / standard workflow (simplified network model)


daren-thomas:
I'm trying to test all plots, but I can't seem to get the Heating Reset Curve plot to work. When trying to debug it, I found there is a difference in the file `get_thermal_network_node_types_csv_file` columns - depending on whether the simplified or the detailed network model is being used.

The following error was produced using the simplified network model and is also the way the file is described in `schemas.yml`.

```
Traceback (most recent call last):
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask_restplus\api.py", line 584, in error_router
    return original_handler(e)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask_restplus\api.py", line 584, in error_router
    return original_handler(e)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "C:\Users\darthoma\Documents\CityEnergyAnalyst\Dependencies\Python\lib\site-packages\flask\app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\interfaces\dashboard\plots\routes.py", line 50, in route_div
    plot_div = plot.plot_div()
  File "c:\users\darthoma\documents\github\cityenergyanalyst\cea\plots\base.py", line 168, in plot_div
    return self.cache.lookup_plot_div(self, self._plot_div_producer)
[...]

PR #2727, opened 2020-05-13
Thermal network script output format

### LINKED ISSUE #2710: Check to make sure the format of outputfiles for thermal-network is the same for simplified / detailed
I'm not sure this is the case - also, we haven't been testing both of these lately. Instead, we have been using the simplified model. 

See also issue #2700 (Heating Reset Curve plot does not work with reference-case-open / standard workflow (simplified network model))

reyery:
This PR fixes #2710 by trying to make the outputs of detailed and simplified versions of `thermal-network` to be consistent with each other. It also includes a fix for #2700 by correctly creating choices for plant nodes.

Testing for #2700 would be trying to select a plant node when creating **Heating Reset Curve** plot

### DIFF
--- a/cea/inputlocator.py
+++ b/cea/inputlocator.py
@@ -562,7 +562,8 @@ class InputLocator(object):
         if os.path.exists(nodes_csv):
             import pandas as pd
             nodes_df = pd.read_csv(nodes_csv)
-            return list(nodes_df[nodes_df['Type'] == 'PLANT']['Building'].values)
+            is_plant = nodes_df['Type'] == 'PLANT'
+            return list(nodes_df[is_plant]['Name'].values)
         return []
 
     def get_thermal_network_edge_list_file(self, network_type, network_name=''):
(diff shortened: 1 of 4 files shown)


reyery:
Some problems that were not fixed:

- Detailed version still has additional columns in `get_thermal_network_node_types_csv_file`:
['coordinates', 'Q_hex_h_ahu', 'Q_hex_h_aru', 'Q_hex_h_shu', 'Q_hex_h_ww', 'Q_hex_plant_kW']
(not sure if important)

- Detailed version creates additional files that are not created in the simplified version
    - `get_thermal_network_edge_node_matrix_file`
    - `get_nominal_edge_mass_flow_csv_file`
    - `get_nominal_node_mass_flow_csv_file`

shanshanhsieh:
@reyery 

- those columns are not used anywhere to my knowledge but might be good for future development. are those variables being documented? if yes, we can keep it like this. Subsequently, please make a new issue on matching these columns from the detailed script. thanks!

- these files are intermediate files that stores the sizing information. they are necessary for the detailed script because there is an option in the config to skip the sizing step and read this directly. we can just keep it.

daren-thomas:
The Heating Reset Curve plot now works with the simplified version. I guess that's good enough to merge!

Issue #2670, opened 2020-03-25
Building Solar doesn't work with underground parkings 


AlexJew:
**Describe the bug**
I have underground parkings that I have defined with _height_ag = floors_ag = 0_ . Even though they won't represent any heat/cooling consumption, they represent an electricity consumption (electric car chargers will be installed) and therefore I need to keep them. 

However, when trying to run **Building Solar Radiation** tool, the program crashes :

1. I run the tool without including them in the list of buildings, and I get the following error : 
```
    floor_to_floor_height = height / nfloors
ZeroDivisionError: float division by zero
```

2. I set _floors_ag = 1_ and _height_ag = 0_, and run the tool without including them in the list of buildings, I get the following error :
```
RuntimeError: Standard_ConstructionError
```

My underground parking geometries overlap with buildings. What could be the solution? **Building Solar Radiation** run completely when excluding the parkings.

**To Reproduce**
1. Download the input and database files from : https://www.dropbox.com/transfer/AAAAAKdzY9szzW_T5157yRChF5m8zOSiR7P0tIGY_OiU5yXtq-Iba_E
2. Create new scenario, import these files as input
3. After running **Archetypes Mapper** and **Weather Helper**, run **Demand Forecasting > Building Solar Radiation**

**Expected behaviour**
Successful running with **Building Solar Radiation** taking into account only the selected buildings and the surroundings

**Screenshots**

**Hardware (please complete the following information):**
 - OS and version: Windows 10, 1909
 - Version of CEA: 2.32

shanshanhsieh:
the radiation toold should skip those buildings have height_ag = floors_ag = 0
but a dummy radiation file should be created to pass on to the demand calculation

AlexJew:
Hi, I don't know if that is of any help, but my terrain isn't flat, so I guess that, even if the parkings are considered as underground, they may clash with lower standing building (terrain-wise) that are purely overground.

daren-thomas:
@JIMENOFONSECA do you have an update on this issue?

JIMENOFONSECA:
No, i am not sure if another developer can help. i am swamp these months with other tasks.
On 30 Apr 2020, 15:33 +0800, Daren Thomas <notifications@github.com>, wrote:

@JIMENOFONSECA<https://github.com/JIMENOFONSECA> do you have an update on this issue?

JIMENOFONSECA:
[...]

PR #2729, opened 2020-05-13
CEA cannot work with only underground structures

JIMENOFONSECA:
this provides a restriction to #2670 @rey mind checking if some of these restrictions are actually happening one way or another. it is surprising that the user was able to run CEA since the restriction in cea.datamanagement.data_verfication should not have allowed for this


rey:
@JIMENOFONSECA sorry dude, I don't think I'm qualified.

Issue #2732, opened 2020-05-14
Refinement of the 'chiller_vapor_compression' COP by accounting for part load operation


emanuel030:
For my research it is imperative to calculate the system efficiency of vapor compression chillers with high accuracy.

I therefore suggest to improve the COP calculation for vapor compression chillers, by introducing a part load factor (PLF) based on the installed capacity.

The end-user of this feature would be researchers as well as practitioners.

The complete equation of the proposed improvement of COP calculation is described in the word document attached and is based on empirical formulas taken from https://comnet.org/index.php/382-chillers. 

[explanation_part_load.docx](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/4626482/explanation_part_load.docx)

The implementation further includes a design of the VCC chillers (sizing&compressor type) according to ASHRAE 90.1 Appendix G, based on the scale ('Building'/'District') and the design capacity.

PR #2733, opened 2020-05-14
#2732 - VCC Partial load COP

emanuel030:
I have implemented the partial load factor feature described in #2732 with a new module 'load_distribution' in the technologies subfolder of cea.  It allows for a design of the vapor compression chillers (sizing/ kind of compressor) to improve the accuracy of the COP. The script has been tested for both the decentralized and the centralized optimization and gives the expected results shown below:
For decentralized VCC:
image

centralized VCC: 
image

In the process i have found that there is a missing capacity constraint for water source VCC for the allocation of Q_cooling_unmet_W over the different technologies. I have therefore adapted 'cooling_resource_activation' module accordingly.

Please let me know if you have any questions regarding the implementation.


daren-thomas:

shanshanhsieh:
@emanuel030 could you please resolve the conflict? (see above)
there are three lines inside the simplified_thermal_network.py are modified, you have to go into the code, choose the lines you want to keep, and then commit.

as you also mentioned that multiple fixes were done in this PR.
could you please also list them out here so we can keep track on all the changes?
thank you!

emanuel030:
@shanshanhsieh here is a list of what was changed:
- addition of 'load_distribution' module and all changes in higher modules to pass along the arguments in functions for it to work.
- fix of missing capacity constraint for water source VCC for the allocation of Q_cooling_unmet_W over the different technologies in 'cooling_resource_activation' module
- fix of the required pump electricity for DCS in module 'pump' due to high overestimation of pump electricity caused by aggregated losses and aggregated mass flows over all building. The output of direct power requirement output from the simplified network simulation was used instead.
- system COP as an additional result output for each configuration of cooling supply in 'decentralized_buildings_cooling' module

shanshanhsieh:
@emanuel030 i will start testing this PR now.
i also notices that there are some updates from yesterday, is it complete?

cea/technologies/load_distribution.py
```diff
@@ -0,0 +1,113 @@
+import numpy as np
```
shanshanhsieh, line 1: it seems like this load distristribution method is specific to VCC.
is there a reason why it is a saperated script?

emanuel030:
@shanshanhsieh it is complete now and working for multiple case study districts. The code is not the most compact and could be improved by aggregating the variables passed down to 'load_distribution' from 'cooling_main' in a VCC class.

cea/technologies/load_distribution.py
```diff
@@ -0,0 +1,113 @@
...
+        q_e = -0.00028254
+        q_f = 0.00052346
+
+    t_chws_F = kelvin_to_fahrenheit(T_chw_sup_K)
+    t_cws_F = kelvin_to_fahrenheit(T_cw_in_K)
+
+    available_capacity = rated_capacity * (q_a + q_b*t_chws_F + q_c*t_chws_F**2 + q_d*t_cws_F + q_e * t_cws_F**2 + q_f*t_chws_F*t_cws_F)
+    return available_capacity
+
+def kelvin_to_fahrenheit(T_Kelvin):
```
shanshanhsieh, line 97: move this function to utilities/physics.py

cea/technologies/load_distribution.py
emanuel030: No particular reason. I will transfer it into the VCC module once the functions are in physics.

shanshanhsieh:
@emanuel030 , sorry for the tiny conflict with the commit daren made a few seconds ago...
we can have a quick chat today/tomorrow on how to create test functions. 
please reach me via cea slack : )

shanshanhsieh:
@emanuel030 please give me a pin whenever this PR is ready to be reviewed again.
(no rush : )

shanshanhsieh:
both cooling/heating case workflows ran through, the unittests are working.
this PR is good to go!

PR #2740, opened 2020-05-21
Run code coverage on CEA test

### LINKED ISSUE #2715: Run coverage on the slow testing workflow to make sure we're testing all code paths
This is important so we know what code paths are acutally being tested on a regular basis and which ones might be "dead code". We need this information along with sample output from such a comprehensive workflow test when we port to Python3, so we know that we're actually creating the same product.

- coverage (i think this is the name of the tool) runs a python script and reports on how much (%) of the script is actually being used and points out the branches (like `if / else`) that don't get used.

### LINKED ISSUE #2730: Inconsistent/duplicated information in schemas.yml
Since we have merged the `glossary.csv` into `schemas.yml`, it introduced a new property `values` which is used to describe the example values found in that column. 

I have realized that numerical types i.e. **int** and **float** are using this property to describe the expected range of the variable.

For example, `1ST_USE_R` expects values between **0.0** and **1.0** based on the property `value`. 
However, this only provides a description for the values but nothing is enforced since ranges for numerical types are enforced with the `min` and `max` property (I don't think we are parsing the string in `values` if I remember correctly).
```
1ST_USE_R:
        description: Fraction of gross floor area for first Use Type
        type: float
        unit: '[m2/m2]'
        values: '{0.0...1.0}'
```
This in turn creates opportunities for inconsistencies in the schema as `values` that do not reflect the range properties.
```
O&M_%:
          description: operation and maintanence cost factor (fraction of the investment
            cost)
          min: 0.0      ^^^ // Should also include `max` property of 100 here
          type: float
          unit: '[%]'
          values: '{0.0...100}' 
```

Because of this duplication of information, we could potentially do away with the `min` and `max` properties and parse `values` instead?

reyery:
This PR resolves #2715 by running coverage on `cea test` with the slow workflow. 
**This would require the user to have `coverage` installed for this to work.
Currently we have **73%** code coverage, meaning that 73% of our code in the slow workflow is being run without any errors.

Testing:
- Make sure `coverage` is installed i.e. `pip install coverage`
- Run `cea test --workflow slow` in the CEA console (might take a while)
- Should show coverage report after test finishes
- A coverage report would also be created in the indicated path and can be opened with the `index.html`

This also fixes #2730 by cleaning up schemas.yml and implementing checks for consistency of numerical ranges in unit test. Currently, the checks will only print the errors and not fail the unit test. There are still some values with 'TODO's especially for the weather file.

Testing:
- Run **main** for `cea.test.run_unit_tests`
- Could try changing some range values for numeric types and run the unit test again and it should show up in the output console


cea/tests/__init__.py
```diff
@@ -35,8 +38,22 @@ def main(config):
         # make sure we're working on a clean slate
         shutil.rmtree(default_config.project)
 
+    cov = coverage.Coverage()
```
daren-thomas, line 41: wait... you hardcoded coverage into the app? shouldn't it be possible to just do ` coverage run cea/tests/__init__.py --workflow slow`? (actually. I just tested it. it _is_ possible...)

Can we at least make this an option in the config? Coverage slows down the testing - we don't really want to do that all the time...

cea/schemas.yml
```diff
@@ -10,6 +10,7 @@ PVT_metadata_results:
         type: float
         unit: '[m2]'
         values: '{0.0...n}'
+        min: 0.0
```
daren-thomas, line 13: woah. i hope you didn't have to do this by hand...

setup.py
```diff
@@ -18,6 +18,7 @@
     LONG_DESCRIPTION = f.read()
 
 INSTALL_REQUIRES = ['SALib==1.2',  # last version to work with python2
+                    'coverage',
```
daren-thomas, line 21: I think we can remove this dependency if we remove coverage from the code itself - any dev wanting to run coverage can just pip install it. btw: did you see pycharm also has coverage functions? right next to the profiling thing. did you check those out?

cea/tests/__init__.py
reyery: Alright, in that case, I think I will take it out from `cea test` and instead update the docs on how to run `coverage` for the tests.

setup.py
reyery: Oh I've seen that before but have always thought that coverage needs to be installed to run it. I didn't try it before.

cea/schemas.yml
reyery: nah once I realized that most of the numerical types didn't have proper range properties, I knew I had to write a small script to do that. Hmm but should I have included that as a utility script? I guess that would be nice.

setup.py
daren-thomas: yes, i think coverage needs to be installed first - i haven't run it either before - but still: that's not something the cea _requires_ so we don't need it in the INSTALL_REQUIRES.

cea/schemas.yml
daren-thomas: nah, i think you're good: we can expect this to be filled out correctly from now on (since you added the unit test, right?) and won't have to do it again.

PR #2748, opened 2020-05-27
2667 clean up unnamed in schemas yml

daren-thomas:
`schemas.yml` now does not contain any "Unnamed: 0" files anymore. To test: run `cea test --workflow slow` and also search schemas.yml for "Unnamed"


reyery:
Tested and double checked output files. Works as described.

PR #2754, opened 2020-06-01
Prevent scenario name from being empty string

### LINKED ISSUE #2549: `scenario-name` parameter in config should not accept empty/null values
Problem
-
Currently the GUI sets `scenario-name` to an _empty string_ when doing some operations regarding projects using the dashboard server endpoints:
- **create new project** - no scenario exists yet
- **open project** - we do not assume which scenario to open for the user
- **delete project/scenario** - no scenario exists anymore

In the GUI, the user is restricted from doing anything, such as running scripts, when `scenario-name` is an _empty string_ by blocking navigation to such functions since it would undoubtedly cause issues.

However this logic only exists in the front end and not in other interfaces such as **cli** or **grasshopper** which might cause issues when using them to run scripts. The _empty string_ generates a valid `scenario-path` (i.e. the project folder) and causes scripts to look for files in the project folder instead where it will definitely not find them.

Proposed solution
-
The config file should be able to accept a scenario name as a valid input which might not exist in the project path  but not an empty / `null` value, since it does not make sense for it to be an empty value (wish I realized that earlier when making the GUI). With a non empty `scenario-name`, we could use it to check if the scenario path exists before running script while an empty `scenario-name` would just give us a false positive.

Therefore, for scripts we could add to check if the scenario path exists before running it, which could save time tracing missing input files and provide a more useful/explicit error message instead of listing them like this 
```
ERROR: Missing input files:
- C:\CEA PROJECTS\reference-case-open - demo\inputs\building-geometry\zone.shp
- C:\CEA PROJECTS\reference-case-open - demo\inputs\building-properties\age.dbf
- C:\CEA PROJECTS\reference-case-open - demo\inputs\building-properties\occupancy.dbf
```
(not apparent that `scenario-path` does not exist just by reading that)

As for the GUI, it will have to use some kind of state within itself to support such operations and **not** use the config file and the empty string approach.

reyery:
This PR resolves #2549 by preventing **ScenarioNameParameter** from being set to an empty string. It also updates project endpoints of the dashboard to only write to config file when being asked to explicitly save it using the `PUT` method. This allows the GUI to not save to config when opening/creating projects, only when opening/creating a scenario.


reyery:
To be tested with https://github.com/architecture-building-systems/CityEnergyAnalyst-GUI/pull/27
(Build: https://we.tl/t-sFr3DITwNI)

Testing:
-  Check `project` and `scenario` in cea.config
- Try opening/creating project. It should work as intended. `project` and `scenario` in cea.config should not change
- Try opening/creating scenario. It should works as intended.  `project` and `scenario` in cea.config should change to scenario being opened/created

shanshanhsieh:
@daren-thomas could you please test this? thanks!

cea/interfaces/dashboard/api/project.py
```diff
@@ -22,92 +22,89 @@
...
-                if 'scenario' not in payload:
-                    config.scenario_name = ''
-                    config.save()
-                    return {'message': 'Project path changed'}
-            else:
-                abort(400, 'Path given does not exist')
+        project_path = request.args.get('path')
+        if project_path is None:
+            config = current_app.cea_config
+            scenario = config.scenario_name
```
daren-thomas, line 45: I think it's a bit confusing to use the word `scenario` when referring to the `general:scenario-name` parameter - let's rename this to `scenario_name`. That way, `scenario` is always used for the string ` "{general:project}:{general:scenario-name}"`

cea/interfaces/dashboard/api/project.py
```diff
@@ -22,92 +22,89 @@
...
     'path': fields.String(description='Path of Project'),
-    'scenario': fields.String(description='Path of Scenario')
 })
 
-PROJECT_MODEL = api.inherit('Project', PROJECT_PATH_MODEL, {
-    'name': fields.String(description='Name of Project'),
-    'scenario': fields.String(description='Name of Current Scenario'),
-    'scenarios': fields.List(fields.String, description='Name of Current Scenario')
+SCENARIO_PATH_MODEL = api.inherit('Scenario Path', PROJECT_PATH_MODEL, {
+    'scenario': fields.String(description='Name of Scenario')
```
daren-thomas, line 28: let's call this `scenario-name` instead, no?

cea/interfaces/dashboard/api/project.py
```diff
@@ -22,92 +22,89 @@
...
+                'scenarios': list_scenario_names_for_project(config)}
 
-    @api.doc(body=NEW_PROJECT_MODEL, responses={200: 'Success'})
     def post(self):
-        """Create new project"""
-        config = current_app.cea_config
-        payload = api.payload
+        """Create new project folder"""
+        name = api.payload.get('name')
+        project_path = api.payload.get('path')
```
daren-thomas, line 60: this is called `project-root` in other parts of the cea...

cea/interfaces/dashboard/api/project.py
```diff
@@ -22,92 +22,89 @@
...
+            abort(400, 'Parameters not valid - Project Name: {name}, Project Path: {project_path}'.format(
+                name=name, project_path=project_path
+            ))
 
-            return {'message': 'Project created at {}'.format(project_path)}
+    @api.expect(SCENARIO_PATH_MODEL)
+    def put(self):
+        """Update Project info in config"""
+        config = current_app.cea_config
+        path = api.payload.get('path')
```
daren-thomas, line 79: why aren't we calling this `project-path` or `project`? that way, we use the same names as in the config files...

cea/interfaces/dashboard/api/project.py
```diff
@@ -22,92 +22,89 @@
...
+        scenario = api.payload.get('scenario')
+
+        if path is not None and scenario is not None:
+            # Project path must exist but scenario does not have to
+            if os.path.exists(path):
+                config.project = path
```
daren-thomas, line 85: this is what i mean... it should read:

```python
config.project = project
config.scenario_name = scenario_name
```

cea/interfaces/dashboard/api/project.py
```diff
@@ -22,92 +22,89 @@
...
 
 @api.route('/scenario/')
 class Scenarios(Resource):
     def post(self):
         """Create new scenario"""
-        config = current_app.cea_config
-        payload = api.payload
-        new_scenario_path = os.path.join(config.project, payload['name'])
+        config = cea.config.Configuration()
+        project_path = api.payload.get('projectPath')
```
daren-thomas, line 102: wait... this is yet another name for `project_path`?

cea/interfaces/dashboard/api/project.py
```diff
@@ -117,57 +114,52 @@ def post(self):
...
-                                               scenario=new_scenario_path,
-                                               zone=files['zone'],
-                                               surroundings=files['surroundings'],
-                                               streets=files['streets'],
-                                               terrain=files['terrain'],
-                                               typology=files['typology'])
-
+                                                scenario=new_scenario_path,
+                                                zone=files.get('zone'),
+                                                surroundings=files.get('surroundings', ''),
```
daren-thomas, line 133: just out of curiosity... could the `''` also be `None`? I don't remember anymore, but I think that would be the "cleaner" version, no?

reyery:
Yes. It definitely makes sense to have the names be consistent for readability. I shall change the parameters to reflect what is in config. i.e. `project` for path of project and `scenario_name` for name of scenario.

cea/interfaces/dashboard/api/project.py
reyery: Currently **`PathParameter`** is unable to encode `None`, it would just cast it to a **`str`**. And `create_new_scenario` checks if it is an empty string for the path parameters as it decodes to a **`str`**. 

We could override the encode method to parse `None` to an empty string.

reyery:
> To be tested with [architecture-building-systems/CityEnergyAnalyst-GUI#27](https://github.com/architecture-building-systems/CityEnergyAnalyst-GUI/pull/27)
> (Build: https://we.tl/t-sFr3DITwNI)
> 
> Testing:
> 
>  * Check `project` and `scenario` in cea.config
> 
> * Try opening/creating project. It should work as intended. `project` and `scenario` in cea.config should not change
> 
> * Try opening/creating scenario. It should works as intended.  `project` and `scenario` in cea.config should change to scenario being opened/created


In addition to these tests, try importing a zone geometry when creating a new scenario to make sure `create_new_scenario` still works.

daren-thomas:
@reyery this is _nearly_ ready to merge - I _did_ find one small bug (by accident). it seems, scenario names may not start or end with whitespace. try it: create a new scenario, call it `"abc "` (notice the space at the end) and try to create it. it won't work, because the config file stores the scenario name with `scenario_name.strip()`.

Could you add that small change please?

PR #2762, opened 2020-06-09
2761 Fix network layout error

emanuel030:
the error could be fixed by increasing the SAPEFILE_TOLERANCE from 6 to 15. An additional comment was made in the ValueError message to hint towards changing the tolerance if an error is encountered.


daren-thomas:

emanuel030:
The adjustment in SHAPEFILE_TOLERANCE alone did not solve the issue! I still had to manipulate the shapefile manually in ArcMap with the the 'Feature to Line' tool (xy-tolerance of 10m) to solve the issue for multiple scenarios. I therefore added this solution as comment in the ValueError message. All erroneous streets.shp files were generated with the streets-helper tool (so from OpenStreetMap).

cea/constants.py
```diff
@@ -18,7 +18,7 @@
 
 
 #Shapefiles precision of decimals
-SHAPEFILE_TOLERANCE = 6 #this is precision of milimiters
+SHAPEFILE_TOLERANCE = 20 #this is precision of milimiters
```
shanshanhsieh, line 21: this value is not consistent with the description in the PR ('15' was indicated there)
also, is it necessary to change the default from 6 to 15? 
is it a common case that tolerance is exceeded?
if yes, please update the documentation to describe what does '15' means (e.g., 6 is the precision of millimeters)

emanuel030: @Shanshanhsieh  I have tested multiple case studies and with 20 mm of SHAPEFILE_TOLERANCE they were no errors, while with 6mm of SHAPEFILE_TOLERANCE there were. I have talked to Jimeno about this and he said there is no harm in increasing the SHAPEFILE_TOLERANCE by this much. Rather a too small Tolerance would be problematic since for more case studies this would throw and error.

shanshanhsieh:
testing the workflow now, will be merged shortly!

PR #2799, opened 2020-07-02
2751 cleanup hardcoded values

emanuel030:
I have worked together with @reyery to accomplish the following:

A class called VaporCompressionChiller was created in the chiller_vapor_compression module to reduce the arguments being passed down in the optimization scripts. This influences both decentralized and centralized optimization. The CONVERSION database has been extended to include the G-Value (isentropic efficiency) of the chillers in sheet 'Chillers', as well as a new sheet called 'Chiller_configuration' was created to store all values relevant for the part load factor calculations.

The unit test for chiller_vapor_compression has been adjusted accordingly.


daren-thomas:

PR #2823, opened 2020-08-25
Origin/i2454 validation tool updated

lguilhermers:
I have updated the code to a version compatible with the master branch and has been functioning well for me (on the old version it has been for a while, on the new version I did some successful tests today). Which means it is probably ready for a review :)

In order for the code to work, one must have a folder in inputs that relates to the energy measurements. Please let me know if you want some case studies that contain some of those real buildings with measurements to test. The code should also work for multiple scenarios that are inside the same project folder (but it will consider only the ones that have the measurements file).

I think the code validation.py is also useful by itself, if someone just want to compare their model with measurements based on ASHRAE standards, but I still haven't tested if it works as a standalone tool.


lguilhermers:
here you can find 2 scenarios, with 5 buildings to be calibrated (4 on Yushun, 1 in Bras Basah)
[scenarios_to_test.zip](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/5122495/scenarios_to_test.zip)

shanshanhsieh:
hi @lguilhermers ,
we will discuss this PR at the next scrum meeting (3 Sep), please do not merge it yet.
thanks!

daren-thomas:
the folder `examples/reference-case-open` should just be the zip file - it seems this commit extracted the zip file to that folder...
@lguilhermers would you mind contacting me to set up a meeting for creating the plugin?

Issue #2815, opened 2020-08-11
Surroundings helper does not capture every building


reyery:
When the zone geometry bounds are of irregular shape as one like below, there is a huge chunk of buildings that are not considered as highlighted in red.

![image](https://user-images.githubusercontent.com/34395415/89852857-6140da80-dbc2-11ea-8809-2e60374ba311.png)

martin-mosteiro:
I have the same issue here:
![Screen Shot 2020-08-20 at 13 10 01](https://user-images.githubusercontent.com/18548065/90763196-892cee00-e2e6-11ea-9984-9d1c86ffb561.png)

This is a duplicate of #2456 though – Which we closed because (a) no one was blocked by this issue, (b) we had no capacity to solve this issue, (c) we had a workaround. This *should* be solved eventually, though.

daren-thomas:
@reyery good job! I think maybe you need to update the estimate on this one?

reyery:
Closed by #2824

PR #2824, opened 2020-08-27
Change surroundings-helper to use a concave instead of a convex hull

reyery:
This PR solves #2815 by extract more relevant surrounding buildings based on zone buildings.

The current implementation uses a convex hull to determine the boundary of the zone buildings which does not accurately represent or follow closely to the actual boundary of the zone buildings. This leads to spaces where surrounding buildings are not extracted for zone areas of irregular shape e.g. Basel. See picture below. (Red represents zone while orange represents surrounding buildings)

This is actual not ideal for the radiation simulation since that is the main use for the surrounding buildings, for calculating shading due to nearby buildings. Having surrounding buildings that are so far away from the zone buildings does not make sense since it would likely not affect the radiation on zone buildings.

This new implementation would use a concave hull to determine the boundary instead, which follows more closely to the edges of zone buildings near the edges of the site. This results in buildings that are closer to the zone buildings which is more ideal. I found out that the function actually already exists in the code but was not used. I did some minor changes to the code to make use of it and this is what I got. (Purple represents surrounding buildings extracted using the new method.)

However, one caveat is that this does make it run a little slower, e.g. for the city of Zurich, the current took around 130s while this took around 210s.

image


daren-thomas:
we will merge this after python3 port

daren-thomas:
I'm not sure this is the expected behavior:

image

shanshanhsieh:
@reyery what is the current status?

daren-thomas:
@reyery any progress?

reyery:
@daren-thomas sorry I'm still working on it. Still finding out ways to do it automatically.

reyery:
This should work now. Right now it uses a spatial join instead of an concave shell to filter the zone buildings from the extracted geometries. It is more accurate (for smaller sites with irregular shape like multiple acute angles) but less efficient. 

Zurich example now takes around 500~ seconds which is at least 3 times slower than the previous method (scales exponentially based on the number of total buildings). But at this point it is a trade off that is acceptable.

daren-thomas:
woah! very nice! works as advertised. good job :-) (I also like how it's _way_ less code!)

Issue #2794, opened 2020-06-29
Ship Reynolds Daysim binaries instead of the official ones


daren-thomas:
@reyery has a new build of the Daysim binaries for x64 and this build avoids a Memory issue with large scenarios.

reyery:
Just found out that there are discrepancies in the outputs for **radiation** when using the different Daysim binaries.
It seems that the newly complied binaries actually runs **faster**, but the outputs are always **different** 
(based on SHA-1 hash of the output folder)

```
Complied Daysim binaries from Daysim repo master
58231D6057F1FEF254C1DDD1DDB479C249B9CD78 69.78s
4953908B20B24FF2587404DB82DDCF31BE7C85C8 69.11s
B8BA47F88AAFF7EB6D68ACE0BB8920C20052E598 69.29s

Daysim binaries from installation
6BFA6479D927DA7B7ECCBAABDC9DC506D0FE5AA5 117.89s
6BFA6479D927DA7B7ECCBAABDC9DC506D0FE5AA5 118.69s
6BFA6479D927DA7B7ECCBAABDC9DC506D0FE5AA5 117.99s
```

Just floating point differences I think, I uploaded some examples for the same one building.

[B1000_radiation.zip](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/4859290/B1000_radiation.zip)

```
Example A
2009-01-01 08:00:00,5.82,1508.63,0.38,197.66,1.79,1093.92,1.86,1093.94,0.35,197.53,0.28,149.11,1.36,825.24,1.41,825.25,0.26,149.01
2009-01-01 09:00:00,101.58,1508.63,22.71,197.66,24.18,1093.92,90.69,1093.94,4.55,197.53,17.05,149.11,18.88,825.24,75.35,825.25,3.52,149.01
2009-01-01 10:00:00,175.62,1508.63,26.59,197.66,41.19,1093.92,116.61,1093.94,8.4,197.53,19.87,149.11,31.73,825.24,67.03,825.25,6.49,149.01
2009-01-01 11:00:00,192.76,1508.63,27.58,197.66,48.97,1093.92,184.01,1093.94,10.88,197.53,20.55,149.11,37.22,825.24,91.94,825.25,8.38,149.01
[...]

PR #2828, opened 2020-09-11
Introduce new Daysim binaries

reyery:
This PR solves #2794 by introducing a new config paramter `use-latest-daysim-binaries` (default `True`) for the radiation script to run using original Daysim binaries or the one complied from their master branch. More details of the issue, such as the non-deterministic results from the newly complied binaries, can be found in the linked issue.

The folder structure of the Daysim binaries shipped with CEA using the installer now consists of folders `bin`, `bin64` and `lib`.
`bin` will contain the original binaries, `bin64` the newly complied binaries, `lib` to contain files used by the binaries and the default path for the `RAYPATH` environment variable that is need for Daysim to run.

Older folder structure:
image

New folder structure:
image

Now the radiation script will check folders in this order for the Daysim binaries:
- path hint provided in the config parameter `daysim-bin-directory` (default empty, will always use user defined location if provided and valid )
- CEA Daysim dependencies binaries folder e.g. `C:\Users\User\Documents\CityEnergyAnalyst\Dependencies\Daysim\bin64` for `use-latest-daysim-binaries` is `True` or `C:\Users\User\Documents\CityEnergyAnalyst\Dependencies\Daysim\bin` when `False`
- CEA Daysim dependencies folder e.g. `C:\Users\User\Documents\CityEnergyAnalyst\Dependencies\Daysim` for backwards compatibility where binaries are in older folder structure.
- Default Daysim installer location (for Windows) i.e. `C:\Daysim`


reyery:
To test this, we need to make sure that the radiation script is using the correct Daysim binaries based on the config.
It can be tested using the following command in CEA console:
image
To check, the console will print out which binaries is used:
image

There are a few different scenario that could happen and needs to be tested:
1. User has the old Daysim folder structure as mentioned above (does not have the new binaries)
    - If binaries are valid, the old binaries would be used
2. User selects a path using `daysim-bin-directory` parameter
    -  If folder structure is in either the old or new structure and binaries are valid, the selected binaries would be used.
3. User has new Daysim folder structure (can be found here [Daysim.zip](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/5207217/Daysim.zip))
    - Based on the `use-latest-daysim-binaries`, the script will check in the `bin64` or `bin` folder for the binaries.

In any of the cases, if the path has a white space in it, e.g. space due to user name `C:\Users\User Name\Documents\CityEnergyAnalyst\Dependencies\Daysim`, it would be considered as an invalid path, since there is a bug in Daysim where is will not work if it's path has a space in it.

shanshanhsieh:
All three tests are past! Thank you for making the improvement, @reyery !

Issue #2844, opened 2020-10-19
Issues with CEA last version (3.11.0)


BeaHerrero:
**Describe the bug**
I tried to use it with a group of students and they all got errors when editing the databases and when running the radiance model or importing weather files. The error was "something went wrong" and no more details were shown. As they only difference with mine was the version (I have 3.7.0) I was wondering if there is some problem with the last version released.
**Hardware:**
 - Windows 10
 - CEA version 3.11.0

Mar-Ges:
I tried to download the last version (3.11.0), but i couldn't install it because of a runtime Error:
"R6034
An application has made an attempt to lead the C runtime library incorrectly. 
Please contact the application's support team for more information."

daren-thomas:
@BeaHerrero thank you for reporting this! I'm looking into it...

daren-thomas:
@BeaHerrero I've been trying to replicate your results. It seems importing weather files works for me, so does the radiation script.

When trying to save changes in the Database error, I get the "Something went wrong" error. I'm looking into that now.

BeaHerrero:
We used version 3.7.0 yesterday for doing the same and no one got any
error. Thank you for looking into it.

daren-thomas:
Looks like the error could be here:

```
  File "c:\users\darthoma\documents\cityenergyanalyst\dependencies\python\lib\site-packages\cea\interfaces\dashboard\api\inputs.py", line 351, in get
    return read_all_databases(locator.get_databases_folder())
  File "c:\users\darthoma\documents\cityenergyanalyst\dependencies\python\lib\site-packages\cea\interfaces\dashboard\api\databases.py", line 62, in read_all_databases
    out[category][db_name]['SCHEDULES'][db_file['name']] = schedule_to_dict(db_file['path'])
  File "c:\users\darthoma\documents\cityenergyanalyst\dependencies\python\lib\site-packages\cea\interfaces\dashboard\api\databases.py", line 40, in schedule_to_dict
    schedule_df = schedule_to_dataframe(schedule_path)
  File "c:\users\darthoma\documents\cityenergyanalyst\dependencies\python\lib\site-packages\cea\utilities\schedule_reader.py", line 94, in schedule_to_dataframe
    out['METADATA'] = pd.DataFrame({'metadata': [next(reader)[1]]})
StopIteration
```

I think maybe that code needs to be checked for py3 compatibility.

daren-thomas:
[...]

PR #2848, opened 2020-10-26
2844 database editor bug

daren-thomas:
This PR fixes issue #2844 (Issues with CEA last version (3.11.0)).

To test it:

- create a new scenario
- open "Database Editor"
- change some values
- click "Save Changes"

You should get a message that your changes were saved:

image


reyery (approved):
Looks fine to me. Just how python3 handles newline escape characters of different OS I suppose. I wonder if this also affects such occurrences in other part of the code.

daren-thomas:
@reyery actually, it was more about opening the file in text mode (as opposed to binary mode) that was necessary. The newline stuff... I just added it that way because someone on stackoverflow meantioned it.

daren-thomas:
@reyery would you mind merging to master so we can have this in the next release please?

PR #2850, opened 2020-10-28
2834 docker container for remap

daren-thomas:
This is a first (experimental) docker container for running the CEA. Steps to test:

- install docker (google knows)
- navigate to the repository in a docker-enabled shell
- `cd docker`
- `docker image build -t cea-server .`
- `docker container run -t -p 5050:5050 cea-server`

You should now be able to start the GUI and the terminal will show you the interactions of the GUI with the `cea dashboard` command that automatically gets started when you run the docker container.


reyery:
@daren-thomas maybe we could upload the docker image to the docker registry after testing, so users could just pull it. That would be cool!

daren-thomas:
Try `docker pull darenthomas/cityenergyanalyst:latest` :)

docker/cea.config
```diff
@@ -0,0 +1,288 @@
+[general]
```
reyery, line 1: Wait. Why do we need to provide the config file?

environment.ubuntu.yml
```diff
@@ -1,14 +1,73 @@
 name: cea
```
reyery, line 1: What's the difference between this and the one used to build the docker image?

docker/Dockerfile
```diff
@@ -1,27 +1,34 @@
...
 
+# create the conda environment and install cea...
 RUN conda env create -q -f /tmp/environment.yml -n cea
-RUN /bin/bash -c "source activate cea && \
-    pip install git+https://github.com/architecture-building-systems/CityEnergyAnalyst.git && \
-    cea extract-reference-case --destination ~/ && \
-    cea-config demand --scenario ~/reference-case-open/baseline && \
-    cea-config radiation --daysim-bin-directory /root/Daysim/bin"
+ENV PATH /opt/conda/envs/cea/bin:$PATH
+RUN /bin/bash -c "source activate cea && pip install --no-deps pip install --no-deps git+https://github.com/architecture-building-systems/CityEnergyAnalyst@2834-docker-container-for-remap#egg=cityenergyanalyst"
```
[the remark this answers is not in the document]
daren-thomas: I wrote an issue for that: #2852 - basically, install from master...

docker/Dockerfile
```diff
@@ -1,27 +1,34 @@
...
-RUN mkdir -p ~/.config/matplotlib && echo "backend: Agg" > ~/.config/matplotlib/matplotlibrc
+# conda-pack the environment (with cea pre-installed)
+RUN conda install -c conda-forge conda-pack
+RUN conda-pack -n cea -o /tmp/env.tar && mkdir /venv && cd /venv && tar xf /tmp/env.tar && rm /tmp/env.tar
+RUN /venv/bin/conda-unpack
+
+# The runtime-stage image; we can use Debian as the
+# base image since the Conda env also includes Python
+# for us.
+FROM debian:buster AS cea-runtime
```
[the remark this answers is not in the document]
daren-thomas: yep :) the images are smaller that way. thanks for the conda-pack tip - that lead me to a nice tutorial on how to do this.

docker/Dockerfile
```diff
@@ -1,27 +1,34 @@
...
 
-# use the environment - no need to source because it's the only environment anyway...
-ENV PATH "/opt/conda/envs/cea/bin:/root/Daysim/bin:$PATH"
-CMD ipython
\ No newline at end of file
+# When image is run, run the code with the environment
+# activated:
+ENV PATH "/venv/cea/bin:/Daysim/bin:$PATH"
+SHELL ["/bin/bash", "-c"]
+ENTRYPOINT source /venv/bin/activate && cea dashboard
```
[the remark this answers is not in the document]
daren-thomas: yes. bothers me too. but would change a lot and we both don't have time to go re-branding everything (including docs!)

docker/cea.config
daren-thomas: to be able to pre-set some stuff. like the default project folder etc. the one normally shipped is geared towards windows.

environment.ubuntu.yml
daren-thomas: not much difference at all! actually, they should be more or less the same.

reyery:
Hmmm I'm not able to open any projects using the GUI.

daren-thomas:
@reyery yes. this is due to the GUI doing client-side verification of the project folder. BUT. The docker container doesn't _have_ a `C:\` drive. instead, it has it's own filesystem. And when starting it, you _could_ mount a volume to map to your local computer. But the GUI would still not work, because it doesn't understand that the folders are not there... To fully have this work, we'll need to come up with a different solution for selecting projects / scenarios that takes into account that the _server_ is responsible for verification? (but for now it's ok. connecting the GUI to the dashboard is not necessarily a goal of the docker image)

reyery:
@daren-thomas alright then. I guess I can merge it now, since the GUI could connect to the docker container.

reyery (approved):

PR #2853, opened 2020-10-29
Add option for user-supplied PV tilt angle and maximum roof coverage

martin-mosteiro:
This PR adds the option for users to define their own PV tilt angle for PV placed on a flat roof. Panel spacing is still calculated the same way as before. In addition, a maximum roof coverage can be selected, whereby if a user supplies a roof coverage of X%, only X% of the roof will be considered for the calculation, but spacing is still calculated (i.e., the actual PV area will only be X% if the panels are placed completely flat).

The current calculation optimizes tilt angle for maximum production per m2, but that does not imply maximizing the total PV production. Hence, it could be desirable for users to use their own assumptions on tilt angle and roof coverage.

@shanshanhsieh: This is not quite the fix we discussed a while back, whereby input databases would have to be expanded. This does, however, allow the user to implement their own tilt angles. Furthermore, the roof coverage part is only considered for the user-supplied tilt angles at the moment - but perhaps this is also something that might be desirable for the case with optimized tilt angles?


shanshanhsieh:
Hi @martin-mosteiro ,

can you summarize the new inputs that the users can make now, and how to test this PR?

thanks!

martin-mosteiro:
- New inputs: tilt angle and maximum roof coverage (included in the tool)
- How to test: I'll go for an exhaustive test to make sure we're good.
1. Select a buidling in a case study
2. Run PV using the traditional method and check the optimal tilt angle selected by the PV potential tool
3. Run the new method with that tilt angle and 100% maximum roof coverage
4. Compare the results of steps 2 and 3
5. Run the new method with that tilt angle and a different maximum roof coverage
6. Compare the results of steps 3 and 5

cea/default.config
```diff
@@ -379,6 +379,11 @@ type-scpanel.type = ChoiceParameter
 type-scpanel.choices = FP, ET
 type-scpanel.help = Type of solar collector: "FP" (Flat-plate) or "ET" (Evacuated-tubes).
 
+type-placement = OPT
+type-placement.type = ChoiceParameter
+type-placement.choices = OPT, USER
+type-placement.help = Panel placement method: either "OPT" for optimal angle and corresponding spacing, or "USER" for a user-defined angle and roof coverage.
```
shanshanhsieh, line 385: please specify that this only applies to panels on roof, but not panels on walls

cea/default.config
```diff
@@ -408,6 +413,14 @@ solar-window-solstice.type = IntegerParameter
 solar-window-solstice.help = Desired time of solar exposure on the solstice in hours.
 solar-window-solstice.category = Advanced
 
+panel-tilt-angle = 10
+panel-tilt-angle.type = RealParameter
+panel-tilt-angle.help = Solar panel tilt angle if using user-defined panel tilt angle and roof coverage.
+
+max-roof-coverage = 0.8
+max-roof-coverage.type = RealParameter
+max-roof-coverage.help = Maximum roof coverage if using user-defined panel tilt angle and roof coverage.
```
shanshanhsieh, line 422: please describe how is the calculation done when the radiation threshold is also set

cea/technologies/solar/solar_collector.py
```diff
@@ -83,11 +83,19 @@ def calc_SC(locator, config, latitude, longitude, weather_data, date_local, buil
...
-                                                                      panel_properties_SC)
-        print('calculating optimal tilt angle and separation done for building %s' % building_name)
+        if config.solar.type_placement == 'OPT':
+            # calculate optimal angle and tilt for panels
+            sensors_metadata_cat = solar_equations.optimal_angle_and_tilt(sensors_metadata_clean, latitude,
+                                                                          solar_properties, max_annual_radiation,
+                                                                          panel_properties_SC)
+            print('calculating optimal tilt angle and separation done for building %s' % building_name)
+        else:
+            # calculate optimal angle and tilt for panels
```
shanshanhsieh, line 93: this documentation is not correct

cea/default.config
shanshanhsieh: also, only possible on flat roofs?

cea/default.config
shanshanhsieh: - the wording has to echo *type-placement*
- add unit

shanshanhsieh (commented):
some comments on the information shown in the interface

shanshanhsieh:
```
1. Select a building in a case study
2. Run PV using the traditional method and check the optimal tilt angle selected by the PV potential tool
3. Run the new method with that tilt angle and 100% maximum roof coverage
    *How can one use the optimal tilt angle but user specified roof coverage?*
```

would it be easier to change the interface to:
custom-tilt-angle : TRUE/FALSE
panel-tilt-angle: float
custom-roof-coverage: TRUE/FALSE
max-roof-coverage: float

this way, this two custom settings can be decoupled as well

cea/utilities/solar_equations.py
```diff
@@ -306,18 +307,94 @@ def optimal_angle_and_tilt(sensors_metadata_clean, latitude, solar_properties, m
...
+    else:
+        module_width_m = panel_properties['module_area_m2'] / module_length_m  # for FP, ET
+    module_flat_surface_area_m2 = module_width_m * (sensors_metadata_clean.array_spacing_m / 2 +
+                                                    module_length_m * cos(input_angle_rad))
+    area_per_module_m2 = module_width_m * module_length_m
+
+    # calculate the pv/solar collector module area within the area of each sensor point
+    sensors_metadata_clean['area_installed_module_m2'] = np.where(
+        sensors_metadata_clean['tilt_deg'] >= 5, roof_coverage * sensors_metadata_clean.AREA_m2,
+        area_per_module_m2 * (roof_coverage * sensors_metadata_clean.AREA_m2 / module_flat_surface_area_m2))
```
shanshanhsieh, line 339: i think this is changing ALL of the coverage, including the walls

shanshanhsieh (changes requested):

cea/default.config
martin-mosteiro: > also, only possible on flat roofs?
well, currently our code doesn't process anything _but_ flat roofs, right? so I can clarify that, but it's not like there is much of a difference :)

martin-mosteiro: rephrased as requested

cea/default.config
martin-mosteiro: rephrased as requested

cea/technologies/solar/solar_collector.py
martin-mosteiro: great catch, thank you!

cea/utilities/solar_equations.py
martin-mosteiro: again, great catch! fixed it

martin-mosteiro:
>  custom-tilt-angle : TRUE/FALSE
> panel-tilt-angle: float
> custom-roof-coverage: TRUE/FALSE
> max-roof-coverage: float

Great idea! I have changed the input parameters accordingly.

cea/utilities/solar_equations.py
```diff
@@ -306,18 +307,95 @@ def optimal_angle_and_tilt(sensors_metadata_clean, latitude, solar_properties, m
         same as the roof. Sensors on flat roofs are all south facing.
     """
     # calculate panel tilt angle (B) for flat roofs (tilt < 5 degrees), slope roofs and walls.
-    optimal_angle_flat_deg = calc_optimal_angle(180, latitude,
+    input_angle_rad = radians(panel_tilt_angle)
+    sensors_metadata_clean['tilt_deg'] = np.vectorize(acos)(sensors_metadata_clean['Zdir'])  # surface tilt angle in rad
```
shanshanhsieh, line 311: is this line redundant?

shanshanhsieh:
@martin-mosteiro the test went well!
there are two more things:
1. my comment in the script (see above)
2. the roof covering is actually the portion of the roof after filtered by radiation threshold, pls also add that in the documentation

thanks!

cea/utilities/solar_equations.py
martin-mosteiro: No, it calculates the tilt angle in radians, which is converted to degrees in the line after.

martin-mosteiro:
@shanshanhsieh 

1. Agreed not necessary.
2. I updated the documentation in default.config and in `solar_equations.py`

shanshanhsieh (approved):

Issue #2849, opened 2020-10-28
cea-doc html fails (TypeError: a bytes-like object is required, not 'str')


daren-thomas:
Found another thing that doesn't work quite yet. We should test this and the RTD documentation generation. This stems from the py3 port.

PR #2857, opened 2020-11-03
cea-doc html now works again

daren-thomas:
This PR addresses #2849 (cea-doc html fails (TypeError: a bytes-like object is required, not 'str')).

To test, run `cea-doc html` in the CEA Console.


shanshanhsieh:
```
Fatal error in launcher: Unable to create process using '"c:\users\darthoma\miniconda2\envs\cea-3.13.0\python.exe"  "C:\Users\Admin\Documents\CityEnergyAnalyst\Dependencies\Python\Scripts\sphinx-apidoc.exe" -f -M -T -o modules ../cea ../cea/databases* ../cea/optimization/master/generation*
echo.': The system cannot find the file specified.

Sphinx has updated cea rst files.
Fatal error in launcher: Unable to create process using '"c:\users\darthoma\miniconda2\envs\cea-3.13.0\python.exe"  "C:\Users\Admin\Documents\CityEnergyAnalyst\Dependencies\Python\Scripts\sphinx-build.exe" -M html . _build ': The system cannot find the file specified.

```

@daren-thomas , it seems like the filepath is fixed now

daren-thomas:
@shanshanhsieh fixed as in broken, right?

daren-thomas:
It seems the `conda-pack` of the sphinx stuff doesn't really work properly. I don't think I can do much about that.

In the cea console, typing `pip install --force-reinstall sphinx` fixes that. Afterwards, `cea-doc html` works.

I'm going to update the installer to do just that. Stay tuned.

daren-thomas:
@shanshanhsieh I've tested this now on a separate machine (with a different user) and it seems to work. You'll need to perform the following steps:

- delete your `Dependencies/Python` (or, in the next step, install to a separate folder)
- install this version: https://www.dropbox.com/s/h8jgjr5ja5gata4/Setup_CityEnergyAnalyst_3.14.0.exe?dl=0
- start CEA Console
- `cd` to your GitHub CityEnergyAnalyst repository (on my machines I need to `cd ..\GitHub\CityEnergyAnalyst` but your setup may differ)
- make sure you've checked out this branch
- `pip install --no-deps -e .` - _waitaminute_ why `--no-deps`? that tells pip not to bother with the dependencies, as we already provide those. It'll install faster, is potentially safer for the conda-installed stuff and should be the default way to go into "developer" mode.
- `cea-doc html` should now run through.

Would you mind testing that please so we can merge?

PR #2867, opened 2020-11-13
Add missing floor assemblies and update construction standards

martin-mosteiro:
While running some simulations I kept noticing this kind of energy balance:
Screen Shot 2020-11-13 at 11 33 39
…which seemed kind of odd.

So I checked it out and it seems that during the migration from the old envelope system database to the new one, only the worst possible floor assembly (U-value 2.9 W/m2K) was included. So, say, that "Minergie standard" assembly? Not Minergie standard at all. In fact, none of the assemblies would be allowed to be built in Switzerland at present day.

I have added the missing floor assemblies and updated the construction standards accordingly. The U-values are the same as they used to be (that is, MuKEN-compliant), the embodied carbon was also taken from the old CEA database.


martin-mosteiro:
not sure who would be best-suited to review this, so I'm assigning Zurich-based @daren-thomas

PR #2875, opened 2020-11-19
2840 allow gui to connect to remote server

daren-thomas:
This PR addresses use cases of `cea dashboard` where the HOST and PORT of the connection needs to change.

New parameters for `cea dashboard`:

- port: default=5050
- host: default=localhost

It goes in tandem with a PR on the GUI repository with the same branch name. To run, check out both repositories to the required PR, `cd` to the GUI repository and `yarn dev`.

Play around with different ports to see the effect.


reyery:
@daren-thomas it seems that running jobs don't work anymore, something to do with socketIO. I think the GUI (or maybe the backend?) is unable to listen to `0.0.0.0`.

daren-thomas:
@reyery it seems to work up until the message "using 7 CPUs"... I'll investigate the cea-worker...

daren-thomas:
It looks like it works for single-processing... Next thing to test is if the socket.io bug mentioned above is also present in `master`...

daren-thomas:
yep. seems ok to me. @reyery would you mind testing again and maybe mention how you tested?

reyery:
I checkout branch 2840 for both cea and the GUI.
Start `cea dashboard --host 0.0.0.0` in the cea console first then start the GUI.

I get this error when I set `--host` as `0.0.0.0` but not when it's `localhost`

image

I looks like it does not work when starting cea in console first then the GUI, but works when just starting the GUI.

reyery:
I think it might a problem with the `socket.io` version in the GUI being not compatible with the cea backend. I'll take a look at this.

shanshanhsieh:
@reyery will check the compatibility of the `socket.io` in the GUI

reyery:
image

Running `yarn why socket.io-client` (CEA GUI _Javascript_)
```
Found "socket.io-client@2.3.0"
```

Running `pip list | grep socket` (CEA _Python_)
```
python-socketio               4.6.0
```

Seems like both versions of the client (Javascript) and server (Python) are compatible with each other.

shanshanhsieh:
@reyery any update on this one? : )

reyery:
I think it works now after I updated the dependencies for the GUI and fix some issues in the backend. We need to rebuilding the GUI and create a new release after merging this PR.

shanshanhsieh:
also tested with `cea dashboard --host 127.0.0.1`
thank you @reyery and @daren-thomas !

PR #2914, opened 2021-01-18
Simplify preview image for scenarios with large number of buildings

### LINKED ISSUE #2910: Timeout function when previewing scenarios with large numbers of buildings
**Describe the bug**
When opening a big scenario (more than 1000 buildings), it takes a long time to create the preview image.
It shows "fetching image" for more than 5 mins.

**To Reproduce**
1. open a big scenario with more than 1000 buildings
2. run any script (e.g., GWR mapper)
3. go back to re-open scenarios

**Expected behaviour**
A timeout function on the hard limit on the number of buildings.
If the number is too big, the GUI should only describe an abstract image of the district.

reyery:
This PR resolves #2910 by generating a simplified preview image for scenarios with large number of buildings to prevent it from blocking other requests for too long which makes the GUI unresponsive.

### DIFF
--- a/cea/interfaces/dashboard/api/project.py
+++ b/cea/interfaces/dashboard/api/project.py
@@ -264,6 +264,8 @@ class Scenario(Resource):
 class ScenarioImage(Resource):
     @api.doc(params={'project': 'Path of Project (Leave blank to use path in config)'})
     def get(self, scenario):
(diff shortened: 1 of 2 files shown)


nasty-bos:
Updated shapeline colour to blue for better visibility.
image

Issue #2947, opened 2021-03-26
Fix call to get main use of each building from OSM


martin-mosteiro:
**Describe the bug**
When running the CEA zone helper, all buildings get imported as *MULTI_RES*. I thought using the `category` field obtained from OSM could be used in order to get the main use in the building... then noticed there's already code to do so in CEA! However, even though by default buildings should be assigned the right typology based on OSM, they are not.

**To Reproduce**
1. Create a new scenario in the CEA dashboard
2. Select any zone you want to – as long as not _all_ buildings are residential
3. Check the typology – see that non-residential buildings are also labeled as *MULTI_RES*

**Solution**
Fixing this issue was a super easy fix – the wrong column name was used in the creation of the typology file, so the categories from OSM were never _actually_ used in creating the typology file.

However, there's relatively few osm categories that are hard-coded into the list categories processed by CEA:
```
    if occupancy_type == "Get it from open street maps":
        no_buildings = typology_df.shape[0]
        for index in range(no_buildings):
            if zone_df.loc[index, "category"] == "yes":
                typology_df.loc[index, '1ST_USE'] = "MULTI_RES"
                typology_df.loc[index, "REFERENCE"] = "CEA - assumption"
            elif zone_df.loc[index, "category"] == "residential" or zone_df.loc[index, "category"] == "apartments":
                typology_df.loc[index, '1ST_USE'] = "MULTI_RES"
                typology_df.loc[index, "REFERENCE"] = "OSM - as it is"
            elif zone_df.loc[index, "category"] == "commercial" or zone_df.loc[index, "category"] == "civic":
                typology_df.loc[index, '1ST_USE'] = "OFFICE"
                typology_df.loc[index, "REFERENCE"] = "OSM - as it is"
            elif zone_df.loc[index, "category"] == "school":
                typology_df.loc[index, '1ST_USE'] = "SCHOOL"
                typology_df.loc[index, "REFERENCE"] = "OSM - as it is"
            elif zone_df.loc[index, "category"] == "garage" or zone_df.loc[index, "category"] == "garages" or \
                    zone_df.loc[index, "category"] == "warehouse":
                typology_df.loc[index, '1ST_USE'] = "PARKING"
                typology_df.loc[index, "REFERENCE"] = "OSM - as it is"
            elif zone_df.loc[index, "category"] == "house" or zone_df.loc[index, "category"] == "terrace" or \
                    zone_df.loc[index, "category"] == "detached":
[...]

PR #2948, opened 2021-03-26
Fix code to get buildings' main uses from OSM

martin-mosteiro:
This PR does a few things:
1. It fixes an issue (#2947) whereby an incorrect column name that led to OSM building categories not being used to assign CEA typologies (i.e., all buildings were imported as 'MULTI_RES')
2. It moves the names of OSM building categories and corresponding CEA use types into cea.datamanagement.constants in order to (a) keep the zone_helper code cleaner, and (b) allow the list of building categories considered in the future to potentially be expanded without needing to change the zone_helper code
3. It expands the number of [OSM building types](https://wiki.openstreetmap.org/wiki/Key:building) considered to include a wider range of possible values
4. It includes [OSM amenity types](https://wiki.openstreetmap.org/wiki/Key:amenity) as another valid source for data on building use type. According to the OSM wiki, if a building was originally built as, say, an industrial building but was later turned into retail, the it would be labeled `building=industrial` and `amenity=retail`, so the amenity type is actually more important than the building type!

As a result of this, the zone_helper now processes the building use types correctly and considers a number of additional use types:
Screen Shot 2021-03-26 at 18 07 31

To test this PR:
1. Create a new scenario in the CEA dashboard and select any zone you want to – as long as not all buildings are residential
2. Check the typology of the buildings and compare it to the `category` label in CEA
In the master branch, you'll see see that non-residential buildings are also labeled as MULTI_RES, but in this branch they will be correctly imported.


shanshanhsieh:
thanks @martin-mosteiro for the catch, i was also wondering about this yesterday!
thanks @reyery for the enhancement, note taken : )

PR #2951, opened 2021-03-31
Make convective heat transfer calculation at building surface dynamic

martin-mosteiro:
This PR implements the dynamic calculation method from ISO 6946 Annex A. The current thermal resistance of external surfaces (R_SE) in CEA is constant at a reference value of 0.04 taken from ISO 6946. This value in the ISO standard assumes an outdoor wind speed of 4 m/s, though:
Screen Shot 2021-03-31 at 15 56 45
However, in Zurich (for example) the outdoor wind speed is lower than that 97.5% of the time (according to TMY data), so the dynamic implementation should provide a better representation of heat transfer at each time of the year.

I believe the dynamic method _should_ be the default in CEA as it presents a more case study-specific parameter for the calculation. However, applying this calculation method had a noticeable impact on the results [in a previous study](https://www.frontiersin.org/articles/10.3389/fbuil.2020.553946/full). Since the CEA demand model with fixed R_SE has been validated (while my version with a dynamic R_SE hasn't), I think it's best not to tinker with the results in the unit test. So I have made `use-dynamic-heat-transfer-calculation` an option in the config file with a default value of `False`. Users can then decide whether they want to use the dynamic method or not.


martin-mosteiro:
As discussed in the CEA scrum, I've removed the option to use a fixed value of R_SE since it doesn't really have any advantage compared to the dynamic method.

shanshanhsieh:
differences in annual heating/cooling demand are less than 1% for reference case in CH and SG

PR #2957, opened 2021-04-27
"Pax" means passenger

martin-mosteiro:
Sorry, this is a bit of a pet peeve, but "pax"?
tumblr_mf5s0vlkqO1rbn1qw

Pax means "passenger", not person. [According to Wiktionary](https://en.wiktionary.org/wiki/pax), it can be used to mean "guest" by extension. But still neither "person" nor "occupant". Its use in this sense is common in Singapore, but not elsewhere, and in any case this meaning (as well as the meanings of "passenger" and "guest") are informal and not meant for a technical unit!

In the norms CEA is based on, the unit _p_ is often used to mean people, so I suggest using that instead. This was already the case in CEA a while ago, but it was changed, presumably because some variables (`Ve_lps`, `Vww_lpd`) look like the _p_ stands for "per" (i.e., "liters per second" and "liters per day"). This can easily be solved by using alternative variable names (i.e., `Ve_lsp`, `Vww_ldp`) instead.

This PR changes the following variable names:
- ~Ve_lpspax~: *Ve_lsp*
- ~Occ_m2pax~: *Occ_m2p*
- ~Qs_Wpax~: *Qs_Wp*
- ~Vw_lpdpax~: *Vw_ldp*
- ~Vww_lpdpax~: *Vww_ldp*
- ~X_ghpax~: *X_ghp*
This is much more consistent with the literature than the nomenclature used so far.

But wait, there's more! Since this change requires you to do some annoying relabeling in existing case studies, I've also uploaded [a Jupyter notebook](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/6381923/reference-case-convert.ipynb.zip) that'll rename these variables for you.


shanshanhsieh:
thank you @martin-mosteiro for this enhancement, consistent and logical variable naming is indeed important.
I am thinking what would be the best way to inform the users?
Perhaps making a new function under "Utilities" like the "Database Migrator"?

martin-mosteiro:
I added the contents of the jupyter notebook to the database migrator and tested on the current `reference-case-open` in the CEA examples folder. Worked out fine!

martin-mosteiro:
One more enhancement: I made the database migrator rename the columns in the corresponding technology database (`USE_TYPE_PROPERTIES.xlsx`) as well. The downside is after migrating the Excel file doesn't have the same pretty formatting it originally had. The upside is if the user has created their own database, they will get to update the column names without losing any of their data.

Again, tested by running the database migrator on the current _reference-case-open_ in the CEA database and obtained the expected results. Success!

PR #3000, opened 2021-07-08
update dockerfile

shanshanhsieh:
This PR resolves #2984 and #2985.
To test, follow the `\docker\README.md` and run any workflow with a case study in a container.

`cea test` is not quite working yet, but this could be solved in a later PR.

PR #3007, opened 2021-08-02
Move Dockerfile to root of the repository

reyery:
This PR moves the Dockerfile from the `/docker` folder to the root of the repository. 

After removing the need to copy special  files into the docker build context and standardizing the conda `environment.yml` file (#3000), the Dockerfile can now exist in the root of the repository. The path of the Dockerfile is now the same as the path of the build context, so running the build command can be done without specifying the path of the Dockerfile i.e `-f .`

Docker build command is changed from
```
# Run in the `/docker` folder
docker build -t user/cea .. -f .
```
to
```
# Run in the repository root
docker build -t user/cea .
```


shanshanhsieh:
@reyery 
`docker run --rm dockeruser/cea cea test` is working, but `docker run -t -p 5050:5050 dockeruser/cea` doesnt

shanshanhsieh:
@reyery here are the results from testing:

1. I can do `docker run -t -p 5050:5050 dockeruser/cea` and open dashboard.
1.1. I cannot open any project, the path is always directed to `\project`
1.2. I cannot see anything displayed here http://localhost:5050/api/ as described in the previous documentation
do we have to somehow modify the project root in the config?
image

shanshanhsieh:
@reyery here are the results from testing:
1. `docker run --name cea_dashboard -t -p 5050:5050 cityenergyanalyst/cea`
i am able to see the API via  http://localhost:5050/api/ 
but through cea dashboard GUI, I cannot access any project
image
I am guessing that we also have to bind a volume to it to transport projects?

2. when running in interactive mode `docker run -it cityenergyanalyst/cea  /bin/bash `
followed by `source /venv/bin/activate`, I can use the container like a cea console.
so everything is good, just that it would be nice to nine a column to share projects.

reyery:
>     1. `docker run --name cea_dashboard -t -p 5050:5050 cityenergyanalyst/cea`
>        i am able to see the API via  http://localhost:5050/api/
>        but through cea dashboard GUI, I cannot access any project
>        image
>        I am guessing that we also have to bind a volume to it to transport projects?

Yeah I think we need to bind a folder that contains the projects to the `/projects` folder in docker. Hmm and for the error I am not sure if it is validating the path based on the paths in the docker container or the local machine. I have a feeling that it is the latter and that would mean that we have to modify the GUI code to not do that.

shanshanhsieh:
this also works now, i will update the documentation and merge the branch 
`docker run --name cea_container -v /home/cea_projects:/projects cityenergyanalyst/cea cea workflow --workflow /projects/workflow.yml`

shanshanhsieh:
> 
> 
> > ```
> > 1. `docker run --name cea_dashboard -t -p 5050:5050 cityenergyanalyst/cea`
> >    i am able to see the API via  http://localhost:5050/api/
> >    but through cea dashboard GUI, I cannot access any project
> >    image
> >    I am guessing that we also have to bind a volume to it to transport projects?
> > ```
> 
> Yeah I think we need to bind a folder that contains the projects to the `/projects` folder in docker. Hmm and for the error I am not sure if it is validating the path based on the paths in the docker container or the local machine. I have a feeling that it is the latter and that would mean that we have to modify the GUI code to not do that.

yes, maybe we should create a new issue for the GUI error

PR #3027, opened 2021-09-20
Send regex info to frontend

### LINKED ISSUE #2990: Error in Edit air-conditioning-systems Table
**Describe the bug**

When editing the air-conditioning systems  table of a building, for example, cool_ends, "15|08" is entered, it shows "cool_ends is not a undefined"

**Screenshots**
Snipaste_2021-06-07_08-14-53

**Hardware (please complete the following information):**
 - OS and version: [ Windows 10]
 - Version of CEA you are using [v. 3.22.0]

reyery:
This closes #2990. It fixes the issue where regex information is not being sent to the front end, that is why is was not able to validate such inputs correctly.

### DIFF
--- a/cea/interfaces/dashboard/api/inputs.py
+++ b/cea/interfaces/dashboard/api/inputs.py
@@ -247,6 +247,10 @@ def get_building_properties():
                     columns[column_name]['choices'] = get_choices(column['choice'], path)
                 if 'constraints' in column:
                     columns[column_name]['constraints'] = column['constraints']
+                if 'regex' in column:
+                    columns[column_name]['regex'] = column['regex']
+                    if 'example' in column:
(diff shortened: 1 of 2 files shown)


shanshanhsieh:
Hey @reyery ,

I think the fix is working pretty well.
There is one tiny glitch that I am not sure if it's related to the cea source code of the GUI.
When editing the air-conditioning tab from dashboard, it requires all four slots to be updated.
However, in the column, it seems like it will remain "unchanged" if left untouched. (see image)
It could be great if that is really the case.

image

reyery:
@shanshanhsieh I have updated the GUI in https://github.com/architecture-building-systems/CityEnergyAnalyst-GUI/pull/53#issuecomment-942865192

PR #3028, opened 2021-09-23
Mac OS installation guide

martin-mosteiro:
I have created a simple installation guide for Mac. It essentially leads to the tutorial on running CEA on Docker plus the installation guide in the CEA GUI repository, plus a few extra commands to run the CEA dashboard.

I tested installing on my Mac, running the CEA test workflow on the command line on Docker and running the radiation script on the dashboard. All of these _worked_ though the radiation script job was perpetually listed as "pending" on the dashboard. I mentioned this issue on the installation guide, as I assume this is not such an easy fix and it doesn't _really_ affect the CEA functionality.


reyery:
Maybe we can update the guide to include a way of running it natively using the new binaries. (If the produced results are correct)

Issue #2002, opened 2019-05-23
Negative values in PV potential


martin-mosteiro:
Some interesting results from Jakub: a PV array producing negative electricity!
<img width="651" alt="image" src="https://user-images.githubusercontent.com/18548065/58244787-a4abf080-7d53-11e9-93f9-2848a5beb880.png">
(That's a building in the Zurich case study and the date is 20 February).

A similar issue had previously been found (and solved) for solar collectors (#1409).

shanshanhsieh:
this behavior cannot be reproduced by `reference-case-open` and `reference-case-cooling`
will be closing this issue for now

PR #3029, opened 2021-09-23
Update photovoltaic.py

### LINKED ISSUE #3022: Very Important Bug in PV calculation: results does not seem to be aligned
**Describe the bug**
As it can be seen in thh figure there are two sever problems in the PV calculation.

1. The efficiency of the panel is 30% sometimes 200%. physically not possible. There cannot be more PV electricity thatn solar radiation.
2. The peak happens at the early morning 6:00am, it is rather not not posible that it is higher than at noon always.

**To Reproduce**
1. Go to PV and rund the script for a location (here it was in Zurich)
2. Click on the output file of the PV calculation
3. Check the diference between electricity generated and radiation.

**Expected behaviour**
PV potential should be around 15% at all times (According to the efficiencies of the database)
It cannot be never higher than 18% of the solar radiation. Never 30% or 40%, or 200%! as it happens now.

**Screenshots**
[PV_total_fehlerhaft.csv](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/7090090/PV_total_fehlerhaft.csv)

**Hardware (please complete the following information):**
 - OS and version: Windows 10
 - Version of CEA: Latest, downloaded yesterday

Asigining to @shanshan for now since the latests modifications belong to one commit.
Since it makes the script unusuable, it may be important to tackle it early.

shanshanhsieh:
This PR resolves #3022 

Two parts are updated:
- calculate the angle of incidence, teta_rad, with pvlib
- limit air mass modifier to a feasible range (0.001,1). according to http://www.sciencedirect.com/science/article/pii/S0038092X05002410. 
image

From my side, the efficiencies for PV1 are around 10% on rooftops in summer.
image
The efficiency peaks in the morning hours could be avoided by increase the lower bound of the air mass modifier.
Any recommendation is welcome!

*To test*
1. switch to branch i3022
2. in cea console, install pvlib by typing: `pip install pvlib`
3. choose your case study and run PV calculation
4. verify the results so we could discuss in the comments below

I am assigning @jimenofonseca  and @justinfmccarty as reviewers since you are both analysing results from this script.


justinfmccarty:
Ok. I have some results. I used the default PV1 and default settings except for the annual threshold, which i set to be 100 instead of 800 to access more surfaces. In short, my evaluation is that @shanshanhsieh has fixed it. 

Steps:
1. Run altstetten model on main branch, export results
2. Run altstetten model on i3022 branch
3. compare outputs

Plots below. General result across all buildings:

Efficiency max of original is 1.75
Efficiency max of i3022 is 0.27

Efficiency mean of original is 0.15
Efficiency mean of i3022 is 0.09

Efficiency min of original is 0.0
Efficiency min of i3022 is 0.0


comparing_i3022

justinfmccarty (commented):
Ok. I have some results. I used the default PV1 and default settings except for the annual threshold, which i set to be 100 instead of 800 to access more surfaces. In short, my evaluation is that @shanshanhsieh has fixed it. 

Steps:
1. Run altstetten model on main branch, export results
2. Run altstetten model on i3022 branch
3. compare outputs

Plots below. General result across all buildings:

Efficiency max of original is 1.75
Efficiency max of i3022 is 0.27

Efficiency mean of original is 0.15
Efficiency mean of i3022 is 0.09

Efficiency min of original is 0.0
Efficiency min of i3022 is 0.0


comparing_i3022

shanshanhsieh:
Thanks for the review, @justinfmccarty 
- It seems like there are some hours with negative efficiency, can you confirm that? it does not seem right.
- As for the smiley face, there is not much i can do with the current model.
Here is a similar observation from literature:
https://link.springer.com/article/10.1007/s10973-020-09883-6/figures/7

jimenofonseca:
I wonder what is the expected efficiency of the panel. If it goes in the range of 0.27, then it seems to work fine. Weird that we have used the same script for 7 years without changing the air modifiers, and until now we notice this issue. I wonder if there was some issues introduced during the last refactoring that we are missing.

@shanshanhsieh any comment?, if not, we can merge I guess. Maybe the theta calc was wrong after refactoring a year ago?

jimenofonseca:
Thanks for the analysis justin. It is troubling indeed. While im not a fan of it, i will then limit to the nominal efficiency any point where the potential exceeds it. The main issue is that having those outliers as potential Peaks of pv generation may lead to understimate the costs of the technology down the road. It will also indicate that the area needed is less to generate 1kWp etc. We may need to do this work around for now. Until we find the bug…

shanshanhsieh:
Hey @jimenofonseca @justinfmccarty ,
thanks for your inputs!

here are some thoughts from my side:
- I do not think the outliers are caused by a bug, but rather the limitation of the current model we are following. As for why this error only pops up now, I can't really tell... 
- In my opinion, I do not think the nominal efficiency is referring to the highest possible efficiency of the pv panel. By definition, the nominal efficiency is the efficiency at the standard testing condition. Following this definition, it is possible to have higher efficiencies than the nominal in some hours.
And it is supported by these experiments:
https://link.springer.com/article/10.1007/s10973-020-09883-6/figures/7
https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=7918763&tag=1
- Looking at the output from this PR, there is no peak power production during the hours with "abnormal"(high) efficiencies. So I think the peak power value is legit.
- What kind of cost calculation are you referring to that uses the peak efficiency of the simulated result? 
For LCOE, performance ratios (e.g, on a surface that receive 1000kWh/yr: PR=simulated production/1000***nominal** efficiency) or energy yield (kWh/**kWp**, kWp=kWp(radiation)*nominal efficiency) are used. In those cases, the simulated peak efficiency is not part of the calculation.

Please let me know what I am missing here, so I can put it to the known issues. thanks!

martin-mosteiro:
My two cents: the negative efficiency issue is not a new one and there's actually an issue about this from 2019 (#2002). At the time, most of the CEA contributors were busy, busy bees with PhD's to finish, so we decided to leave this as a known issue.

No idea about the large outliers, though, maybe they've always been there and nobody actually bothered to check whether the results made sense...

Issue #2973, opened 2021-05-11
PCM thermal storage for district Cooling


jimenofonseca:
**Describe the feature you would like to have**
We will be integrating a PCM (phase change material) thermal storage for the district cooling system of CEA. THis is part of the DUCT modeling effort of Cooling Singapore. I will be supporting this effort in the next three months :-)

@shanshanhsieh, I wonder if your have come across about ideas of an ice thermal storage that we can easily integrate into CEA. Here are the top options we have found so far for the singaporean context:

1.	2021, Stefano Mazzoni, Jia Yin Sze, Benedetto Nastasi, Sean Ooi, Umberto Desideri, Alessandro Romagnoli. A techno-economic as-sessment on the adoption of latent heat thermal energy storage sys-tems for district cooling optimal dispatch & operations, Applied Ener-gy.
2.	2020, Y.L. Shao, K.Y. Soh, Y.D. Wan, M. Kumja, Z. Khin, M.R. Islam, K.J. Chua, Simulation and experimental study of thermal storage sys-tems for district cooling system under commercial operating condi-tions, Energy,
3.	2019, Stefano Mazzoni, Sean Ooi, Benedetto Nastasi, Alessandro Romagnoli, Energy storage technologies as techno-economic parame-ters for master-planning and optimal dispatch in smart multi energy systems, Applied Energy.
4.	2017, Gabriele Comodi, Francesco Carducci, Jia Yin Sze, Nagarajan Balamurugan, Alessandro Romagnoli, Storing energy for cooling de-mand management in tropical climates: A techno-economic compari-son between different energy storage technologies, Energy,

What do you think of this?

**Describe the end purpose of this feature**
It will simulate an ice thermal storage system as part of the district cooling simulation of CEA. this feature will add a quasi-steady-state model to do this.

**Describe the end-user of this feature**
A researcher (Future position for district cooling expert for Cooling Singapore).

jimenofonseca:
@shanshanhsieh I wonder if you have any other model I should consider?

shanshanhsieh:
@jimenofonseca , this is a great addition to the technology library!
i dont have any reference specifically for ice storage, but i have quickly looked through the references you provided. 
the model formulation from Mazzoni (1&3) could be compatible to CEA, and the details are well documented
it is definitely a good place to start.
let me know how it goes : )

jimenofonseca:
[...]

PR #3056, opened 2021-10-29

MatNif:
Resolves Issue #2973 
To test go to x, y do this


jimenofonseca:
This was a test, we close it until we have it ready with Matthias

PR #3076, opened 2021-12-21
Clean up GitHub repository

reyery:
This PR adds documentation on how to clean the GitHub repository, mostly to reduce its overall size by deleting old files that are not being used anymore.

There are still things that we need to follow up before running the script on our end.
- Inform internal devs to merge their branches.
- @shanshanhsieh clone a local backup.
- @shanshanhsieh make a fork for the current state as backup.
- @reyery  to run the script and relevant commands to update history in GitHub
- After introducing new history to GitHub: Make sure that we rebase new PRs from external users.


shanshanhsieh:
1. all active developers are being informed
2. a copy is saved to the A/S server under the path `03-Research\01-Projects\20_CEA\2022_CEA_Repository_Archive`
3. a clone has been imported to: https://github.com/shanshanhsieh/CityEnergyAnalyst_v3.29.0_backup
importing method: https://docs.github.com/en/enterprise-cloud@latest/get-started/importing-your-projects-to-github/importing-source-code-to-github/importing-a-repository-with-github-importer

@reyery , I am not sure if the import method is correct. if yes, perhaps you can also make a clone?

PR #3081, opened 2022-01-27
redo commit eb298828dd2e22ea6437d96e4c84eaea422cc77e)

### LINKED ISSUE #3079: Defining a Temp folder structure that allows for parallelization of multiple scenarios
**Describe the bug**
I am trying to generate a script that runs multiple scenarios in parallel (instead of the typical use in CEA of a single scenario running buildings in parallel). 
While developing the code with @jimenofonseca he provided a hotfix for temp folders being scenario specific. @martin-mosteiro noted that this actually broke the main branch so he reverted it. I am now looking for solutions or ideas on how to tackle this problem, since I am not familiar with parallelization and the temp folders structure in CEA.
This helps to close #2998.

lguilhermers:
Fixes #3079 .
reimplemented changes from Jimeno's HotFix ("CEA needs a temp file per scenario. This should not change main functionality.")

I have tested with the district_cooling_system and district_heating_system workflows, as well as my own scenario and it seems to work without any additional changes. Please let me know if you spot any problem.

### DIFF
--- a/cea/inputlocator.py
+++ b/cea/inputlocator.py
@@ -37,7 +37,7 @@ class InputLocator(object):
         self._wrap_locator_methods(plugins)
         self.plugins = plugins
 
-        self._temp_directory = None
+        self._temp_directory = tempfile.TemporaryDirectory()
 
     def __getstate__(self):
         """Make sure we can pickle an InputLocator..."""
@@ -45,7 +45,8 @@ class InputLocator(object):
             "scenario": self.scenario,
             "db_path": self.db_path,
             "weather_path": self.weather_path,
-            "plugins": [str(p) for p in self.plugins]
+            "plugins": [str(p) for p in self.plugins],
+            "_temp_directory": self._temp_directory
         }
 
     def __setstate__(self, state):
@@ -56,6 +57,8 @@ class InputLocator(object):
         self.weather_path = state["weather_path"]
         self.plugins = [instantiate_plugin(plugin_fqname) for plugin_fqname in state["plugins"]]
         self._wrap_locator_methods(self.plugins)
+
+        self._temp_directory = state["_temp_directory"]
 
     def _wrap_locator_methods(self, plugins):
         """
@@ -1013,8 +1016,6 @@ class InputLocator(object):
     # OTHER
     def get_temporary_folder(self):
         """Temporary folder as returned by `tempfile`."""
-        if self._temp_directory is None:
-            self._temp_directory = tempfile.TemporaryDirectory()
         return self._temp_directory.name
 
     def get_temporary_file(self, filename):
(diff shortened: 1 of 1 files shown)


reyery:
I think instead of writing our own logic of creating a temporary directory, we should use python built-in module `tempfile` and remove the methods from the `InputLocator`. 

Directories created using `tempfile` would be cleared when the python interpreter ends. This would be a more consistent behavior.

shanshanhsieh:
thanks for your input, @reyery , i was about to ask for your opinion!
@lguilhermers , could you please implement Rey's suggestion?

lguilhermers:
Not sure I understood that, because in the inputlocator the folders are already using the module tempfile (line 1015 of input locator), therefore clearing when the python interpreter ends:
 `temp_folder = os.path.join(tempfile.gettempdir(), self.scenario.split("\\")[-1])`
while still giving the flexibility to call for this exact location in different parts of the cea code. Since it is used in multiple parts of the radiation code, it seems to me it is still beneficial to have the inputlocator related to it, did I miss something?

lguilhermers:
@reyery @shanshanhsieh are you able to provide me some more information so I can work on it? Or perhaps meet to discuss it, if you have the time. I wanted to finish this so I can test and publish the plugin as well.

Anther option I would suggest is to test if the current implement works and publish it (since it actually fixes my bug with the plugin), and open another issue to improve the current method related to the input locator.

reyery:
Hey @lguilhermers sorry for the delay and not providing any examples. 

Using `tempfile.gettempdir` only returns the temporary directory and does not do any cleanup after. I think my idea is that the temporary directory should be scoped by the particular **script** that is using it and not the **scenario**, as these files should technically be transient. Any concrete files that potentially other scripts might use should not live in the temporary directory. One way to solve it could be to delete the temp directory manually every time after the temp directory is used, but that could potentially cause problems as well (e.g. forgetting to delete it).

Therefore, when creating a temporary directory, it should just be done within the script by calling:
```python
import tempfile

temp_dir = tempfile.TemporaryDirectory()  # create a new temp dir instance
temp_dir_path  = temp_dir.name  # get the path of the temp dir

# can then use this path for creating file, etc.
with open(os.path.join(temp_dir_path, "test"), "w") as f:
    f.write("testing")

do_something(temp_dir_path)

# `temp_dir_path` will be deleted after this script ends
```

lguilhermers:
I have reviewed @reyery latest changes and they seem to be working well.
I tested with CEA testing scripts, my own scenario, as well as the plugin (stochastic scenario generator) and it seems to work with no problem.
The Temp file does cleanup after (I noticed only a small increase of 2-4 MB in the folder, but it doesn't seem to be from the CEA scripts)

shanshanhsieh:
@martin-mosteiro , would you have time to review this?

PR #3095, opened 2022-02-24
Bugfix in cooling resources activation

MatNif:
The calculation of the absorption chiller dimensions before and after accounting for the cold storage occasionally led to two different absorption chillers being selected in the two sections. Therefore the requested capacity (determined in the 2nd iteration) could lie outside of range of possible setpoints of the absorption chiller (determined in the 1st iteration) at line 162 (former line 170): NG_Trigen_req_W = Q_used_prim_CC_fn_W(Q_CHP_gen_W)

### DIFF
--- a/cea/optimization/slave/cooling_resource_activation.py
+++ b/cea/optimization/slave/cooling_resource_activation.py
@@ -117,11 +117,24 @@ def cooling_resource_activator(Q_thermal_req,
     if master_to_slave_variables.NG_Trigen_on == 1 and Q_cooling_unmet_W > 0.0 and not np.isclose(
             T_district_cooling_supply_K,
             T_district_cooling_return_K):
+        # Check what the maximum cooling capacity of the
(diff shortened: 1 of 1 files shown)


shanshanhsieh:

shanshanhsieh:
Hi @MatNif ,

Thanks for the PR.
Could you please describe the steps on how to test this PR?

MatNif:
There are essentially 2 possibilities for testing the current behaviour of the model:

1. Run the optimisation script "Supply System Part II (centralised)" in the PCMStorageTestI Scenario unter the following link: https://we.tl/t-DI9JnmIG0U. This scenario triggers an error using the current master branch. In 3092 this error is resolved. 
2.  A more elaborate but also more time intensive way to test would be to run the "cooling_resource_activator"-function in "cooling_resource_activation.py" repeatedly while incrementally going through the possible range of "Q_thermal_req"-values. In the current master branch this should return an error when "Q_thermal_req" is close to the max capacity of small version of the absorption chiller type selected.

PR #3109, opened 2022-03-17
3102 decentralized buildings cooling

### LINKED ISSUE #3102: Error in ranking of decentralised building cooling/heating systems
**Describe the bug**
The ranking algorithm of for decentralised building cooling systems based on cost and emissions does not seem to identify the best solution but rather just returns the first one in the array that is handed to the function.

**To Reproduce**
1. Debug 'cea>optimization>preporcessing>dencentralised_buildings_cooling.py' while setting break points at lines 528 and 547 while using the scenario and config file found under the following link: https://we.tl/t-K1ZAvvnL43 (make sure you change project path in config file)
3. Check the emissions and costs-vectors once the debugger reaches the first break point. (solution 2 should have both the lowest emissions and cost)
4. Check the 'Best' solution once the debugger reaches the second breakpoint (should be solution 1) 

**Expected behaviour**
I am not sure what the idea behind the algorithm was at the start.
We could either transform this into a selection of one random non-dominated solution OR
try to return all the non-dominated solutions (i.e. all pareto-optimums) <- requires a lot more change in the code

**Screenshots**
Here is a visual representation of what the algorithm currently does:

Screen Shot 2022-03-11 at 12 21 34 PM

shanshanhsieh:
This PR resolves #3102 .
The error came from: 
(1) the incorrect/hard-coded value for the number of objectives (`optsearch.fill(3)`).
(2) the initialization of `indexBest=0` is changed to `None`, it was the reason why the first configuration always got picked

To fix the error, the number of objectives is updated to 2 (cost and emission), but it is still hard-coded at the moment.
Some refactoring is also done to avoid hard-coded values for number_of_configurations.

To test:
- Run the same scenario provided in #3102, and check whether the best index is found.

### DIFF
--- a/cea/optimization/preprocessing/decentralized_buildings_cooling.py
+++ b/cea/optimization/preprocessing/decentralized_buildings_cooling.py
@@ -542,7 +542,7 @@ def rank_results(TAC_USD, TotalCO2, TotalPrim, number_of_configurations):
         optsearch[int(CO2S[rank][0])] -= 1
         if np.count_nonzero(optsearch) != number_of_configurations:
             Bestfound = True
-            indexBest = np.where(optsearch == 0)[0][0]
+            indexBest = np.where(optsearch == 0)[0][0] #FIXME: it is possible to have more than one configuration that reaches zero here
         rank += 1
     # get the best option according to the ranking.
     Best[indexBest][0] = 1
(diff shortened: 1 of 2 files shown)


MatNif (approved):
Changes fix the bug and select the configuration with the best rankings for emissions and cost.
If the top two configurations have the exact same rankings (e.g. config 4: emissions rank 1, cost rank 3; config 7: emissions rank 3, cost rank 1), the routine will select the first of the two configurations (config 4 in the example above). Issue #3110 proposes an extension to address this.

PR #3114, opened 2022-03-22
`zone_helper` cannot import LineString geometry types

martin-mosteiro:
While testing a case study, I found a couple of instances of `LineString` geometry type, which the `zone_helper` was unable to deal with.

After checking, in my specific case study, these were categorized in OSM as "garden" and "way", i.e., neither was an actual building. So the easiest solution is to ignore this geometry type, much like `Point` geometry types are already ignored by CEA.

### DIFF
--- a/cea/datamanagement/zone_helper.py
+++ b/cea/datamanagement/zone_helper.py
@@ -296,7 +296,7 @@ def clean_geometries(gdf):
         if geometry.type == 'Polygon':  # ignore Polygons
             return geometry
         elif geometry.type in ['Point', 'LineString']:
-            print('Discarding geometry of type: {}'.format(geometry.type))
+            print(f'Discarding geometry of type: {geometry.type}')
             return None # discard geometry if it is a Point or LineString
         else:
             joined = unary_union(list(geometry))
--- a/cea/datamanagement/zone_helper.py
+++ b/cea/datamanagement/zone_helper.py
@@ -84,9 +84,9 @@ def clean_attributes(shapefile, buildings_height, buildings_floors, buildings_he
         data_floors_sum = [x + y for x, y in zip([parse_building_floors(x) for x in data_osm_floors1],
                                                  [parse_building_floors(x) for x in data_osm_floors2])]
         data_floors_sum_with_nan = [np.nan if x < 1.0 else x for x in data_floors_sum]
-        data_osm_floors_joined = int(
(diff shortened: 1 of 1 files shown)


martin-mosteiro:
While fixing this issue, I also found a strange issue. The following line
```[int(x) if x is not np.nan else data_osm_floors_joined for x in data_floors_sum_with_nan]```
led to a `ValueError`: ```ValueError: cannot convert float NaN to integer```
...which is weird, because that line shouldn't be trying to convert x if x is NaN. Weirder still, the following line (which should be equivalent, as far as I can tell) seems to fix the problem: ```[int(x) if not np.isnan(x) else data_osm_floors_joined for x in data_floors_sum_with_nan]```

File: cea/datamanagement/zone_helper.py [hunk 1]
```diff
@@ -295,9 +295,9 @@ def flatten_geometries(geometry):
         from shapely.ops import unary_union
         if geometry.type == 'Polygon':  # ignore Polygons
             return geometry
-        elif geometry.type == 'Point':
-            print('Discarding geometry of type: Point')
-            return None # discard geometry if it is a Point
+        elif geometry.type in ['Point', 'LineString']:
+            print('Discarding geometry of type: {}'.format(geometry.type))
```
reyery, line 299: I think we can start using [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) (introduced in Python 3) for formatting strings. It looks neater and more concise. 

e.g.
```suggestion
            print(f'Discarding geometry of type: {geometry.type}')
```

martin-mosteiro: Updated accordingly!

reyery:
> While fixing this issue, I also found a strange issue. The following line `[int(x) if x is not np.nan else data_osm_floors_joined for x in data_floors_sum_with_nan]` led to a `ValueError`: `ValueError: cannot convert float NaN to integer` ...which is weird, because that line shouldn't be trying to convert x if x is NaN. Weirder still, the following line (which should be equivalent, as far as I can tell) seems to fix the problem: `[int(x) if not np.isnan(x) else data_osm_floors_joined for x in data_floors_sum_with_nan]`

I think `x not np.nan` does not actually test for `nan` since `float('nan') is np.nan` is false and only `np.nan is np.nan` is true, so it might only be checking if it is actually of type `np.nan` (which might not always be the case). `np.isnan()` is a more accurate way of testing for `nan`. We should probably change that.

reyery (approved):

PR #3116, opened 2022-03-23

MatNif:


shanshanhsieh:

MatNif:
The best way to test the branch is probably to run the optimisation script (centralised) in the dashboard with this branch activated. I would recommend doing so and freely choosing one of the thermal storage types (TES1-TES5) and activating the debug function as highlighted below:

Screen Shot 2022-03-30 at 3 23 52 PM
 
To reduce the runtime of the test the number of generations under "Advanced" could be reduced to e.g. 10.

shanshanhsieh:
hi @MatNif , i am getting to review this branch.
could you please update from master and resolve conflicts first? thanks!

MatNif:
I just updated the branch and resolved conflicts. @shanshanhsieh you should be able to start with the review now.

shanshanhsieh:
@MatNif , thanks for the quick solve.
there is a new conflict from the latest PR, please update `cea/databases/CH/components/CONVERSION.xlsx` and commit again.
thanks!

MatNif:
Hi @shanshanhsieh.
I couldn't find any difference in the values of the database so I accepted the version from the master.

shanshanhsieh:
hey @MatNif , the check is failing because of the reference case `cea\examples\reference-case-open.zip`
from the history, it seems like you have done some changes there.
if that is the case, please download the `cea\examples\reference-case-open.zip` from master, implement you changes and then commit again.
the main problem I see here is that the excel files in the technology folder is still using `.xls` but not `.xlsx` 
thanks!

shanshanhsieh:
```
  File "C:\Program Files (x86)\Jenkins\workspace\run cea test for pull requests\cea\tests\test_technologies.py", line 46, in setUpClass
    cls.type_storage_list = cls.locator.get_database_conversion_systems_cold_thermal_storage_names()
  File "C:\Program Files (x86)\Jenkins\workspace\run cea test for pull requests\cea\inputlocator.py", line 425, in get_database_conversion_systems_cold_thermal_storage_names
    data = data[data["type"] == "COOLING"]
  File "C:\ProgramData\ceajenkins\ceatest-py3\Python\lib\site-packages\pandas\core\frame.py", line 2800, in __getitem__
    indexer = self.columns.get_loc(key)
  File "C:\ProgramData\ceajenkins\ceatest-py3\Python\lib\site-packages\pandas\core\indexes\base.py", line 2648, in get_loc
    return self._engine.get_loc(self._maybe_cast_indexer(key))
  File "pandas\_libs\index.pyx", line 111, in pandas._libs.index.IndexEngine.get_loc
  File "pandas\_libs\index.pyx", line 138, in pandas._libs.index.IndexEngine.get_loc
  File "pandas\_libs\hashtable_class_helper.pxi", line 1619, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas\_libs\hashtable_class_helper.pxi", line 1627, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'type'
```

hi @MatNif , when running `cea test` there is an error in `get_database_conversion_systems_cold_thermal_storage_names`, i think it is related to `CONVERSION.xlsx`, please resolve the error there, and try running `cea test` locally to spot more errors if the checks didn't pass. thanks!

cea/optimization/master/master_to_slave.py
```diff
@@ -255,6 +256,7 @@ def calc_master_to_slave_variables(locator, gen,
 
     # initialise class storing dynamic variables transferred from master to slave optimization
     master_to_slave_vars = slave_data.SlaveData()
+    master_to_slave_vars.debug = config.general.debug
```
shanshanhsieh, line 259: what is the use of this?

cea/technologies/storage_tank.py
```diff
@@ -215,6 +217,15 @@ def calc_tank_temperature(T_start_C, q_loss_W, q_discharged_W, q_charged_W, V_ta
 # cold water storage tank design
 # ================================
 
+def calc_storage_tank_mass(Qc_tank_capacity_Wh,
```
shanshanhsieh, line 220: should it be names `calc_storage_tank_volume` instead?

shanshanhsieh:
@MatNif , another thing i dont understand is the `CTES` tab in `cea/databases/SG/components/CONVERSION.xlsx'
please delete the tab if that is not relevant, thanks!

cea/optimization/master/master_to_slave.py
MatNif: This is handed from 'evaluation.py' down to 'storage_tank_pcm.py' (through 'cooling_main.py', where the PCM storage is initialised). In 'storage_tank_pcm.py' you can see that the boolean is used to output basic properties of the storage upon initialisation and then charge-levels at each charging and discharging stage within 'cooling_resoucre_activation.py'.

The alternative would be to hand down config as a separate variable or to remove the debugging functionalities entirely. But I believe the current solution makes sense.

cea/technologies/storage_tank.py
MatNif: This function calculates the weight of the phase change material in the tank. Since the material's phase changes as the storage is charged and discharged, so does its density and therefore its volume. So I would say the current name is more accurate.

MatNif:
> @MatNif , another thing i dont understand is the `CTES` tab in `cea/databases/SG/components/CONVERSION.xlsx' please delete the tab if that is not relevant, thanks!

Yes, I believe the file was used for testing only. Or at least that is what I assume, since the only apparent difference from the 'TES' tab were the maximum and minimum temperatures attributed to the chillers and I could not find any reference to the tab in any of the scripts. So I ended up removing the tab as you suggested.

cea/optimization/master/master_to_slave.py
shanshanhsieh: thanks for the clarification!

cea/technologies/storage_tank.py
shanshanhsieh: i think we are referring to two different functions with the same name.
you are describing `calc_storage_tank_mass` in `storage_tank_pcm.py`, and i am referring to `storage_tank.py`
anyways, i can resolve that part

Issue #3121, opened 2022-03-30
OverflowError: math range error in thermal loads calculation


lguilhermers:
**Describe the bug**
The error was previously mentioned in #3067, while devoling a stochastic scenario generator for CEA (#2998).

Some particlar numbers of `theta ` would be problematic while trying to solve
`p_sat_int = 611.2 * math.exp(17.62 * theta / (243.12 + theta))`.
This is probably due to a very large number of decimals and doesn't fit into a double (as found here: https://stackoverflow.com/questions/4050907/python-overflowerror-math-range-error).

**To Reproduce**
1. Download [here](https://www.dropbox.com/s/j8e7kjxf6motzxl/err_overflow_mathrange.rar?dl=0) the scenario where error appears.
2. Run demand simulation
3. See error

```
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "C:\Users\luis.santos\Documents\CityEnergyAnalyst\Dependencies\Python\lib\multiprocessing\pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "C:\Users\luis.santos\Documents\CityEnergyAnalyst\Dependencies\Python\lib\multiprocessing\pool.py", line 44, in mapstar
    return list(map(*args))
  File "c:\users\luis.santos\documents\cityenergyanalyst\cityenergyanalyst\cea\utilities\parallel.py", line 131, in __apply_func_with_worker_stream
    result = func(*args)
  File "C:\Users\luis.santos\Documents\GitHub\stochastic_scenario_generator\data_generation.py", line 170, in stochastic_scenario_generator
    demand_main.demand_calculation(locator, config)  # runs demand simulation
  File "c:\users\luis.santos\documents\cityenergyanalyst\cityenergyanalyst\cea\demand\demand_main.py", line 130, in demand_calculation
    repeat(debug, n))
  File "c:\users\luis.santos\documents\cityenergyanalyst\cityenergyanalyst\cea\utilities\parallel.py", line 149, in wrapper
    result = func(*instance_args)
  File "c:\users\luis.santos\documents\cityenergyanalyst\cityenergyanalyst\cea\demand\thermal_loads.py", line 122, in calc_thermal_loads
    use_dynamic_infiltration_calculation, config)  # end-use demand latent and sensible + ventilation
[...]

PR #3122, opened 2022-03-30
changed exp solver to numpy

lguilhermers:
This solves #3121 .

To test: download the scenario mentioned in the issue above and test in master and in this branch.

Note: Even with the error gone, maybe would still be appropriate to look for the values being produced and see if they are making sense, but this is a bit beyond my capabilities and resources at the moment.

### DIFF
--- a/cea/demand/latent_loads.py
+++ b/cea/demand/latent_loads.py
@@ -185,7 +185,7 @@ def calc_saturation_pressure(theta):
     :rtype: double
     """
 
-    p_sat_int = 611.2 * math.exp(17.62 * theta / (243.12 + theta))
+    p_sat_int = 611.2 * np.exp(17.62 * theta / (243.12 + theta))
 
     return p_sat_int
 


shanshanhsieh:
@lguilhermers , the test has passed, thanks!
I also noticed that the temperature is quite odd, see warnings below:
```
Exception: Temperature in RC-Model of building B1001 out of bounds! First occured at timestep = 8049. The results were Tint = -103.89, theta_c = -119.28, theta_m = 25.71.
If it is an expected behavior, consider turning off over-heating warning in the advanced parameters to continue the simulation.
If it is not expected, check building geometry and internal loads.
Building might be too small in size or architecture parameter Hs_ag = 0.17 might be toosmall for this geometry. Current bounds of range for RC-model temperatures arebetween -30.0 and 50.0.
```

lguilhermers:
Indeed, those are not normal (unless cooling Singapore was too efficient in its tasks).
Maybe we need a hackathon and go through the thermal loads script, see if anything major is being missed. It seems there are problems that we can't understand that both overestimate and underestimate internal temperatures, which is a major point of the demand calculations.

For low Hs_ag (<0.25) and we can see something is wrong sometimes, displayed by the warning. My concern is that this error still has significant impacts on medium Hs_ag (0.25-0.6), without being so evident (hiding for less unrealistic results). 
One way I imagine we can analyze this is to do a sensitivity of Hs in relation to Tint and see when we observe unrealistic values.

What do you suggest?

Issue #3124, opened 2022-03-30
Demand forecast tool only considers ARU for cooling when no ventilation system is selected


martin-mosteiro:
**Describe the bug**
It appears that the demand forecast tool is only running ARUs, **NOT AHUs and SCUs**!
For example, look at the operating temperatures for a building with cooling system `HVAC_COOLING_AS4` (3-for-2, i.e., AHU+ARU+SCU).
![image](https://user-images.githubusercontent.com/18548065/160794130-bd043193-9111-4595-b701-bfdeeb9b3a3c.png)

Okay, that's not great. But what about the space cooling demand?
![image](https://user-images.githubusercontent.com/18548065/160795164-3ec47676-bc0a-418d-a155-04a84c3296fc.png)
i.e., the entire load is being supplied by the ARU.

I've checked what would happen if I had an office building with `HVAC_COOLING_AS3` (central AC, i.e., AHU+ARU) with similar results: the entire cooling load was provided by the ARU.

@shanshanhsieh Can you think of a non-buggy reason for this to happen?

**To Reproduce**
1. Create a case study along with its input databases
2. Assign a cooling system that includes AHUs and/or SCUs (e.g. `HVAC_COOLING_AS3`, `HVAC_COOLING_AS4`)
3. Run the solar radiation, schedule maker, and demand forecast tools
4. Check the results for `Qcs_sys_kWh`, `Qcs_sys_scu_kWh`, `Qcs_sys_ahu_kWh`, `Qcs_sys_aru_kWh`

**Expected behaviour**
The load shouldn't be supplied entirely by the ARU.

martin-mosteiro:
I know what you're thinking: "oh, I bet this is an issue that only shows up in Singaporean case studies!"
Well, check out this case study in Zurich (a large office building in the ETH campus with cooling system `HVAC_COOLING_AS3`):
![image](https://user-images.githubusercontent.com/18548065/160804920-88d61319-a2a4-484a-917e-6b351cbd2494.png)
So... yeah, there's an issue.

shanshanhsieh:
Hi @martin-mosteiro ,
The reason why AHUs are not triggered is probably because `type_vent=HVAC_VENTILATION_AS0`
AHU is part of the mechanical system, therefore, will only be used if `m_ve_mech>0`
[...]

PR #3130, opened 2022-04-01
3124 buildings with central AC getting cooled only by ARU

martin-mosteiro:
This PR solves issue #3124 by adding a function (`verify_hvac_system_combination`) to check whether an infeasible combination of systems (i.e., `HVAC_COOLING_AS3` or `HVAC_COOLING_AS4` with no mechanical ventilation) was selected.

If a building is found to have such an infeasible combination, a warning is printed to warn users of this issue, and a default ventilation system (`HVAC_VENTILATION_AS1`) is selected for the building in question.

This ensures that central and hybrid AC systems are operated correctly regardless of whether users select a ventilation system for the building or not.

To test:
1-Create a case study
2-Assign a building cooling system `HVAC_COOLING_AS3` or `HVAC_COOLING_AS4`, and ventilation system `HVAC_VENTILATION_AS0`
3-Run demand simulation
4-Verify whether the desired warning is printed and whether the systems are correctly activated in the simulation results.


shanshanhsieh:
Hi @martin-mosteiro ,
thanks for the bug fix, i have made a modification to make the verification function more generic.
this way, it could be applied to any other databases. 
re-assigning you as the reviewer now. 
please perform the same test and merge if it is working as intended, thanks!

shanshanhsieh (approved):
as mentioned in the previous comment, re-assigning @martin-mosteiro as the reviewer now

cea/demand/building_properties.py
```diff
@@ -1019,3 +1017,21 @@ def get_thermal_resistance_surface(prop_envelope, weather_data):
 
     return thermal_resistance_surface_wall, thermal_resistance_surface_roof, thermal_resistance_surface_win
 
+
+def verify_hvac_system_combination(prop_hvac, prop_emission_cooling, prop_ventilation_system_and_control):
```
martin-mosteiro, line 1021: Great idea!

I wonder if this could be made even more generaliz: perhaps there could be a boolean in the cooling tab of the HVAC database similar to the ones for the ventilation systems, e.g. `MECH_VENT`: for AS0-AS2 and AS5, it would be `FALSE` and for AS3-AS4 it would be `TRUE`. Then
```
needs_mech_vent = prop_emission_cooling[prop_emission_cooling['code'] == type_cs]['MECH_VENT'].values[0]
have_mech_vent = prop_ventilation_system_and_control[prop_ventilation_system_and_control['code'] == type_vent]['MECH_VENT'].values[0]
if (needs_mech_vent) & (not have_mech_vent):
[...]
```

Or am I just over-optimizing a solution that is already ok?

martin-mosteiro (commented):
Looks great! But I think it could look even better. If you think it's worth it @shanshanhsieh , I can spend an hour doing what I suggested in the comment. But at the same time, it won't change any functionality so... we can also just avoid that.

shanshanhsieh:
hi @martin-mosteiro ,
in my opinion, changing the database is a lot more tedious than an hour of work, especially since you also need to implement checks for backward compatibility ...
also, I am not sure if adding "MECH_VENT" into "prop_emission_cooling" is logical, because one could argue that why are the other four ventilation modes not specified inside  "prop_emission_cooling"?

nevertheless, I do think your code suggestion could help optimize the existing code, see comments above, it would be great if you could implement that!

cea/demand/building_properties.py
shanshanhsieh: we should use this logic to optimize the code and avoid the for loop!
inside the code, we could still write the array 'needs_mech_vent' based on 'type_cs', and compare 'needs_mech_vent' and 'have_mech_vent' for all buildings at the same time
would it work?

martin-mosteiro:
Hey @shanshanhsieh,

I think I haven't explained myself well enough. I didn't mean moving the `MECH_VENT` parameter from the ventilation database to the cooling database – I agree, that wouldn't make any sense. I mean creating a new column in the cooling database (could be named something like `NEEDS_MECH_VENT`, for example) to label systems that need a mechanical ventilation system to be selected. The advantage would be to avoid hard-coding system names in the CEA codes and thus allowing users to create new systems that use the same logic as the CEA database.

So the steps to take would be:
1. Update the CH and SG databases
2. Create a database migrator that adds this column to user databases
3. Run tests

Fair enough, maybe it won't take an hour of work. But I don't mind spending a bit more time and getting this PR right IFF you agree it makes sense to do what I described above. If you think it's not worth it, I'll just

shanshanhsieh:
@martin-mosteiro , thanks for the great addition to completing the warning msg!!!

PR #3154, opened 2022-04-27
Remove hardcoded floor height from calculations

martin-mosteiro:
This PR removes all instances in which the constant `H_F` is used in calculations. This constant represents the "average floor height" and is meant to be used as the floor height when none is introduced by the user. However, it shouldn't be used in any calculations, as the actual floor height should be whatever the CEA model inputs define.

Note: the tests fail because, precisely, the use of `H_F` in some calculations meant that for the test building B1010 (which has a height per floor of 4 m) the calculation used to be wrong.


martin-mosteiro:
While I was at it, I fixed one pet peeve of mine (changed 0.000277778 to 1 / 3600).

lguilhermers:
@martin-mosteiro the geometry_generator already has a floor_to_floor_height variable. Wouldn't it be better to take from there, instead of calculating again from databases_verification?

lguilhermers:
Besides the point above, I checked the files and tested going through a simple scenario. It all seems to work fine. 
Feel free to suggest other ways to test if you feel it is needed.

lguilhermers (approved):

reyery:
Seems like the test is only failing because of precision error. 

Does anyone remember how to update the test data? I remember there's a script that could do that.

```
FAIL: test_calc_thermal_loads (test_calc_thermal_loads.TestCalcThermalLoads)
  File "C:\Program Files (x86)\Jenkins\workspace\run cea test for pull requests\cea\tests\test_calc_thermal_loads.py", line 80, in test_calc_thermal_loads
    self.assertAlmostEqual(values[i], df[column].sum(), msg='Sum of column %s differs, %f != %f' % (
AssertionError: 157941.228 != 157941.20799999998 within 3 places (0.02000000001862645 difference) : Sum of column Qhs_sys_kWh differs, 157941.228000 != 157941.208000

FAIL: test_calc_thermal_loads_other_buildings (test_calc_thermal_loads.TestCalcThermalLoads)
Test some other buildings just to make sure we have the proper data
  File "C:\Program Files (x86)\Jenkins\workspace\run cea test for pull requests\cea\tests\test_calc_thermal_loads.py", line 101, in test_calc_thermal_loads_other_buildings
    self.assertAlmostEqual(expected_qhs_sys_kwh, qhs_sys_kwh,
AssertionError: 155418.337 != 155418.314 within 3 places (0.022999999986495823 difference) : qhs_sys_kwh for B1013 should be: 155418.31400, was 155418.33700

Ran 41 tests in 54.936s
```

martin-mosteiro:
Thanks @lguilhermers! I didn’t know that value was stored in the BuildingProperties object. Indeed, no need to calculate it again. I’ll change the PR accordingly

...when I’m back from my holidays ;)

@reyery I don’t think it’s a rounding error because the building that raises the error is the only building in the example case study that DOESN’T have a floor to floor height of 3 meters (i.e., the only one that doesn’t match the hard-coded value I removed). The difference in the results is indeed minimal, though. I’ll update the values in the test data accordingly

...when I’m back from my holidays ;)




Am 03.05.2022 um 14:49 schrieb Reynold Mok ***@***.******@***.***>>:

        - External Email -



Seems like the test is only failing because of precision error.

Does anyone remember how to update the test data? I remember there's a script that could do that.

FAIL: test_calc_thermal_loads (test_calc_thermal_loads.TestCalcThermalLoads)
  File "C:\Program Files (x86)\Jenkins\workspace\run cea test for pull requests\cea\tests\test_calc_thermal_loads.py", line 80, in test_calc_thermal_loads
    self.assertAlmostEqual(values[i], df[column].sum(), msg='Sum of column %s differs, %f != %f' % (
AssertionError: 157941.228 != 157941.20799999998 within 3 places (0.02000000001862645 difference) : Sum of column Qhs_sys_kWh differs, 157941.228000 != 157941.208000

FAIL: test_calc_thermal_loads_other_buildings (test_calc_thermal_loads.TestCalcThermalLoads)
Test some other buildings just to make sure we have the proper data
  File "C:\Program Files (x86)\Jenkins\workspace\run cea test for pull requests\cea\tests\test_calc_thermal_loads.py", line 101, in test_calc_thermal_loads_other_buildings
    self.assertAlmostEqual(expected_qhs_sys_kwh, qhs_sys_kwh,
AssertionError: 155418.337 != 155418.314 within 3 places (0.022999999986495823 difference) : qhs_sys_kwh for B1013 should be: 155418.31400, was 155418.33700

Ran 41 tests in 54.936s

martin-mosteiro:
@lguilhermers Thanks again for your review. The `floor_to_floor_height` you mentioned is part of the radiation workflow, so the floor to floor height needs to be calculated in the `building_properties` workflow in any case. One way to do so would be to call up the function that calculates `floor_to_floor_height` in the geometry generator, but that function does a lot of other things as well - meaning essentially we would be making the process to generate the `BuildingProperties` option more time consuming with no advantage (other than avoiding some redundant code).

In order to avoid having two different parts of the code do the same thing, I have instead created a function `calc_floor_to_floor_height` and now both the `geometry_generator` and `building_properties` call that same function when they need it.

martin-mosteiro:
While trying to update the unit test data as suggested by @reyery, I noticed that something is wrong with the `create_unittest_data` script. Well, two things are wrong:
1-If you run the script, it fails due to the database check in `schedule_maker` – that's an easy fix.
2-If you run the script (after the easy fix) and run the unit tests, the unit tests fail, even on the master branch. This means that the results generated by `create_unittest_data` are actually incorrect. That requires a somewhat more involved fix, I'm looking into it...

martin-mosteiro:
I have found and solved the issue in `create_unittest_data`, and I have updated the unit test data accordingly.

@lguilhermers I am requesting your review again because I implemented a slightly different version of the change you suggested.

lguilhermers (approved):
Looks good to me, I think it is a neat solution for the floor height calculations.
The test functions I have just looked through it quickly (since Rey already commented on them) and seems fine. Let me know if you want me to actively try it out and see if it works as well.

PR #3160, opened 2022-05-05
3159 unit test for district cooling activation

MatNif:
This unit test is meant to make the testing of the district cooling activation scripts easier and at the same time introduce a control for future adaptations. 

Please check the following things:
1. Test if unit test 'test_district_cooling.py' runs properly by launching it in python IDE (or from terminal?)
2. Check if unit test structure is correct

The unit test was not yet introduced in any of the workflow tests. Would that make sense?


cea/tests/test_district_cooling.py
```diff
@@ -0,0 +1,200 @@
...
+
+class TestDistrictCooling(unittest.TestCase):
+
+    weather_features = None
+    slave_variables = None
+    test_config = None
+    locator = None
+
+    @classmethod
+    def setUpClass(cls, debug=False):
```
reyery, line 30: I don't think the parameter `debug` is required here

cea/tests/test_district_cooling.py
```diff
@@ -0,0 +1,200 @@
...
+        # Storage Cooling
+        cls.slave_variables.Storage_cooling_on = \
+            json.loads(cls.test_config.get("input_storage", "Storage_cooling_on"))
+        cls.slave_variables.Storage_cooling_size_W = \
+            json.loads(cls.test_config.get("input_storage", "Storage_cooling_size_W"))
+
+    def test_cooling_main(self):
+        # config-project directory to the reference case
+        project_reference_case = os.path.join(tempfile.gettempdir(), 'reference-case-open')
+        self.config.__setattr__('project', project_reference_case)
```
reyery, line 138: You should be able to just call `self.config.project`

cea/tests/test_district_cooling.py
MatNif: The debug parameter is not really needed in the function definition, so I removed it there but still left it in the body of the function. I would like to use this parameter to print out certain important intermediary results of the optimisation script in the future.
Would there be a better way to do that?

cea/tests/test_district_cooling.py
MatNif: Thanks. I integrated your suggestion.

MatNif:
Not sure why the test is failing now. I can't replicate the error my own PC. @reyery might it have to do with the way the new testing system is reading the config files?

cea/tests/test_district_cooling.py
reyery: I think having it in the function body is enough for now.

reyery:
It's because CEA is installed in the testing environment in the non-developer mode, so adding new files in `cea/tests` need to be included in the `MANIFEST.in` file. I have fixed that and any files added there now will be automatically included.

Issue #3131, opened 2022-04-04
water source VCC activation


MatNif:
**Describe the bug**
In the cooling resource activation script of the optimisation algorithm (cea/optimization/slave/cooling_resource_activation.py), the water source vapour compression chillers are not always activated to their fullest potential:

The general layout assumed by the scripts is the following:

<img width="1085" alt="Screen Shot 2022-04-04 at 11 40 18 AM" src="https://user-images.githubusercontent.com/93093343/161470052-e97fddad-77f1-48ea-9558-66d9b10e0c96.png">
<img width="478" alt="Screen Shot 2022-04-04 at 11 45 29 AM" src="https://user-images.githubusercontent.com/93093343/161470446-67d7ca95-aced-4a3a-842f-ec5321528bb1.png">

If the temperature of the water body allows it, the VCC is bypassed and the water is used for free cooling. And if the cooling demand is larger than the capacity of the cooling components the thermal storage is discharged, otherwise it is charged.

However, this second condition uses `min(Q_wb;C_VCC)` to determine if the cooling energy provided is sufficient. This leads to an underutilisation of free cooling from the water body in case `Q_wb > C_VCC` and `T_wb < T_DC - 2K`. And it also leads to a underutilisation of the VCC in case `C_VCC > Q_wb` and `T_wb > T_DC - 2K`.

**Expected behaviour**
I would suggest changing the conditions so that these two cases are handled adequately.

MatNif:
Here is a more detailed description of what I would suggest as a solution:

<img width="682" alt="Screen Shot 2022-04-14 at 5 29 50 PM" src="https://user-images.githubusercontent.com/93093343/163357612-fa9a9e05-fc1f-4418-9aa1-11fd8c1e76f6.png">
<img width="482" alt="Screen Shot 2022-04-14 at 5 30 27 PM" src="https://user-images.githubusercontent.com/93093343/163357620-612ecb81-240b-4218-8e2a-01fb1e093079.png">

It is important to note that the thermal efficiency is strictly speaking dependent on the VCC and its operating point.

![image](https://user-images.githubusercontent.com/93093343/163370775-5cad19dc-bcf7-4da8-928c-639bd9b85487.png)

Assuming that all the heat from the VCC is absorbed by the cold water cycle (water body) and none of it is dissipated in the air, the following holds true under the 2. law of thermodynamics:

[...]

PR #3161, opened 2022-05-06
3131 capacity correction for vcc & free cooling

MatNif:
**What was changed**

The capacity correction uses water source technologies to a larger extent to cover the cooling demand. How that is performed exactly is explained in issue #3131 . 

**How to fix**

1. Review the code: check that the 4 activation criteria for each of the 4 activation cases are set correctly for both Base VCC (line 205-247) activation and Peak VCC (line 311-353) activation.
2. a) Once pull request #3160 has been merged, set `debug=True` in test_district_cooling.py and run the unit test (either from your IDE or terminal) with the master branch activated. 
2. b) Then run the same unit test but with branch #3131 activated. You should see that the amount of water body cooling used is higher, i.e. the water body potential is used more effectively. The branch should also fix an issue with the thermal storage temperature.


cea/technologies/storage_tank_pcm.py
```diff
@@ -61,8 +61,9 @@ def __init__(self, size_Wh:float, database_model_parameters, T_ambient_K:float,
         # initialize variables and properties (empty storage)
         self.T_tank_K = self.T_tank_fully_discharged_K
         self.current_phase = 1  # "1=liquid, 2=phasechange, 3=solid"
-        self.current_storage_capacity_Wh = 0.0  # since the simualtion start at mid-night it is considered that the storage is half full by then.
-        self.current_thermal_gain_Wh = self.current_storage_capacity_Wh * (1.05 ** (self.hour - self.hour_of_last_activation) - 1)  # Assumption of losses of X% when discharged during this hour. TODO: Check this value with an example
```
lguilhermers, line 65: There seems to be a 5% factor (losses?) here which is not considered by your changes. Was that intentional?

cea/optimization/slave/cooling_resource_activation.py
```diff
@@ -148,145 +148,266 @@ def cooling_resource_activator(Q_thermal_req,
...
-            E_BaseVCC_WS_req_W = calc_vcc_operation(Q_BaseVCC_WS_gen_W,
+            Qc_BaseVCC_WS_gen_W = Qc_BaseVCC_WS_gen_directload_W + Qc_BaseVCC_WS_gen_storage_W
+
+        # Activation Case 3: The water temperature allows for free cooling, therefore the VCC is bypassed.
+        # The unmet cooling demand is larger than the water body's cooling potential, hence the storage is discharged.
+        elif T_district_cooling_supply_K - T_source_average_water_body_K >= DT_COOL \
+                and Q_cooling_unmet_W >= Qc_water_body_remaining_W:
+            FreeCooling_WS_activated = True
+            Qc_FreeCooling_WS_directload_W = Qc_water_body_remaining_W
+            Qc_water_body_to_storage_W = 0.0
```
lguilhermers, line 233: This value is declared as 0.0 and I missed to find where it changes (since it wouldn't make sense to remain 0 throughout the loops). Can you point me to where this happens?

lguilhermers (commented):
Hi Mathias,

I took a look at the code changes and all seems fine/logical to me (although it is quite complex, so I stayed more in a superficial analysis). 
I did some comments to understand better the code and suggest some improvement (if you think it is applicable).
Since I wasn't able to run the unitest, do you want to ask someone else to try that? I think it will more staightforward than to find out where the issue is comming from.

cea/optimization/slave/cooling_resource_activation.py
```diff
@@ -148,145 +148,266 @@ def cooling_resource_activator(Q_thermal_req,
...
+        # TODO: Replace the current calculation of the thermal efficiency (Carnot efficiency) to a more realistic value
+        thermal_efficiency_VCC = T_district_cooling_supply_K / T_source_average_water_body_K
+        capacity_PeakVCC_WS_W = master_to_slave_variables.WS_PeakVCC_size_W
+        Qc_output_PeakVCC_WS_max_W = min(capacity_PeakVCC_WS_W, thermal_efficiency_VCC * Qc_water_body_remaining_W)
+
+        # Activation Case 1: The water temperature doesn't allow for free cooling, therefore the VCC is activated.
+        # The unmet cooling demand is larger than the maximum VCC output, therefore the storage is discharged.
+        if T_district_cooling_supply_K - T_source_average_water_body_K < DT_COOL \
+                and Q_cooling_unmet_W >= Qc_output_PeakVCC_WS_max_W:
+            PeakVCC_WS_activated = True
```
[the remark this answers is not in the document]
lguilhermers: Ah yes, true. I will try to take a look on this other PR this week as well.

cea/optimization/slave/cooling_resource_activation.py
MatNif: These few lines of code describe the activation case, where water from the water body is used directly to cover the cooling demand (free cooling) but the cooling capacity of the water body is large enough to cover the demand (i.e. the there is no leftover capacity for storage). If this activation case is applied for a specific time step, the value does not change in the rest of the activation chain (the cooling capacity of the water body is used up for the time step). 
However, in the next time step the whole function get's called again and a different activation case might apply.

cea/optimization/slave/cooling_resource_activation.py
```diff
@@ -148,145 +148,266 @@ def cooling_resource_activator(Q_thermal_req,
...
+
+        # Activation Case 4: The water temperature allows for free cooling, therefore the VCC is bypassed.
+        # The water body's cooling potential is larger than the unmet cooling demand, therefore the storage is charged.
+        elif T_district_cooling_supply_K - T_source_average_water_body_K >= DT_COOL \
+                and Q_cooling_unmet_W < Qc_water_body_remaining_W:
+            FreeCooling_WS_activated = True
+            Qc_FreeCooling_WS_directload_W = Q_cooling_unmet_W
+            Qc_water_body_to_storage_W, Qc_DailyStorage_content_W = \
+                daily_storage_class.charge_storage(Qc_water_body_remaining_W - Q_cooling_unmet_W)
+            Qc_from_storage_W = 0.0
```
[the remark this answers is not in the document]
MatNif: Similar logic applies here, as above. However, the `Qc_from_storage_W` variable-name is actually reused in the activation of each technology and denotes the discharge of the storage as technology X is activated. So even though the amount of cooling energy drawn from the storage during the activation of technology X does obviously not change one we move further down in the activation change, the variable's value might change in the activation of other technologies down the line (e.g. air-source VCCs).

cea/technologies/storage_tank_pcm.py
MatNif: Yes, that was intentional. I mistakenly added these statements to the class description a while back, since I thought it would be called again every time a `storage_tank_pcm` class object is modified, these lines would be triggered again.

The lines replacing this section are [194, 244, 250] when the storage is charged and [267, 310, 316] when the storage is discharged.

MatNif:
Thank you very much for reviewing @lguilhermers.

I'm not sure how to resolve the issue with the unit-test. But I think we should definitely try, because the unit-test might be relevant for checking quite a few pull-requests down the line. Maybe @reyery could help us with finding the reason for the unit-test giving you different results than it is giving me. He knows the script a bit after all, from reviewing the branch where it was introduced.

lguilhermers:
All clear to me in your replies, thanks! Let's cordinate a meeting with @reyery to try to solve the issue I am having with the unitest then.

lguilhermers:
After re-installing CEA, the unites works for me again. No idea why but the turn off/ turn on approach solved the issue I was having.

PR #3168, opened 2022-05-12
Run tests when pushed to master

reyery:
Runs `cea test --workflow slow` when commits are pushed to master


shanshanhsieh:
does it mean that we don't need jenkins? @reyery

reyery:
Potentially yes. I'm trying to test if this works and if it is a viable option.

reyery:
I will just merge this first, since it does not affect the main code and it's hard to test this unless it is merged.

Issue #3158, opened 2022-04-29
docker image build error


shanshanhsieh:
Currently, the Pipeline for CEA docker image on Jenkins is failing.
Here is the error msg:
```
#10 [cea-build 3/8] RUN conda env create -f /tmp/environment.yml -n cea && conda clean -afy
#10 sha256:80f83bb36f36c6ce8daef02860cc07448823ec2489881e56cbc38a03f512b7a2
#10 78.05 Solving environment: ...working... 
#10 CANCELED
------
 > [daysim-build 4/5] RUN mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_HEADLESS=ON -DOpenGL_GL_PREFERENCE=GLVND ../Daysim && make ds_illum && make epw2wea && make gen_dc && make oconv && make radfiles2daysim && mv ./bin /Daysim_build:
------
executor failed running [/bin/sh -c mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_HEADLESS=ON -DOpenGL_GL_PREFERENCE=GLVND ../Daysim && make ds_illum && make epw2wea && make gen_dc && make oconv && make radfiles2daysim && mv ./bin /Daysim_build]: exit code: 2
```

any idea on what could be the issue? @reyery ?

PR #3172, opened 2022-05-18
Docker daysim fix

reyery:
This solves #3158


reyery:
Now running `docker build . -t cea:test` should work now

reyery:
@shanshanhsieh maybe we should test if it works in singularity on the HPC as well

shanshanhsieh:
@reyery , thanks for the fix! i will test it this week!

PR #3175, opened 2022-05-18
Move daysim binaries

reyery:
Daysim binaries are moved closer to the radiation script so it would be easier for the script to find them. The priority of paths the check would still be the path provided by the user in the config file, then the ones shipped with the script.


reyery:
This PR would also fix the "merge to master" test in github actions, since currently running `cea test --workflow slow` in Linux and Mac does not work. And moving the binaries to a fix position near the radiation script fixes this.

martin-mosteiro:
@reyery Tested locally on my Mac by running `cea test --workflow slow`` as requested, the script ran all the way through:

```
[...]
Plotted 'Pumping Duration Curve' to /var/folders/fj/jhstllfn1n9cmkpr9bnqypkc0000gq/T/reference-case-open/baseline/outputs/plots/new_basic/thermal-networks/DC__pumping-duration-curve.html
Loading plot_div from cache: /var/folders/fj/jhstllfn1n9cmkpr9bnqypkc0000gq/T/tmpevyqigk8/.cache/thermal-networks/pumping-duration-curve/049fb64952dd739aad1da8f0f0f0d785.div
    - plotting div (len=244892)

----------------------------------------
Script completed. Execution time: 78.21s
--------------------------------------------------------------------------------
Execution time: 5245.45s
```

PR #3187, opened 2022-06-22
3137 VCC model update

MatNif:
This change brings the VCC model back to its original state. Where the VCC's COP is assumed to correspond to a certain percentage of the Carnot COP. This model clearly overestimates the efficiency of VCCs, but in contrast to the formerly established model is at least mathematically sound. I will attempt to introduce a more realistic model in the future.

To test simply:
1. Run the district cooling unit test in debug  mode, with a breakpoint set at line 183.
2. Once the debugger reaches line 183 compare the test values to the expected values. 

The values should differ, since I have not yet updated the expected values (they are still the same as in the current master branch). You should be able to observe:
A. that the electricity demand of the VCCs is significantly lower than before. This can be explained through the overestimation of the COP that the new model calculates.
B. that there is a slight change in most cooling dispatch amounts (except Tirgen direct load). This is due to a higher cooling load being covered by the WS VCCs in cases, where the water body cooling potential is restricted and the resulting shift in the activation chain.

Once you confirm these observations and agree to continuing with the merge, I will update the expected results of the unit test.


MatNif:
Checks are failing for now, since I have not yet updated the expected values in the district cooling unit test.

cea/technologies/chiller_vapor_compression.py
```diff
@@ -77,22 +77,35 @@ def calc_COP(T_cw_in_K, T_chw_re_K, q_chw_load_Wh):
...
-    Pump for Heating Service Using a “Hardware-In-The-Loop” System
-    TODO: Continue looking for a better source or change the formula (possibly check ASHRAE Standard 90.1-2001 on
-          chiller rated efficiency if available)
+    Calculate the approximate COP at rated operating conditions using the g-value (sometimes also called
+    the second-law efficiency [Bejan, 2016]).
+    Assuming a rated COP for all calculations is a strong simplification, but accurate enough for most cases in CEA.
+
+    [Bejan, 2016] Adrian Bejan, 2016, Andvanced engineering thermodynamics (p. 106)
     """
-    PLF = calc_averaged_PLF(peak_cooling_load, q_chw_load_Wh, T_chw_sup_K, T_cw_in_K, VC_chiller)
```
lguilhermers, line 91: Do you know if this function will still be used somewhere in CEA? If not, isn't it better to remove it?

lguilhermers:
When doing the test, only district_cooling_fixed_costs and district_cooling_capacity_installed are matching exactly, as expected.

In the generation dispatch, values are slightly different, as expected ('Q_Trigen_NG_gen_W' is different although 'Q_Trigen_NG_gen_directload_W' is the same, is this as expected as well?)

In the electricity_requirements_dispatch, there is a great reduction for VCC, as expected.

In fuel_requirements_dispatch, there is a small difference in 'NG_Trigen_req_W', I assume this is also as expected and caused by differences in the generation dispatch (probably from 'Q_Trigen_NG_gen_directload_W', is that so?)


Also look at the code, didn't spot any issues, just made a question for clarification

cea/technologies/chiller_vapor_compression.py
MatNif: This function is still used in one of the unit-tests. But is no will not in any part of the CEA scripts. I've been contemplating what to do about functions like this. The function is a valid implementation of what is presented on comnet.org (which is a partial replica of the ASHRAE-norms on cooling systems). So my thinking is it might come in handy at some other point in time.

Is there maybe some sort of a repository, where we could move functions like this to while they are not being used in the main code of CEA? Or can we create one?

lguilhermers: @shanshanhsieh @reyery I remember there was some legacy code in CEA for cases like this. Should we still follow this approach for cases like this?

lguilhermers:
> > In the generation dispatch, values are slightly different, as expected ('Q_Trigen_NG_gen_W' is different although 'Q_Trigen_NG_gen_directload_W' is the same, is this as expected as well?)
> 
> Yes, this is expected. Since Trigen is the first technology to be activated and the demand curve doesn't change, the directload remains the same before and after the change in the VCC-function. However, since the Trigen can also be activated to fill the cold storage, the total cooling generated by the Trigen can change as charge-levels of the storage change with the higher efficiency of the VCCs.
> 
> > In fuel_requirements_dispatch, there is a small difference in 'NG_Trigen_req_W', I assume this is also as expected and caused by differences in the generation dispatch (probably from 'Q_Trigen_NG_gen_directload_W', is that so?)
> 
> It is due to the change in 'Q_Trigen_NG_gen_W' not 'Q_Trigen_NG_gen_directload_W', but I assume that is what you meant.

-Got it, makes sense!
- Yeah, named the wrong variable while writing haha

cea/technologies/chiller_vapor_compression.py
MatNif: After asking the question on the developers team slack channel, the decision was to delete the unused functions of the script.

MatNif (commented):
The unit test for VCC chillers was removed alongside the PLF-functions here since this test 'only' checked the accuracy of the PLF-based model.

Issue #3163, opened 2022-05-09
Additional objective functions for DC optimisation


MatNif:
**Describe the feature you would like to have**
With this new feature two objective functions will be added to the optimisation script for district cooling networks. Specifically, the two objective functions to be added are anthropogenic heat emissions and DC energy demand (i.e. electricity and gas demand of all district cooling technologies).
In order to make the selection of relevant objective functions easy for the user, a new line will be added to the 'Optmization: Supply System Part II (centralized)' script page of dashboard offering a multi-selection field (similar to the technology selection).
Whether this feature should be introduced into the master is up for debate.

**Describe the end purpose of this feature**
CEA users might have varying priorities when it comes to the selection of an 'optimal' district cooling network. This is already taken into account to some extent with the Pareto-curve representation of solutions (cost- vs. emissions-minimisation).
However, in discussions with public authorities it became evident, that there was also a strong interest in knowing which solutions minimise anthropogenic heat emissions and DC energy demand.

**Describe the end-user of this feature**
All users interested in optimisation: e.g. city planners, utility companies, researchers ...

PR #3204, opened 2022-07-20
3163 new objective functions for dc

MatNif:
As described in #3163 this branch introduces 'heat release' and 'system energy demand' as new objective functions of the centralised optimisations (i.e. Supply System Part II) with **district cooling** networks. 

To test:

1. Choose any CEA-scenario
2. Run everything up to the centralised optimisation script
3. Select a set of objective functions you would like run the optimisation with:
Screen Shot 2022-07-20 at 11 23 04 AM

4. Run the script.
5. Check 'scenario/outputs/data/calibration/clustering/checkpoints/ind_%(ind_num)s_buildings_building_scale_performance.csv' to see evolution of objective functions throughout the different generations of the genetic algorithm.


martin-mosteiro:
Not 100% sure what Step 2 actually entails, but I did the following:
1. created a scenario
2. ran solar radiation
3. ran the schedule maker
4. ran the demand tool
5. ran all the energy potential tools
6. ran decentralized optimization
I have now been running centralized optimization for a while, but there's no folder `outputs/data/calibration/clustering/checkpoints`. @MatNif?

MatNif:
Sorry, @martin-mosteiro , my bad. 
The path I indicated above corresponds to the the locator-function description, but that is outdated (see image below). 
Screen Shot 2022-08-04 at 12 47 46 PM

The path should be `[scenario]/outputs/data/optimization/slave/gen_[gen_num]/generation_[gen_num]_individuals.csv` instead. 
By default this file is only created once, at the end of the optimisation run. To get one printed for each generation you can activate the debug button.

martin-mosteiro:
Thanks, @MatNif. Could you please update that function's documentation accordingly?

That folder didn't exist in my case study either, though, which made me notice issue #3214. I now also ran the thermal network simulation first, which is a prerequisite for the centralized optimization tool Here's the updated list of steps, then:
1. created a scenario
2. ran solar radiation
3. ran the schedule maker
4. ran the demand tool
5. ran all the energy potential tools
6. ran all thermal network tools
7. ran decentralized optimization
The folder `outputs/data/optimization/slave` still doesn't exist on my case study, but at least things are happening so far and folders/files are being created, so I'll let it run for a while and check later.

martin-mosteiro:
When running from the dashboard, the optimization appears to get stuck on the first configuration:
```
PRE-PROCESSING
PRE-PROCESSING 0/4: initialize directory
PRE-PROCESSING 1/4: weather features
PRE-PROCESSING 2/4: conversion systems database
PRE-PROCESSING 3/4: feedstocks systems database
PRE-PROCESSING 4/4: network features
3 seconds process time for Network summary for configuration 111111111111111
SUPPLY SYSTEMS OPTIMIZATION
```
and on the Terminal I get the following error over and over again:
```
Process SpawnPoolWorker-16:
  File "/Users/martin/opt/miniconda3/envs/cea/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/Users/martin/opt/miniconda3/envs/cea/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/martin/opt/miniconda3/envs/cea/lib/python3.8/multiprocessing/pool.py", line 114, in worker
    task = get()
  File "/Users/martin/opt/miniconda3/envs/cea/lib/python3.8/multiprocessing/queues.py", line 358, in get
    return _ForkingPickler.loads(res)
AttributeError: Can't get attribute 'Individual' on <module 'deap.creator' from '/Users/martin/opt/miniconda3/envs/cea/lib/python3.8/site-packages/deap/creator.py'>
```

I tried running the optimization from Pycharm and it seems to work so far. I discovered a minor issue, which I fixed (see commit above)

lguilhermers:
> When running from the dashboard, the optimization appears to get stuck on the first configuration:
> 

I managed to run the optimization in dashboard but my scenario is very simple (9 buildings) , not sure if that is why.

martin-mosteiro:
So, after running the optimization for the heating case I get some results:
image

However, when I run the cooling case, I get just two generations in the `slave` folder (84 and 100), and the latter doesn't even contain any `ind_%(ind_num)s_buildings_building_scale_performance.csv` file (in fact, it only contains two files: `gen_100_total_performance_pareto.csv` and `generation_100_individuals.csv`). There are no error messages so I don't know if this is expected behavior?

MatNif:
> However, when I run the cooling case, I get just two generations in the `slave` folder (84 and 100), and the latter doesn't even contain any `ind_%(ind_num)s_buildings_building_scale_performance.csv` file (in fact, it only contains two files: `gen_100_total_performance_pareto.csv` and `generation_100_individuals.csv`). There are no error messages so I don't know if this is expected behavior?

If you ran the optimisation with the `debug = False` config, I believe that only getting 'check-point'-generations and the final generation as output folders is indeed expected. I'll try to double check that early next week just to be sure.

But having no `ind_%(ind_num)s_buildings_building_scale_performance.csv`-files in your generation 100 folder would mean that no individuals improving on your objective functions have been found since the last checkpoint. This is not impossible, but quite unlikely.  Could you run the optimisation once more with `debug = True` and could we meet up some time next week so that I can have a closer look at your results? 

> 
> On a separate note, it seems like every time you run the optimization the previous run's results are deleted. So since there is no case where you can run the cooling and heating system optimization at the same time, there seems to be between 5 and 8 unnecessary columns in each results file?

And this is a very good point. I'm planning to overhaul the 'evaluation'-portion of the optimisation siginificantly over the next few months, but I'll keep than in mind for the upcoming version.

martin-mosteiro:
Thanks @MatNif. Running in debug mode did generate the desired folders and results files:
image
We can meet next week, or otherwise I can upload these results somewhere. Your choice!

martin-mosteiro (approved):
Seems to be doing what it's meant to do. Can't provide much more detailed feedback, but it seems to be ok!

lguilhermers:
I have run 3 cases, with different objectives, getting the following results:

image

The results are as expected, as I get smaller values in the outputs which are part the objective functions (I highlighted those in yellow).
The last case seems to converge to a global minimum, which I think makes sense since both objectives are highly correlated.
Let me know if you think testing more scenarios would be useful.

From this analysis, I feel the code is doing what is proposed, so I will approve. 
I will add some more general comments below, which are more related to the usability rather than the performance.

lguilhermers:
-  The default config for Singapore is set for District Heating, should be District Cooling.
-  The default config for the objectives are system energy demand and anthropogenic heat. I think making it cost and system energy demand would make more sense, as it is a more complex optimization (non correlated objectives).
- The explanation says `When DH is selected, the two set objective functions are 'GHG_emissions' and 'cost'` this gives me an idea that the others are still available, just not set. I would rephrase to something like:
`When DH is selected, the only two available objective functions are 'GHG_emissions' and 'cost'`.
- The simulation gives a non-clear error for single objective runs. Maybe it could have a check for that and throw an early error, or just a description below the selection part.
- for cold-storage-type, it seems it could use a drop down arrow for selection. Even if it is the only available option for now, it is better than having as a editable textbox.
- The error due to missing decentralized optimization was not very informative, as it happens to some of the energy potentials.

lguilhermers (approved):

shanshanhsieh:
this looks nice, @MatNif  and great testing report, @martin-mosteiro @lguilhermers !!!
i would be happy to test, aiming to finish by 17th Sep
didnt mean to stall your process, feel free to merge if all of you think it's working as intended

MatNif:
> this looks nice, @MatNif and great testing report, @martin-mosteiro @lguilhermers !!! i would be happy to test, aiming to finish by 17th Sep didnt mean to stall your process, feel free to merge if all of you think it's working as intended

Hi @shanshanhsieh,

I think it's alright. Yes, @lguilhermers and @martin-mosteiro have done a great job in testing and made some very valid comments that I will try to address in the next iteration of the optimisation algorithm. 
So rather than testing this branch, if you have some time, I would really appreciate it if you could have a look at the document I sent on the Slack channel, that explains the upcoming changes. Or if you prefer, I'm very happy to guide you through it in a call as well.
I plan on making these changes over the next 2 months, so it would likely make sense to have another round of testing at the end of this.

@lguilhermers, @martin-mosteiro: Could one of you merge this branch?

martin-mosteiro:
Good job @MatNif!

PR #3208, opened 2022-07-26
fixing overlapping geometries

MatNif:
This change allows one to eliminate overlapping geometries using a toggle-option in the zone-helper.

To test:
1. Select an area you know has overlapping building geometries (on the map when creating new scenario).
2. Run the zone-helper with the 'fix-overlapping-geometries' set to on under settings.

Alternatively, if you are unsure which areas contain overlapping building geometries, select random dense area (preferably with high and architecturally complex buildings) and run once with the toggle button on and once with toggle off and compare the two.


lguilhermers:
I tested for a small area in CBD (Singapore) and it worked as expected.
There were many podiums and towers which would normally overlap and now they were divided with no overlap in their geometry. It also did this in efficient time (it was a small case, so I can't guarantee it scales well for large scenarios).

image

Since it seems to work as expected, I will merge with master.
@reyery feel free to check the code in more detail, I have looked quickly and have no comments on that.

lguilhermers (approved):

Issue #3211, opened 2022-08-01
CRS mismatch between zone and surroundings


MatNif:
**Describe the bug**
I recently created a zone.shp file using the zone helper and had to do some post-processing of the building geometries using QGIS. In the process I must have changed the coordinate reference system (CRS) of the zone.
When I then launched the terrain helper I received the following error.

```
generating terrain from Surroundings area
Proceeding to calculate terrain file with fixed elevation in m of  1
Traceback (most recent call last):
  File "/Users/mathiasniffeler/Documents/GitHub/CityEnergyAnalyst/cea/worker.py", line 146, in worker
    run_job(config, job, server)
  File "/Users/mathiasniffeler/Documents/GitHub/CityEnergyAnalyst/cea/worker.py", line 108, in run_job
    script(config=config, **parameters)
  File "/Users/mathiasniffeler/Documents/GitHub/CityEnergyAnalyst/cea/api.py", line 60, in __call__
    self._runner.__call__(*args, **kwargs)
  File "/Users/mathiasniffeler/Documents/GitHub/CityEnergyAnalyst/cea/api.py", line 38, in script_runner
    script_module.main(config)
  File "/Users/mathiasniffeler/Documents/GitHub/CityEnergyAnalyst/cea/datamanagement/terrain_helper.py", line 141, in main
    terrain_elevation_extractor(locator, config)
  File "/Users/mathiasniffeler/Documents/GitHub/CityEnergyAnalyst/cea/datamanagement/terrain_helper.py", line 102, in terrain_elevation_extractor
    calc_raster_terrain_fixed_elevation(crs, elevation, grid_size, raster_path, locator,
  File "/Users/mathiasniffeler/Documents/GitHub/CityEnergyAnalyst/cea/datamanagement/terrain_helper.py", line 110, in calc_raster_terrain_fixed_elevation
    cols = int((x_max - x_min) / grid_size)
ValueError: cannot convert float NaN to integer 
```

This is due to the terrain helper combining the zone and the surroundings into a single GeoDataFrame (GDF), assuming that their CRS are identical.  The subsequent projection of the combined data frame to a new CRS can therefore cause the projected values of the surroundings to land outside the range of admissible values.

**Expected behaviour**
This is a clear edge case, but to avoid this error in the future, the terrain helper should guarantee that the zone and surroundings are projected to the same CRS before combining the two GDFs.

PR #3212, opened 2022-08-01
Handling potential CRS mismatch

MatNif:
This fixes the error described in issue #3211 .

To review:
1. Check code changes. (there are very few of them, so it should be fairly simple)
OR
2. Purposely change the CRS of one of the two .shp files right before running the terrain helper. See if the error still occurs.


lguilhermers:
I just had this issue with a scenario that Rey sent to me. Your branch seem to have fixed it =D.

reyery (approved):

Issue #3218, opened 2022-08-10
Retrieve buildings heights from OSM


lguilhermers:
I checked a few buildings on OSM which have height informed. CEA seems to ignore this information, fetching always the number of floors and assuming a floor height of 3m to calculate height. Wouldn't this be preferable only if the height information was not available?

@shanshanhsieh @reyery @martin-mosteiro please let me know if this was discussed in the past and if it makes sense to implement such feature in the zone helper.

martin-mosteiro:
Fully agree, the height should be taken from OSM if available

Issue #3219, opened 2022-08-10
Surrounding buildings does not use number of floors


lguilhermers:
I thing the only feature requires by surroundings is its height (for solar radiation). Building floor seems to be used with floor height to calculate this parameter, but if height is taken directly from OSM (see my other issue), number of floors isn't even required.

Even in the case number of floor is used to calculate, they become independent parameters in the input editor. So I am in favor of removing this parameter from the input editor, leaving only height as an attribute. 
Please let me know if this makes sense.

martin-mosteiro:
I don't mind either way. Even if number of floors is more intuitive for users, if they want to change the number of floors in the surrounding buildings they'll have to adjust the building height anyway, so it doesn't make a difference imo.

lguilhermers:
Exactly, I think it is a parameter that doesn't impact any calculation and therefore, we don't need it. 
I will confirm if that is the case and, if so, remove it.

PR #3222, opened 2022-08-10
Gets building height from OSM in zone helper

lguilhermers:
Fixes issue #3218.
This is a quick fix to make CEA get height from OSM when attribute 'height' is existent.

Question: Should the same process be applied in lines 100, 103 and 106? I am a bit confused about this particular part, which seems to be for when the user has defined the geometry.

Comment: The same can be done for surrounding building and I plan to address this in issue #3219

### DIFF
--- a/cea/datamanagement/zone_helper.py
+++ b/cea/datamanagement/zone_helper.py
@@ -90,8 +90,9 @@ def clean_attributes(shapefile, buildings_height, buildings_floors, buildings_he
             np.nanmedian(data_floors_sum_with_nan))  # median so we get close to the worse case
         shapefile["floors_ag"] = [int(x) if not np.isnan(x) else data_osm_floors_joined for x in
                                   data_floors_sum_with_nan]
+        print(shapefile.info())
         if 'height' in list_of_columns:
-            shapefile["height_ag"] = shapefile['height'].fillna(shapefile["floors_ag"] * constants.H_F)
+            shapefile["height_ag"] = shapefile["height"].fillna(shapefile["floors_ag"] * constants.H_F).astype(float)
         else:
             shapefile["height_ag"] = shapefile["floors_ag"] * constants.H_F
     else:
--- a/cea/datamanagement/zone_helper.py
+++ b/cea/datamanagement/zone_helper.py
@@ -100,6 +100,9 @@ def clean_attributes(shapefile, buildings_height, buildings_floors, buildings_he
             #  Replaces 'nan' values with CEA assumption
             shapefile["height_ag"] = shapefile["height"].fillna(shapefile["floors_ag"] * constants.H_F).astype(float)
             #  Replaces values of height = 0 with CEA assumption
+            # TODO: Check whether buildings with height between 0 and 1 meter are actually mostly underground
+            #  The radiation script cannot process buildings with height 0m (underground buildings) those at the moment.
+            #  Once the radiation script can process underground buildings, this step might need to be revised.
             shapefile["height_ag"] = shapefile["height_ag"].where(shapefile["height_ag"] != 0,
                                                                   shapefile["floors_ag"] * constants.H_F).astype(float)
 
--- a/cea/datamanagement/zone_helper.py
+++ b/cea/datamanagement/zone_helper.py
@@ -91,8 +91,7 @@ def clean_attributes(shapefile, buildings_height, buildings_floors, buildings_he
         shapefile["floors_ag"] = [int(x) if not np.isnan(x) else data_osm_floors_joined for x in
                                   data_floors_sum_with_nan]
         if 'height' in list_of_columns:
-            shapefile["height_ag"] = shapefile['height'].fillna('0')
-            shapefile["height_ag"] = shapefile["height_ag"].where(shapefile["height_ag"]==0,shapefile["floors_ag"] * constants.H_F)
+            shapefile["height_ag"] = shapefile['height'].fillna(shapefile["floors_ag"] * constants.H_F)
         else:
             shapefile["height_ag"] = shapefile["floors_ag"] * constants.H_F
     else:
--- a/cea/datamanagement/zone_helper.py
+++ b/cea/datamanagement/zone_helper.py
@@ -62,10 +62,15 @@ def clean_attributes(shapefile, buildings_height, buildings_floors, buildings_he
     no_buildings = shapefile.shape[0]
     list_of_columns = shapefile.columns
     if buildings_height is None and buildings_floors is None:
-        print('Warning! you have not indicated a height or number of floors above ground for the buildings, '
+        print('Warning! you have not indicated number of floors above ground for the buildings, '
               'we are reverting to data stored in Open Street Maps (It might not be accurate at all),'
               'if we do not find data in OSM for a particular building, we get the median in the surroundings, '
               'if we do not get any data we assume 4 floors per building')
+
+        print('Warning! you have not indicated height above ground for the buildings, '
+              'we are reverting to data stored in Open Street Maps (It might not be accurate at all),'
+              'if we do not find data in OSM for a particular building, we estimate it based on the number of floors,'
+              'multiplied by a pre-defined floor-to-floor height')
 
         # Check which attributes the OSM has, Sometimes it does not have any and indicate the data source
         if 'building:levels' not in list_of_columns:
@@ -92,7 +97,12 @@ def clean_attributes(shapefile, buildings_height, buildings_floors, buildings_he
                                   data_floors_sum_with_nan]
 
         if 'height' in list_of_columns:
+            #  Replaces 'nan' values with CEA assumption
             shapefile["height_ag"] = shapefile["height"].fillna(shapefile["floors_ag"] * constants.H_F).astype(float)
+            #  Replaces values of height = 0 with CEA assumption
+            shapefile["height_ag"] = shapefile["height_ag"].where(shapefile["height_ag"] != 0,
+                                                                  shapefile["floors_ag"] * constants.H_F).astype(float)
+
         else:
             shapefile["height_ag"] = shapefile["floors_ag"] * constants.H_F
     else:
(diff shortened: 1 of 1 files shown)


lguilhermers:
I just noticed that the geometries get flat in the 3D view, not sure yet what causes it. The values seem fine.
image

cea/datamanagement/zone_helper.py
```diff
@@ -90,7 +90,11 @@ def clean_attributes(shapefile, buildings_height, buildings_floors, buildings_he
             np.nanmedian(data_floors_sum_with_nan))  # median so we get close to the worse case
         shapefile["floors_ag"] = [int(x) if not np.isnan(x) else data_osm_floors_joined for x in
                                   data_floors_sum_with_nan]
-        shapefile["height_ag"] = shapefile["floors_ag"] * constants.H_F
+        if 'height' in list_of_columns:
+            shapefile["height_ag"] = shapefile['height'].fillna('0')
+            shapefile["height_ag"] = shapefile["height_ag"].where(shapefile["height_ag"]==0,shapefile["floors_ag"] * constants.H_F)
```
martin-mosteiro: A couple of things:
1. In line 94, NaNs are filled as a string, but in line 95 the condition is specified as an int. I don't think this is desired behavior, correct?
2. More importantly, this `where` appears to do exactly the opposite of what it's supposed to do: it's filling in all the values that were _not_ NaN as equal to `shapefile["floors_ag"] * constants.H_F`. Or at least that's what happens when I run this code on my computer?
3. I think these two lines can be written in a more foolproof way:
`shapefile["height_ag"] = shapefile['height'].fillna(shapefile["floors_ag"] * constants.H_F)`

martin-mosteiro:
@lguilhermers I believe I found an issue in the code you wrote (see my comment). This is the shapefile after the first line in your code:
Screen Shot 2022-08-10 at 16 37 08
And this is the same shapefile after the second line of your code:
Screen Shot 2022-08-10 at 16 37 34
The expected output (which my proposed fix produces) is the following:
Screen Shot 2022-08-10 at 16 42 56

martin-mosteiro (changes requested):
@lguilhermers In principle, I agree with the change, but I found an issue in the implementation. I have fixed it, please check if you agree with it!

martin-mosteiro:
Also, note that the heights were being imported as strings, not floats (even in my last picture). I just fixed that as well:
Screen Shot 2022-08-10 at 16 59 15

lguilhermers:
Thanks for the fixes Martin, indeed it slipped through me that the founded values were also getting replaced.
I think that, besides what you implemented, the cases with height 0 should also be handled by the origal CEA formula to avoid errors. That was what the `where` should do, but it seems it needs a better working solution.

PS: actually some of those height = 0 are actually underground buildings, which CEA does not handle now anyway. So they technically shouldn't be there, but I think is better that this check is done manually by the user than automated to leave this building out, for example.

martin-mosteiro:
Hm, I guess then the question is what should we do with underground buildings. If CEA can't process them, then I guess they shouldn't be imported at all (rather than being given an artificial height)?

Which part of CEA isn't able to process underground buildings, though? I guess solar radiation?

lguilhermers:
Yes, solar radiation. 

For now I have included a fix to treat height =0 the same as nan, using CEA's old method. I am not sure how reliable and consistent is 'height=0' in OSM and want to avoid cases where real buildings don't get generated.

Perhaps we treat the underground issue separately by opening a new one on how to treat underground buildings? @MatNif was also experiencing issues with that in his case study in MBS and can have some ideas about that.

martin-mosteiro:
Hm... I guess the most consistent way to deal with underground building is to import a "short" (1m or something) building with `WWR = 0` and `Hs_ag = 0`. But yeah, it's probably best to revert back to # of floors in that case.

On a related but actually separate issue, I also think OSM doesn't differentiate underground and above ground floors either, so if OSM says a building is 5 floors it might still be, say, 2 above ground floors and 3 underground. So we might still be importing the wrong height. That appeared to be the case for a lot of Swiss buildings I worked with back in the day.

martin-mosteiro:
Back to the issue at hand, should we maybe go as follows then?
```
shapefile["height_ag"] = shapefile["height"].replace({0: np.nan}).fillna(shapefile["floors_ag"] * constants.H_F).astype(float)
```

MatNif:
> Perhaps we treat the underground issue separately by opening a new one on how to treat underground buildings? @MatNif was also experiencing issues with that in his case study in MBS and can have some ideas about that.

Yes, the issue I encountered with underground buildings in the radiation script is in the portion where CEA tries to determine which portion of the building lies under the terrain surface (set at 1m height by default). If the script can't find that intersection it ends up crashing. 
A simple work-around for this portion would be to select the buildings by height range [height_bg,height_ag] to determine which ones should even be considered for the terrain intersection portion of the script. For the buildings with a height range that is entirely under the terrain line we could then simply fill the radiation results with 0s.

lguilhermers:
> Hm... I guess the most consistent way to deal with underground building is to import a "short" (1m or something) building with `WWR = 0` and `Hs_ag = 0`. But yeah, it's probably best to revert back to # of floors in that case.

Yeah, I agree. I will open another issue and put those ideas there.

> On a related but actually separate issue, I also think OSM doesn't differentiate underground and above ground floors either, so if OSM says a building is 5 floors it might still be, say, 2 above ground floors and 3 underground. So we might still be importing the wrong height. That appeared to be the case for a lot of Swiss buildings I worked with back in the day.

Yeah, but this we have no idea about. Either OSM has to provide both information or we probably just assume everything is above ground (as it is now). I think anyway the idea is that the scenario is extract with near-real information but for higher accuracy more data and some manual work will always be required.





> ```
> shapefile["height_ag"] = shapefile["height"].replace({0: np.nan}).fillna(shapefile["floors_ag"] * constants.H_F).astype(float)
> ```

I tested this but it doesn't replace the "0"s. My scenario also has height of 0.5 from OSM, which I think it will also give problems (not 100% sure).

lguilhermers (commented):
I have used this but included a line to handle `height = 0`

martin-mosteiro:
After checking again, I noticed @lguilhermers was right: there's some buildings with infeasible heights, like 0.1 m height but 4 above ground floors:
Screen Shot 2022-08-16 at 15 33 56
However there are some odd but feasible combinations, e.g. 1 m height and 1 above ground floor (probably not a _building_ but some physical component, therefore feasible)
Screen Shot 2022-08-16 at 15 34 04

As a rule of thumb, I suggest the following: any building where `floors_ag > 1` and `height_ag < 1` don't make sense. So rather than replacing all heights that are 0 (maybe there _could_ be a building with 0 floors above ground and a height of 0?), we should probably replace all values where `(floors_ag > 1) & (height_ag < 1)`.

@lguilhermers any thoughts?

lguilhermers:
The additional rules look good. I would go even beyone and implement it for heights < 2, since those would be very particular structures, not really aimed to be buildings.

Regarding those low heights, I feel divided on how to address them. 
Part of me thinks it should give the CEA rule of `floors*floor_height`, as it is currently doing, and can fix buildings which the number of floors was provided on OSM and prevent errors from small geometries.
Another side (which I am leaning the most), is that those heights in OSM are perhaps a good indicator that those are particular stuctures (maybe underground, maybe not a proper building). If that is the case (and happened with me on a case in the CBD), CEA's rule could turn them in very bulky buildings if the buildings around provide a considerable median height.

martin-mosteiro:
I'm okay with importing a 1m tall building. Of course it's not a _real_ building, but it can easily be adjusted to have no demands and can be a potential source of shading. 

I think we shouldn't turn buildings with 1 floor and low heights into 1 floor and 3 meters (i.e., `floors*floor_height`), this would create an arbitrary, artificially large building block.

But you're right: these buildings are potentially actually underground structures, such that they should _only_ be considered as such. Maybe we need to check these cases in a bit more detail before we create ham-fisted rules like the ones we've proposed so far?

martin-mosteiro:
By that I mean: rather than trying to fix "unreasonable" values like 0 m heights or 4 floors in 0.5 m height, maybe we first need to check whether those floors are actually above ground for the cases we have observed so far. If they turn out to be underground, I would suggest keeping the 0m heights and adjusting the solar radiation script as suggested by @MatNif. Though that should probably be its own issue.

lguilhermers:
Yes, I agree with all that you said. So, should we keep the current rules of this PR and open a known issue related to #3224 and this one which also aims to give a good solution for small height geometries?

PS: this current PR still edits height = 0, giving them the CEA rule. This I would keep for now, as height = 0 seems to break the 3D vizualizatitions and will definetly be a source of multiple errors on the code as well.

martin-mosteiro:
I think what broke the 3D visualizations was not height 0, but rater height '0' (i.e. the 'height_ag' data type was string or mixed, not float or int). Did you test what would happen if the height was 0.0?

Either way, I agree this makes sense for now. I'll add a `TODO` label and approve.

lguilhermers:
Ok, makes sense. I haven't tested it further, so it could be.
Thanks for the review!

martin-mosteiro (approved):
This PR imports building heights for buildings with height > 0 m from OSM.
Some buildings have heights of 0 - 1 m in spite of having multiple floors. These might be underground buildings, but the radiation script can't process buildings with height 0 m at the moment. These issues are discussed in #3224 should be addressed in the future.
For now, buildings with height 0m are assumed to be an error and replaced with the original CEA assumption (i.e., building height = number of floors * height per floor)

MatNif (commented):
I think line 107 of the script should say `shapefile["height_ag"] == 0` instead of `shapefile["height_ag"] != 0`. 
Otherwise I believe the changes are good.

martin-mosteiro:
@MatNif counterintuitively, it's the other way around. If you do `shapefile.where(shapefile["height_ag"] == 0)` you get `True` for all values that are _not_ zero, and for `shapefile.where(shapefile["height_ag"] != 0)` you get `True` for all values that are 0. Not sure why.

MatNif:
Oh. That is a bit counterintuitive. But perfect that you checked it. Thanks. In that case I'll also approve the PR.

PR #3227, opened 2022-08-16
Use micromamba instead of miniforge

reyery:
This could help improve the speed of GitHub actions test, by reducing the time needed to prepare the environment.


reyery:
This does not actually speed up conda environment setup that much, since a lot time is still spent extracting the dependencies when it hits cache, compared to solving it normally. Therefore improvements in total time is insignificant as bulk of the time is still spent running the tests. It is still interesting how different the time taken for each OS is when running the CEA test.

I will still merge this since it makes the workflow steps much neater.

PR #3234, opened 2022-08-19
Several fixes to zone_helper to solve issues found after selecting fix_overlapping 

### LINKED ISSUE #3229: Zone helper cannot process `min_heights` of type `str`
**Describe the bug**
When running the zone helper after the merging of the polygon intersection PR (#3208), a `site.shp` file that was previously functional can no longer be used to create a zone shapefile. The reason is that the `min_height` parameter for at least one building has a string value, and hence the zone helper cannot process it.

### LINKED ISSUE #3231: Building intersection option in the zone_helper creates unusable geometry types
**Describe the bug**
While running the `zone_helper` with a previously functional `site.shp` shapefile, geometry types that CEA cannot process are generated, namely: `LineString`, `GeometryCollection`, `MultiPolygon`. The function `flatten_geometries` _could_ be called again to transform `MultiPolygon`s to `Polygon`s, but any other geometry types will simply be dropped. This might lead to useful building footprints being dropped.

martin-mosteiro:
This PR solves a few different issues I found when I tried to run the `zone_helper` again on an old case study while using the new `fix_overlapping` option. The test case was a site shapefile that generated a valid CEA case study months ago, but now caused the following issues:
1. `min_height` could not be converted to `int` because it was imported as a string, not a float (#3229)
2. after splitting the geometries of overlapping buildings, some geometry types that CEA cannot process were created (`LineString`, `MultiPolygon`, `GeometryCollection`) (#3231)

While fixing these issues I furthermore noted that the `zone_helper`, as currently implemented, will do the following with geometry types that cannot be processed by CEA:
1. If it's a LineString or Point, discard them (reasonable)
2. If it's a MultiPolygon, try to convert it to a Polygon by `unary_union`
3. If the result of the union is a Polygon, keep it
4. If the result of the union is a MultiPolygon, keep the first Polygon in the MultiPolygon and drop the rest
However, Step 4 can lead to arbitrarily dropping building geometries (#3231). The other Polygons may be useful parts of the building, and even if you did choose to keep only one, the current approach doesn't ensure that the most representative/useful part of the building is kept, just whatever is first in the list. This problem is exacerbated by the `fix_overlapping` option, as by removing parts of buildings it is highly likely that a MultiPolygon will be formed.

So this PR does the following:
1. convert `min_heights` and `min_levels` to floats before using them – this ensures that we are not passing unusable strings that will cause the script to fail (Resolves #3229)
2. pass the shapefile generated after fixing overlapping geometries back to the `flatten_geometries` function to ensure the resulting shapefile only includes `Polygons` (Resolves the first part of issue #3231)
3. change the `flatten_geometries` function to keep all Polygons in a MultiPolygon or GeometryCollection type while dropping Points and LineStrings (Resolves #3231 – the second part)

The resulting `flatten_geometries` method is much slower than the original one (0.432 s vs 5.921 s for my case study on my computer), but I believe this is a necessary improvement – and also, my case study is city-scale, so a few-second delay for such a large case study is unsurprising and in any case negligible compared to the computational expense for the `fix_overlapping` option.

To test:
1. create a case study with many different building geometries (you can try creating a case study the size of Singapore to replicate my exact case)
2. run the zone helper with `fix_overlapping = True`
3. verify that only Polygon shapes are exported

I'm requesting reviews from @MatNif since he wrote the `fix_overlapping` part of the script and @reyery since he wrote the `flatten_geometries` method.


martin-mosteiro:
@MatNif: I found the issue that was causing the tests to fail. It was a fairly trivial non-issue. All tests pass now.

MatNif (approved):
The changes generally make sense to me, even though I don't really know what the `.get_level_values`-function in `flatten_geometries` does. And in most cases the new script seems to work and yield a more realistic geometry.
I ran into one error though, as I was running the zone_helper on a rather big area in the heart of Tokyo. It looked like this:   

`zone_helper: could not convert string to float: '7;13'`

```
File "/Users/mathiasniffeler/Documents/GitHub/CityEnergyAnalyst/cea/datamanagement/zone_helper.py", line 407, in polygon_to_zone
    shapefile = assign_attributes(shapefile, buildings_height, buildings_floors,
  File "/Users/mathiasniffeler/Documents/GitHub/CityEnergyAnalyst/cea/datamanagement/zone_helper.py", line 89, in assign_attributes
    shapefile[col] = shapefile[col].astype(float)
```

Is there a way we could handle even the most weird of OSM entries?

In my opinion this is not really a hindrance to pulling this branch though, as it does what it is supposed to.

martin-mosteiro:
@MatNif Great catch. Just noticed that I solved that issue on #3238 instead of #3232. But it's actually part of this issue, so I'll add it to #3232 for you to test it again.

Here's the part where that issue is solved:
```
        for col in ['building:min_level', 'min_height', 'building:levels', 'height']:
            if col in list_of_columns:
                try:
                    shapefile[col] = shapefile[col].astype(float)
                except ValueError:
                    for i in shapefile[col].dropna().index:
                        try:
                            shapefile.loc[i, col] = float(shapefile.loc[i, col])
                        except ValueError:
                            print(f'{col} = {shapefile.loc[i, col]} in building {i} cannot be interpreted, replacing with NaN.')
                            shapefile.loc[i, col] = None
```
However, that fix isn't super efficient, as it requires going through all the values in the column one by one. In my test case (all of Singapore) that didn't take that long – but it could be done better. I'm on it!

martin-mosteiro:
@MatNif This should fix the issue you found, feel free to test again!

MatNif:
> @MatNif This should fix the issue you found, feel free to test again!

Yes, nice. This seems to have fixed it.

Issue #3236, opened 2022-08-26
`fix_overlapping_geometry` might select wrong information over right information


martin-mosteiro:
**Describe the bug**
The `fix_overlapping_geometry` function assigns building heights according to the `height_ag` and `height_bg` attributes in the generated shapefile. However, the comparison of two buildings' heights on a CEA shapefile might be apples to oranges: one might have a height assigned from the OSM height or number of floors data, but another might have a height assigned by taking the case study area's median. By making the selection on which height to assign based on this comparison, we might be assigning a totally fake value (the median of the entire case study area) over a more plausible one (an actual height from OSM).

But does this ever really have an effect? Well, look at these two versions of Lau Pa Sat created by selecting two different case study areas: 
<img width="985" alt="Screen Shot 2022-08-26 at 10 17 24" src="https://user-images.githubusercontent.com/18548065/186802425-877ec4be-6892-4821-85e5-0e60de1ff2dd.png">
<img width="985" alt="Screen Shot 2022-08-26 at 10 17 58" src="https://user-images.githubusercontent.com/18548065/186802459-99b5f735-3471-4ecf-9e65-3a630eb27d9b.png">
Note: there are two overlapping polygons for Lau Pa Sat; the "pizza pie" version is the correct height per OSM; the other polygon has no assigned height nor number of floors, and therefore gets assigned the median height of the case study in CEA. When selecting a smaller area, the median height is much higher such that it exceeds the height taken from OSM, and therefore the taller polygon (the one with a completely artificial 17 floor height) gets selected during the `fix_overlapping_geometry` workflow.

**To Reproduce**
1. Find an area with overlapping polygons with different assigned heights (one has an assigned height, the other one doesn't) – yes, it might be tricky to find another case of this.
2. Run the zone_helper for that area
3. Check the output

**Expected behaviour**
Just like in the geometry creation process, data should be used with the following priority:
1. 'height' taken from OSM
2. 'building:levels' taken from OSM
3. 'height_ag'/'height_bg'

@MatNif please let me know if you have any feedback or ideas on this

martin-mosteiro:
Note: the example case in Lau Pa Sat above was generated by including polygons tagged as `building:part` in OSM (separate issue, #3220), but could happen for other overlapping building polygons.

MatNif:
Hey @martin-mosteiro,

First of all, nice find once again.
[...]

PR #3238, opened 2022-08-26
Use shapes tagged as `building:part` in OSM as well as `building` in `zone_helper`

### LINKED ISSUE #3220: OSM: "Building Part" category seems to give a better representation of building geometries
Looking more into OSM, I also noticed some buildings are better represented by a category in OSM named "Building Part" rather than "Building". At least for the central region of Singapore (see below), "Building" tends to specify the whole plot areas, while "Building Part" seems to by quite accurate to describe the polygons of the high rise buildings.

From my experience, there is a tendency to over predict the energy consumption in those buildings with reasonable parameters, and I think a good portion is this simplification of towers to plot areas.

Does it make sense to consider "Building Parts" to evaluate the geometries instead?

PS: For this to be implemented, it would be good to see how consistent this pattern is across different regions (in Singapore and globally).  So far my insights are only coming from a specific area in the Singaporean CBD.

image

martin-mosteiro:
This PR implements the fix suggested by @lguilhermers in #3220, whereby shapes tagged as `building:part` in OSM are imported when running the `zone_helper`. This is implemented as follows:
1. `building` shapes are imported as done in the zone_helper so far
2. `building:part` shapes are imported and merged to the same GeoDataFrame as the buildings
3. The combined shapefile is passed to the `fix_overlapping_geometries` script by @MatNif to ensure buildings split as required

I'm adding a Boolean parameter so users can decide whether they want to use building parts or not. Adding building parts means creating a lot of overlapping geometries, so I am making `fix_overlapping_geometries == True` if `include_building_parts == True`.

Here's the results of this PR:
- 'Shapefile created by the master branch with `fix_overlapping_geometries == False`:'
Screen Shot 2022-08-26 at 00 11 29
- 'Shapefile created after this PR:'
Screen Shot 2022-08-26 at 00 07 13
Note:

1. Lau Pa Sat was originally imported as a  single polygon with no height information – now it is imported as a number of overlapping polygons (see #3236)
2. Two large towers were being imported as much shorter before (solved by #3222), and the complex shape of the Asia Square towers were imported as two blocks with the wrong height
3. Robinson Centre was originally imported as a single big block, even though it's actually a podium and a tower – the new version still has the same volume, but the podium is a separate polygon for which users can adjust the height as required.

In solving this issue, I ended up discovering a few others, as discussed in #3220 and #3232, including the fact that by using `building:part`s some information from OSM was being lost. So now I have added a step to copy any building use type available in the OSM `building` types to the OSM `building:part` types (which generally have no use type information). I furthermore noticed some inconsistencies in the way typologies were assigned:
1. The CEA code claims to "Calculate the most likely use type" in `calculate_typology_file`, but actually any data not imported from OSM is just assigned "MULTI_RES". So I added a step to assign the most common use type to the entire case study area to those buildings (similar to the median used in the heights and number of floors, but in this case I had to use to mode)
2. I also noticed that the `amenity` information from OSM was not being used in assigning the use types, which means that , for example, food courts were being assigned as "RETAIL" (rather than, say, "RESTAURANT"). I have rectified this as well.

How to test this PR:
1. Select an area where OSM includes both polygons tagged as buildings and polygons tagged as building:part – ideally not any of the examples shown above _and_ where the polygon's geometry is not `None` (which was the case or the Zurich example I mentioned in #3220), though finding such an area might be tricky
2. Run the zone_helper
4. Check whether building and building:part polygons are imported as separate buildings

I'm assigning @lguilhermers as a reviewer since he originally requested this feature and had many ideas on how to implement it. I'm also assigning @MatNif since getting this PR to work required some reworking of his code to fix overlapping geometries.

This PR requires #3232 to be merged first and closes #3220


MatNif (approved):
Both, the code and the results it produces, look good to me.

This is a cut-out of Tokio produced without importing building parts. 
Screen Shot 2022-09-01 at 5 42 15 PM

... and here is the same zone with building parts. 
Screen Shot 2022-09-01 at 5 37 44 PM

As the current 'optimisation-guy', I have the optimisation in mind, and for that purpose, I think importing the building parts makes sense for smaller use cases. However, for larger use cases the building parts might almost be too precise, i.e. 'fix-overlapping' might cut them in too many parts. That means, the optimisation problem would need to handle far more 'buildings' (since each part gets treated like a separate building) and thus result in far longer optimisation times.

But that is absolutely no reason not to merge this PR.

It's great that there is an option to turn the building parts based OSM-import off. And if I find the time, I might even introduce a routine into the optimisation algorithm that treats interlaced geometries as a single building for the sake of the optimisation.

martin-mosteiro:
Thanks for the detailed review, @MatNif. I agree, the level of detail can be too much now. As you saw in #3236, Lau Pa Sat, for example, becomes 8 buildings when using `building:part`. The Shoppes at Marina Bay similarly becomes a number of transversal slices of different heights due to the undulating roof shape as well. Perhaps a tool to merge such cases as suggested by @lguilhermers could help, but for now I think a simple on/off switch to let users decide works.

lguilhermers:
@martin-mosteiro after some searching, I found New York is also a good test case, with many Building: Parts in Manhattan area.

First I got some generic error when I tried to take the entire island with `include-building-parts=True`. If you are interested in this, I can try to reproduce and get some more information.

When I selected a smaller region, I could see the expected behaviour: generic contour and often wrong heights for buildings and detailed (very often too detailed) but somehow accurate when using Building: Parts.

Here are some screenshots to compare:

include-building-parts = False:

image
image
image
image
image



include-building-parts = True:

image
image
image
image
image




Note: 
fix-overlapping-geometries didn't seem to make a difference in those cases. I would expect the 'Building' attribute to still be there, under all the detailed geometry which is Generated by 'Building:Part' attributes but it doesn't seem to be the case. 
I couldn't see any changes by having this option as True or False when building parts are on.

Overall it seems to me that this PR is good to go. Besides the error for a large scenario, everything went smooth.

Issue #3254, opened 2022-10-12
Expanding functionalities of centralised optimisation


MatNif:
**Describe the feature you would like to have**
- Allow for the centralised optimisation to investigate the introduction of multiple networks into a given zone.
- Allow for more flexibility in building cooling supply systems (i.e. combine cooling system components freely)
- Allow for a comparison of optimal solutions with the status quo (i.e. the systems given in the input editor). 

**Describe the end purpose of this feature**
This extension of the centralised optimisation script will maintain all of the functionalities of the current script, while giving the user more options to influence the decision space and making the optimisation viable for large scale cases.

**Describe the end-user of this feature**
The end-user of this feature could be anyone needing to design or investigate optimal thermal energy systems. E.g. energy engineers, civil engineers, policy makers (government agencies) etc..

**Provide Aditional information if possible**
A preliminary description of the new model is described in the following document: 

[District_Cooling_Optimisation.pdf](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/9761960/District_Cooling_Optimisation.pdf)

PR #3260, opened 2022-10-27
Update of the centralised optimisation algorithm

MatNif:
** this is ongoing work, the PR is launched to make the review easier, I will update this message when the branch is ready for a final review before the merge **

The new optimisation problem this update is based on is described in issue #3254. 
The code structure is described in the class diagram [here](https://lucid.app/lucidchart/102d99bd-7e82-4966-81df-cb628bcc8c5b/edit?viewport_loc=1413%2C-98%2C1472%2C740%2C0_0&invitationId=inv_4074555f-a935-46e5-a833-6f66f8b3d472), which will be updated regularly to reflect the latest changes of the code.


MatNif:
@shanshanhsieh , @lguilhermers whenever you find the time to review yesterday's commit ('introducing networks') could you check if the following code lines make sense to you? These parts were all taken from the old thermal networks script.

line 124 `optimization_new/network.py` (line 351 `technologies/thermal_network/simplified_thermal_network.py`):
Why does the linear pressure loss conversion uses a factor of 1000?

line 189 `optimization_new/network.py` (line 466 `technologies/thermal_network/simplified_thermal_network.py`):
Why are the thermal losses of supply pipes counted twice instead of counting losses of return pipes?

line 470 `optimization_new/network.py` (line 320 `technologies/thermal_network/simplified_thermal_network.py`):
Why is the conversion factor `FT_TO_M` used twice and why is a factor of 1000 used for the conversion here again?

line 475-477 `optimization_new/network.py` (line 335 `technologies/thermal_network/simplified_thermal_network.py`):
Why is the thermal unit design head multiplied by a factor of 1.2? How should it be accounted for, depending on the number of substations\buildings in the network?

cea/technologies/constants.py
```diff
@@ -79,3 +79,8 @@
 
 #natural gas conncetion
 GAS_CONNECTION_COST = 15.5 / 1000  # CHF / W, from  Energie360 15.5 CHF / kW
+
+# Thermal Network
+TYPE_MAT_DEFAULT = 'T1'
```
lguilhermers, line 84: Unclear to me what this variable means, is it material type? Perhaps add a comment to make more clear. Pipe diameter is self explanatory, but I would add the units is being considered, just for documentations

cea/technologies/network_layout/main.py
```diff
@@ -82,16 +82,19 @@ class NetworkLayout(object):
 
     def __init__(self, network_layout=None):
         self.network_type = "DC"
-        self.connected_buildings = []
+        # self.connected_buildings = []
+        self.connected_buildings = network_layout.connected_buildings
         self.disconnected_buildings = []
         self.pipe_diameter = 150
```
lguilhermers, line 88: should this be called from the constants instead?

cea/optimization_new/energyPotential.py
```diff
@@ -0,0 +1,193 @@
...
+    def _get_building_potentials(self, energy_potential_files, building_codes, main_potential_column_name,
+                                 temperature_column_name=None, auxiliary_potential_column_name=None):
+        """
+        Gets main and auxiliary potentials from the stored energy potential files and stores them in the corresponding
+        object attributes. In case a temperature column name is indicated, the average temperature (when operating
+        the corresponding component at maximum potential) is returned.
+        """
+        # check if there are potential files for any of the buildings
+        if not any([exists(file) for file in energy_potential_files]):
+            print(f"No {self.type} potentials could be found for the indicated buildings.")
```
lguilhermers, line 145: Maybe add. "Consider running potentials scripts" or something like that (CEA has some warnings like that which are useful when another script can solve your problem.)

cea/optimization_new/energyPotential.py
```diff
@@ -0,0 +1,193 @@
...
+            average_temp = np.nan
+        return average_temp
+
+    @staticmethod
+    def _temp_to_thermal_ec(temperature, thermal_energy_carriers):
+        if not np.isnan(temperature):
+            index_closest_mean_temp = (thermal_energy_carriers['mean_qual'] - temperature).abs().nsmallest(n=1).index[0]
+            energy_carrier_code = thermal_energy_carriers['code'].iloc[index_closest_mean_temp]
+        else:
+            energy_carrier_code = "T???"
```
lguilhermers, line 192: placeholder?

lguilhermers:
In ENERGY_CARRIERS.xlsx, does it make sense to dfine a default for all thermal and radiation costs and ghg to 0.0? Or you expact to find some values for it? (In the case of leaving it blank bringing errors later on)

lguilhermers:
> Why is the conversion factor `FT_TO_M` used twice and why is a factor of 1000 used for the conversion here again?
both FT_TO_M seem reduntant to me, as it is a constant factor. 
`FT/kFT = FT/1000 *FT = M/ 1000 M` (this number is adimentional so it doesn't matter)
I guess it was done literally for clarity, but can be removed without compromising the calculations

cea/technologies/constants.py
MatNif: Yes. this is meant to declare the type of pipes (material and construction) that is used for the network. This is based on the existing thermal network code. I added a comment in the constants-file.
However,  I just realised, that the pipe type is currently a bit meaningless. The only method this could play into is the `calc_linear_thermal_loss_coefficient`-function from `simplified_thermal_network.py`, but the function assumes a constant construction (steel pipe + calcium silicate insulation). So, there is currently no database in CEA specifying the properties of these types, but that could be added in the future.

cea/technologies/network_layout/main.py
MatNif: This is not really part of the new optimisation script. But this would indeed be cleaner and doesn't change the functionality of the code, so I changed it.

cea/optimization_new/energyPotential.py
MatNif: Good point. Done.

cea/optimization_new/energyPotential.py
MatNif: Yes, that is a placeholder for now. I'm currently linking that up. Should be part of the next bigger commit at the end of this week.

cea/technologies/constants.py
lguilhermers: Got it! Could it be relevant for the LCA parts, perhaps?

lguilhermers:
> Why is the thermal unit design head multiplied by a factor of 1.2? How should it be accounted for, depending on the number of substations\buildings in the network?

This seems to me a safety factor, as you pointed. I think it should be declared before and individually, for clarity. Not sure how it would scale, but my guess is that it is for each substation.

lguilhermers:
> Why are the thermal losses of supply pipes counted twice instead of counting losses of return pipes?

I agree, even the comment says it is to account for return, which is calculated right before, so it seems logical to change that as well.

MatNif:
> In ENERGY_CARRIERS.xlsx, does it make sense to dfine a default for all thermal and radiation costs and ghg to 0.0? Or you expect to find some values for it? (In the case of leaving it blank bringing errors later on)

I think if the cells of a .csv-file are empty, they get read as 0.0 by default. But I'll check that either today or tomorrow.
My initial thinking was that it makes more sense to leave the cells empty because e.g. there is no real way to define what the price of UV-light is to an energy system? 
But I guess putting zeros or dashes in those spots might make it more intuitive for a potential user that might want to change the data base themselves, so I changed the values now.

MatNif:
@reyery , @shizhongming 

Would you be able to give the multiprocessing a go on Mac. If that works and you can't see any other immediate issues, I think we can merge. I'm sure, as people use it, more errors/unexpected behaviors will become apparent, but I think it would be easier to handle fixes to those through normal issues from here on out.

shizhongming (approved):

reyery:
@MatNif should the "new" database files replace the old ones?

MatNif:
Hi @reyery,
The short answer is no. 
We have not yet removed the old optimisation and that still gets all of its information from the old databases. Once we gain a bit more confidence in the new script, I'll procede to remove the old script along with the old databases.

reyery (approved):
I was able to successful run it on my Windows machine

Issue #3120, opened 2022-03-29
Apple silicon machines require Rosetta to be installed to run CEA


reyery:
Currently the python environment for CEA on Apple silicon machines is based on x86 due to some dependencies not found for osx-arm.

Python modules that prevent CEA from running natively on Apple silicon machines (according to conda):
```
  - flask-restplus
  - wntr 0.2.2.1**
  - numpy 1.19.1**
  - pythonocc-core
```

Also found out that daysim binaries are running under x86 even though they are compiled as Universal binaries. Probably due to python running bash as x86.

reyery:
Update: `pythonocc-core` is now available for `osx-arm64` in conda-forge

https://github.com/tpaviot/pythonocc-core/issues/1071

jimenofonseca:
Hey rey...and what happened to Wntr? I see that the thermal network does not run on Master for M1-2 CHIPS.

reyery:
@jimenofonseca You might need to update the conda/python environment since there were changes. You could try installing it for Mac again from the docs using the new instructions [here](https://city-energy-analyst.readthedocs.io/en/latest/installation/installation-on-macos.html).

jimenofonseca:
Yes. I did a fresh install in a new mac. There i found the problem. How is it in your end? Does it work the network simulation on Master in a mac?JimenoOn 10 Mar 2024, at 01:32, Reynold Mok ***@***.***> wrote:﻿
@jimenofonseca You might need to update the conda/python environment since there were changes. You could try installing it for Mac again from the docs using the new instructions here.

reyery:
@jimenofonseca I have tested on two different apple silicon macs and I was able to run every script. What was the actual error that you are getting? I can try to reproduce it.

jimenofonseca:
Hoi rey i got the problem, that the class _elevated in wntr does not exist.Do you see this also?JimenoOn 18 Mar 2024, at 08:12, Reynold Mok ***@***.***> wrote:﻿
Closed #3120 as completed.

reyery:
@jimenofonseca that issue was resolved in https://github.com/architecture-building-systems/CityEnergyAnalyst/pull/3265/commits/c60e7dbaa63d8d7390a3fe4804fd8dd0149e2ba8

Could you check what version of `wntr` is in your environment?

PR #3265, opened 2022-11-15
Update pythonocc to enable Mac ARM support

reyery:
Could potential fix https://github.com/architecture-building-systems/CityEnergyAnalyst/issues/3120


shizhongming:
We paused merging this PR for now until @reyery figure it out what to do with it.

reyery:
Not sure if this can be merged since the results for the thermal networks is very different due to changes to `wntr`

@jimenofonseca maybe you can take a look at the differences and see if it makes sense. I attached the outputs in the zip file
[Archive.7z.zip](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/11351453/Archive.7z.zip)

reyery:
@MatNif maybe you have some ideas as well. Apparently one of the changes in `wntr` is that it is using EPANET version 2.00.12 vs 2.2.0, not sure if that matters.

https://wntr.readthedocs.io/en/latest/whatsnew.html

MatNif:
When I try to start up the dashboard from a Windows PC on this branch I am running into a bit of an issue already. It looks like the libpysal.io library is not installed by default. I'm not sure if the warnings-handling in utilities/dbf.py that was there before the changes made in this branch was meant to handle these errors.

Screenshot 2023-06-22 112652

MatNif:
Then, as I run the code of the network-layout script, I run into the following error:

Screenshot 2023-06-22 123510

which might have something to do with the fact that the 'nearest'-method is not available for all versions of the geopandas library:

Screenshot 2023-06-22 123532

reyery:
I have tried recompiling `wntr` for the version that we are using (v0.2.3), to avoid the difference in results for the latest version. Everything seems to be working now, and there should not be differences when running it anymore.

@MatNif the EPANET version used in `wntr` would still be v2.00.12 vs v2.2.0, not sure what the differences are but I guess we can investigate this next time. So no further actions are required from your end.

@shizhongming It now works for Apple Silicon Macs natively, it _should_ be a little faster now as well, since there is no need for a translation layer anymore. I have updated a lot of the python dependencies as well, so testing it would be much more complicated, as you would need to create a new conda environment from the new environment file.

First run a workflow using the current environment that you have, in the `master` branch.

Then checkout to this PR branch, run the following commands in the CEA git repo folder:
```shell
mamba env create -n cea-arm -f environment.yml
conda activate cea-arm
pip install -e .
```
Run the same workflow using this new environment. Compare if the results are the same (especially for thermal network results)

Let me know if you encounter any issues. Thanks!

shizhongming:
@reyery Thanks for the PR again.
I have just tested it as instructed using two scenarios: one in Zurich and one in Singapore.

For `thermal-network`, for both scenarios using `master` and this `PR`, the results were identical.
For `solar-potentials`, in the Singapore scenario, the results were identical, while in the Zurich scenario, the results are mostly slightly different. I have attached the two .csv here.

Zurich-PR:
[PV_total_buildings.csv](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/12577169/PV_total_buildings.csv)

Zurich-Master:
[PV_total_buildings.csv](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/12577166/PV_total_buildings.csv)

reyery:
Interestingly there are differences in the area of PV, but I don't think it is that significant. Would it be possible to try a larger scenario and see if affects more buildings.

shizhongming (approved):

PR #3268, opened 2022-11-23
GRID_MWhyr in total demands can't handle cases where components provide 'nan' values

lguilhermers:


lguilhermers:
I observed in some cases, some of the outputs in Total_domand (GRID_cs and GRID_aux, in my cases), get filled with nan values. This causes errors whenever the total demand (GRID) is computed. Therefore, as a quickfix, I put to convert any nan to 0 before making those sums, which seem to have solved the issue.

It is still pending to understand what causes some of those specific vectors to become nan in the first place. I tried to dig into that a bit, but it got complicated very fast.

To test:
1 - Create a scenario which has buildings that generate those one of the GRID_component outputs to result in nan values. This can be reached by making building with very high setpoints which will never require cooling (singapore context). In this case, GRID_cs should become nan. If there is difficulties to reproduce this step, I can provide some scenarios.
2 - Check variable GRID_MWyr in Total_demand.csv. It should be the sum of all following GRID processes (columns I to T) instead of 0.

reyery:
@lguilhermers could you provide me a sample scenario for me to test this?

lguilhermers:
Hi Rey, here are the scenarios I got errors for some of the buildings.
You can (re)run demand for them and check Total_demand.csv
Look for GRID_MWhyr which is 0 or very low for some buildings. Depending on the case, GRID_cs and GRID_aux will be empty or have nan.

Scenarios here (expires in 7 days):
https://www.transfernow.net/dl/20221125xmrgDIDo

MatNif (commented):
Changes make sense to me. The same result could probably be achieved using a slightly more concise syntax (e.g. using `numpy.nansum([E_xx, E_xy, E_yy...])`), but the changes achieve what they are intended to achieve.

MatNif:
The only reason I can think of for not addressing these `nan`-values in the past (assuming it was found in the past), would be that they are meant hint at an error in the input-files. 

I tested the Lakeside case and found that the two buildings that do not produce any result (or near 0), B1006 and B1018, happen to have the same cooling system set point of 25°C. That is the lowest set point among all buildings.
While I don't think it should play into the electricity demand of the buildings, it is the only common 'odd'-value that I could find in the buildings inputs.

For the lakeside scenario, the changes made in this branch, seem to result in electricity demands for B1006 and B1018 that are comparable to the rest of the buildings in the domain. (_small comment on the side: I know this has nothing to do with the changes of this branch, but all demands in this scenario are extremely low, even for buildings that do not use electrically powered AC-systems_)

So as a conclusion, I can't find out why these `nan`-values would not have been addressed if the issue had been found in the past. So, I think the changes here can be merged. However, we should also open up an issue for further investigation of these `nan`-values.

reyery (approved):

PR #3278, opened 2023-01-04
Browser GUI

reyery:
This PR helps serve the frontend GUI from Flask to the browser.

To test, run `cea dashboard --browser true` in the cea console.


lguilhermers:
Hi @reyery nice feature, I will add comments here as I notice them.
1 - I can't cancel a simulation that is running, as it was possible on the dashboard.

lguilhermers:
The box with error messages looks bugged:
image

reyery:
> Hi @reyery nice feature, I will add comments here as I notice them. Current Jobs tab: 1 - I can't cancel a simulation that is running, as it was possible on the dashboard (or maybe it just shows as pending) 2 - (minor) The simulation box (aka Current Jobs) does not pop up once clicking Run Script (so to confirm the simulation is running you have to manually click the button on the lower right corner, which may not even be noticed by new users) 3- Once a simulation is finished it still shows as pending.

I just realized that all these issues are caused by the new `socketio` dependency (the one responsible for job notifications) that was added. Let me double check. 

`osmnx` was updated in this branch as well.

reyery:
@lguilhermers I have fixed the error box overflow. Other than the dependency issues (`osmnx` and `socketio`), did you notice any issues? If not I think it should be safe to merge it.

reyery:
I shall merge this pull request first since, the new command `cea dashboard --browser true` works. The other issues mentioned here are due to differences in python dependencies, which could be solved in another PR.

PR #3299, opened 2023-04-27
fix issue with werkzeug

jimenofonseca:


reyery:
@jimenofonseca the issue has actually already been fixed in master

reyery:
Will close this PR since the `werkzeug` issue has already been fixed in the latest version of `flask-socketio` that is already being installed by CEA.

PR #3300, opened 2023-04-27
Lifecycle costs and emissions.

jimenofonseca:
Drastic improvements to costs and lifecycle calculation. It adds also a new "Column" tab in the Envelope.xlsx files and a new database for Latin American buildings. 

@reyery it will be cool to integrate this in the next release?

@shizhongming can you test on your side that this works on your end?


reyery:
@jimenofonseca there seems to be some errors in `archetype-mapper`

shizhongming:
@jimenofonseca 
I just tested this PR. For me, the `system_costs` script did not run through. 
For example, "['IR_%'] not in index" is not included in the Electricity Tab in the Database. Also, I encountered KeyError: 'PV_export_MWhyr'.
 
Meanwhile, to add new database, could you include the reference of your data? Thanks.

jimenofonseca:
@shizhongming please try again. the unittests both locally and on the server pass. I added references in line wherever possible to the databases. where not than "educated guess". LATAM is in principle CH a bit improved to some data we got from case studies in Mexico and Colombia

jimenofonseca:
so reference: Diverse Case studies: Donation of various Anonymus Users for study cases in Mexico, Colombia, and Chile. 2020

jimenofonseca:
@shizhongming @reyery , we take it down until further notice.

PR #3302, opened 2023-04-30
Fixes to output PV

jimenofonseca:
There was a problem in the output that made difficult to post good graphs in Solart Tehcnologies. The date was wrong.
@shizhongming easy fix for the new version?

### DIFF
--- a/cea/technologies/solar/photovoltaic_thermal.py
+++ b/cea/technologies/solar/photovoltaic_thermal.py
@@ -653,7 +653,11 @@ def calc_Cinv_PVT(PVT_peak_W, locator, technology=0):
             PVT_peak_W = PVT_cost_data['cap_min'].values[0]
         PVT_cost_data = PVT_cost_data[
             (PVT_cost_data['cap_min'] <= PVT_peak_W) & (PVT_cost_data['cap_max'] > PVT_peak_W)]
-        Inv_a = PVT_cost_data.iloc[0]['a']
+        try:
+            Inv_a = PVT_cost_data.iloc[0]['a']
+        except Exception as error:
+            print('Caught this error: ' + repr(error))
+            print(PVT_cost_data, PVT_peak_W)
         Inv_b = PVT_cost_data.iloc[0]['b']
         Inv_c = PVT_cost_data.iloc[0]['c']
         Inv_d = PVT_cost_data.iloc[0]['d']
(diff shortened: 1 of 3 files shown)


shizhongming:
@jimenofonseca 
The PR runs through and all the graphs work well, with two minor exceptions.
1. PVT_total.csv has a column of unnecessary index (i.e. 0,...,8759);
2. PV_total.csv's ['date'] column is not as index.

Meanwhile, I do not really understand you meant by good graphs. In the test case I had in Singapore, the output dates look fine and graphs generated by Solar Technology all looked fine. Can you elaborate? 
or @reyery can you also take a look?

jimenofonseca:
1. fixed
2. does it work for you?

yeah, in my version I could not plot. the Date was repeated several time. you can see the error in the original files.

shizhongming:
@jimenofonseca 
1. working fine now
2. Please refer to the image below for what I meant. The ['Date'] column is supposed to be column A, is it?
Screenshot 2023-05-01 at 10 51 42 PM

cea/technologies/solar/photovoltaic_thermal.py
```diff
@@ -653,7 +653,11 @@ def calc_Cinv_PVT(PVT_peak_W, locator, technology=0):
             PVT_peak_W = PVT_cost_data['cap_min'].values[0]
         PVT_cost_data = PVT_cost_data[
             (PVT_cost_data['cap_min'] <= PVT_peak_W) & (PVT_cost_data['cap_max'] > PVT_peak_W)]
-        Inv_a = PVT_cost_data.iloc[0]['a']
+        try:
+            Inv_a = PVT_cost_data.iloc[0]['a']
+        except Exception as error:
+            print('Caught this error: ' + repr(error))
+            print(PVT_cost_data, PVT_peak_W)
```
reyery, line 660: I think we should raise an Exception when `PVT_cost_data` is empty, instead of ignoring the error, since `Inv_a` will not be defined if we catch the exception like this.

reyery:
@jimenofonseca what was the main issue with the PV potential plots? I tried running it on the main branch and everything seems normal to me and I did not find any repeated dates in the files.

cea/technologies/solar/photovoltaic_thermal.py
jimenofonseca: fixed

jimenofonseca:
@reyery did you run photovoltaic thermal on master? there is where you get this output. PVT_Total.csv

Then to me the plots do not work well and in any case the files is just wrong


[PVT_total.csv](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/11412244/PVT_total.csv)

jimenofonseca:
@shizhongming fixed that too

reyery:
@jimenofonseca #3305 I have created a better way of fixing this at the root cause of the problem. Let me know if it works for you.

shizhongming:
We close this PR (thanks to @jimenofonseca for the effort) as PR#3305 (by @reyery) is solving this issue and reducing computing usage.

PR #3305, opened 2023-05-06
Fix aggregation for solar potential scripts

reyery:
During aggregation of the solar potential data, the csv is being read by pandas without parsing the `Date` column as `datetime` type (default behavior), but reading it as it is causing the `Date` column values to be parsed as string.

When "summing" 2 dataframes together, pandas will "sum"/concatenate the `Date` column as strings, resulting in a long string of dates.
https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/1dc92c5fea5a03085032fc2483b17d686fbf4b22/cea/technologies/solar/photovoltaic_thermal.py#L725
https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/1dc92c5fea5a03085032fc2483b17d686fbf4b22/cea/technologies/solar/solar_collector.py#L1017

```
# example of concatenated Date column value
2009-01-01 01:00:00+01:002009-01-01 01:00:00+01:002009-01-01 01:00:00+01:00
```

Setting date as index when reading csv will prevent this as pandas will not "sum" indexes.

### DIFF
--- a/cea/technologies/solar/photovoltaic_thermal.py
+++ b/cea/technologies/solar/photovoltaic_thermal.py
@@ -714,7 +714,7 @@ def main(config):
     # aggregate results from all buildings
     aggregated_annual_results = {}
     for i, building in enumerate(building_names):
-        hourly_results_per_building = pd.read_csv(locator.PVT_results(building))
+        hourly_results_per_building = pd.read_csv(locator.PVT_results(building)).set_index("Date")
         if i == 0:
             aggregated_hourly_results_df = hourly_results_per_building
             temperature_sup = []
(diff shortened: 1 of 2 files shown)


shizhongming:
@reyery Does it mean if we merge this PR, we will no longer need PR#3202?

jimenofonseca:
Cool lets do it. And take down the othee PR.On 9 May 2023, at 12:57, Reynold Mok ***@***.***> wrote:﻿
Yeah, because this solves the root of the problem instead of fixing the outcome. The other PR could still take up more memory when aggregating the results because the date column is still being concatenated.

shizhongming (approved):

PR #3306, opened 2023-05-07
Read and write sensor data using feather instead of json

### LINKED ISSUE #3304: Change sensor data format to save on disk space
Sensor data from radiation script is now being saved in JSON format. When running the script on a larger scale, these files add up and might consume a lot of disk space. This could be solved by saving the files in another format that supports compression. Since users rarely open sensor data file manually, I think it is a good compromise to save it in another format while removing the readability of the file (i.e. opening it in a text editor). 

I just did a simple size comparison for 1012 buildings, using different formats for the sensor data with compression (hdf5 and feather). 
We could achieve about 4-6x reduction in size with other formats:
```
feather    6.379GB
h5         4.333GB
json      24.212GB
```

I would suggest using feather since it is consumes much lesser CPU resources (while reading and writing) compared to hdf5 even though the size is not as small.

reyery:
Closes #3304

You would have to install `pyarrow` using `pip install pyarrow` in the CEA console for this to work.

As mentioned in the issue, using feather instead of JSON actually saves both time and disk space. This would greatly help scenarios with large amount of buildings.

> It's actually faster to read/write feather compared to json. The time taken for feather is quite consistent and is sub 0s even for large buildings.
> 
> For a single building with 2917 sensors
> 
> ```
> write json       20.513s
> read json         4.054s
> 
> write feather     0.582s
> read feather      0.182s
> 
> size json        305.4MB
> size feather      40.3MB
> ```
> 
> For a single building with 217 sensors
> 
> ```
> write json       1.4673s
> read json         0.265s
> 
> write feather     0.507s
> read feather      0.126s
> 
> size json         22.7MB
> size feather       6.6MB
> ```

### DIFF
--- a/cea/inputlocator.py
+++ b/cea/inputlocator.py
@@ -887,12 +887,12 @@ class InputLocator(object):
         return self._ensure_folder(self.scenario, 'outputs', 'data', 'solar-radiation')
 
     def get_radiation_building(self, building):
-        """scenario/outputs/data/solar-radiation/${building}_insolation.json"""
+        """scenario/outputs/data/solar-radiation/${building}_radiation.csv"""
         return os.path.join(self.get_solar_radiation_folder(), '%s_radiation.csv' % building)
 
     def get_radiation_building_sensors(self, building):
-        """scenario/outputs/data/solar-radiation/${building}_insolation_Whm2.json"""
-        return os.path.join(self.get_solar_radiation_folder(), '%s_insolation_Whm2.json' % building)
+        """scenario/outputs/data/solar-radiation/${building}_insolation_Whm2.feather"""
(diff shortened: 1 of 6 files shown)


shizhongming:
@reyery I need to take a look at this feather file first, because: 1. Some CEA plug-ins are using the older json format; 2. I some visualisation idea for the sensors are using the older json format.

shizhongming:
@reyery I was able to run this PR smootly.
I tested it on a cluster of ten small buildings (around 120 sensors per building).

The total execution time was 98.79s v. 93.05s (using json v. using feather).
The total radiation file size was 134.3mb v 65.9mb (using json v. using feather).

The improvement was not as substantial as the test case in your comment.

Well, let's discuss tomorrow at the CEA Team Meeting. : )

shizhongming (approved):

Issue #3308, opened 2023-05-10
zone_helper creates shapefile with incorrect building heights


martin-mosteiro:
**Describe the bug**
I created a case study and for most buildings I got the number of floors as the building height
![Uploading Screen Shot 2023-05-10 at 8.57.56 PM.png…]()
This is not the case when I check the case study on OSM, where most buildings have a number of floors but not a building height

martin-mosteiro:
I just had this issue happen again on a fresh installation of CEA on an old computer. It's weird that this only seems to be happening on (a) old Macs, (b) old Macs that _I'm_ using 😂

I switched to branch `3308-zone-helper` again and it solved the issue for me – Maybe it's worth testing whether that branch breaks anything for other users, if it doesn't it break anything for mainstream users and solves an issue for fringe users (e.g., me... 😂) maybe it's worth considering merging it.

shizhongming:
@martin-mosteiro will take a look at it.

ShiZhongming:
Will check with @martin-mosteiro again on Friday 15.11.2024.

PR #3309, opened 2023-05-10
Remove forced incorrect building heights

martin-mosteiro:
I created a new case study and found some odd behavior. After checking the code I realized (a) there was a repeated block of code that served no purpose, (b) building heights of None value were filled in with 1 for no good reason.

Removing those two parts of the code solved issue #3308


shizhongming:
Hi @martin-mosteiro 
Can you advise a case study file that we can use to test this? : )

martin-mosteiro:
The problem is precisely in generating a case study, so providing a case study file won't help. You can test it by creating any case study of your preference, for example here's the city center of Zürich:
1. Master branch
Screen Shot 2023-05-23 at 5 50 47 PM
2. This PR
Screen Shot 2023-05-23 at 6 17 50 PM

Here's the [site shapefile](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/11542542/site.zip) for your reference.

martin-mosteiro:
Interesting behavior. After checking with @shizhongming this happens on my computer but not his. I'll check the issue again and come back to it by next week.

shizhongming:
Close for now - Martin will follow up

PR #3324, opened 2023-06-15
Fixed deap-creator error

MatNif:
When running in multiprocessing the the 'Individual'-class created using the deap-creator module could not be accessed since it was not set as a global value. This fix omits creating unnecessary global values by introducing an new 'explicit' Individual-class.


shizhongming:
Thanks for this @MatNif
I use tested a DC and a DH cases on CEA Dashboard for MacOS. Both worked fine now.

shizhongming (approved):

reyery (approved):

Issue #3289, opened 2023-02-17
Dependency error while creating a scenario: zone_helper: module 'osmnx' has no attribute 'geometries'


Arunim21:
**Describe the bug**

**To Reproduce**

**Expected behaviour**

**Screenshots**

**Hardware (please complete the following information):**
 - OS and version: [e.g. iOS 10.7 or Windows 10]
 - Version of CEA you are using [e.g. v. 2.31.1]

chrisbay914:
We are experience the same problem, running the latest version of CEA on both latest OS on Mac and W10.

1. create scenario
2. we tried diffrent locations in Stockholm, Sweden to extract building information

lguilhermers:
@Arunim21 , @chrisbay914 
As a workaround, can you please try the following steps:
1. Open CEA Console (should be together with your CEA Dashboard and cea.config)
2. run `python -m pip uninstall osmnx`
3. run `python -m pip install osmnx`
4. Re-open the dashboard and try to create your scenarios.

chrisbay914:
Thank you, now it works. However, the area in CEA for the automated building generation process is off from the official records. Is there a way to manually change the footprint area?

lguilhermers:
Yes, you can open the generated building zone.shp or surroundings in any GIS editor and edit the polygon corners, for example.
Alternatively, you can add new buildings or even do everything from scratch by following this blog post:
https://www.cityenergyanalyst.com/blog-posts/2021/01/20/2021-1-20-how-to-add-own-building-geometries-to-a-scenario-import-of-manually-created-shapefiles

Just make sure your new files follow the CEA structure for building names and attributes.

Juveria-shah:
This did not fix my issue.
The updated version of osmnx 1.4.0 was creating a problem, we tried to fix this by relying on version 1.2.2(osmnx==1.2.2)

jimenofonseca:
yea lets reopen this

PR #3328, opened 2023-06-17
Update environment.yml

Juveria-shah:
This solves issue #3289


jimenofonseca:
THis works for me, I have tested in macos and windows. thanks fo rthe cool solution.

PR #3331, opened 2023-06-19
Fix EV input validation in editor

### LINKED ISSUE #3327: Bug in EV schedule
Under the Input Editor/schedules/electromobility the input accepts only binary values and can not be modified for integer values.
Path to reproduce error:
Input Editor -> Schedules->Electromobility-> Select building-> input EV number -> ERROR

image

reyery:
Closes #3327

Run to test
```
cea dashboard --browser true
```

Non-browser version (GUI application) will not be updated

### DIFF
--- a/cea/interfaces/dashboard/frontend/__init__.py
+++ b/cea/interfaces/dashboard/frontend/__init__.py
@@ -14,11 +14,11 @@ def frontend():
 
 
 def get_build():
-    output = os.path.dirname(os.path.abspath(__file__))
 
     if os.path.exists(os.path.join(output, "build", "index.html")):
         return
     url = "https://github.com/architecture-building-systems/CityEnergyAnalyst-GUI/releases/download/v3.32.1-browser/build.zip"
+    output = os.path.join(os.path.dirname(os.path.abspath(__file__)), "build")
 
     print("Fetching GUI build...")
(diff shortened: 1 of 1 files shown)


jimenofonseca:
Is the Idea that from now on we will not support the normal version? if so how do we Install and Launch CEA from now on? und starting as of when is this ready?
I missed the discussion about this important change :)

jimenofonseca:
Fun!So i guess the webversion has full funktionality now?On the other hand:Do I define the schemas for that table in the same way as the others?On 19 Jun 2023, at 17:18, Reynold Mok ***@***.***> wrote:﻿
Significant changes were made to enable the browser version and most fixes have been applied to it. Applying the same fixes to the non-browser version would require quite a bit of work. And I currently do not have the resources to support two versions of the frontend.
As long as you have the dev version, you should be able to run the browser version without any problems or installing anything else.

reyery:
@jimenofonseca I have included this fix into the GUI application. A new release will be up soon that includes this and can be installed.

jimenofonseca:
Very cool. Please let me know!On 29 Jun 2023, at 14:12, Reynold Mok ***@***.***> wrote:﻿
@jimenofonseca I have included this fix into the GUI application. A new release will be up soon that includes this and can be installed.

PR #3340, opened 2023-07-06
Change selection process to prevent duplicates

MatNif:
With this change, offspring created from mutation and crossover operations are compared against the existing population to determine which ones are novel and prevent duplicates from being added to the population.

This should also make the optimisation script run faster since duplicate systems are not unnecessarily evaluated.


shizhongming:
Hi @MatNif 
Thanks for the update!
How should we test this? Using the same scenarios for the PR test before? :D

MatNif:
Yes. The scenario you've used before should work. Ideally, rename the optimisation_new folder in the results so you can compare the debugging files of both runs.

shizhongming (approved):

Issue #3351, opened 2023-07-21
Embodied emission update 


MMeshkinkiya:
This is a request about essential updates in building Life Cycle Assessment (LCA) calculations in City Energy Analyst (embodied.py).

Whole building Life Cycle Assessment is typically calculated as:

_LCA=LCAproduction + LCAreplacment + LCAend of life + LCAoperation_

However, the LCAreplacement is currently missing in CEA. Therefore, following updates is necessary:
    •	 The LCA of replacement which considers the specific Reference service life for 
        each component.
    •	 Updating the Envelop dataset and adding new inputs as 
        Service_Life_Component (Reference service life).

The additional eq. to consider LCA of replacement is: 

_ni=(((SERVICE_LIFE_OF_BUILDINGS/Service_Life_Component)-1)+1)._

ni should be calculated for each component and multiplied to the GHG_component and the component's area.

Additional comments for source code:
     •	  In the Envelop dataset, GHG_component_KgCO2m2 refers to the sum of GWP layers of each component, which returns the total GWP as: 
  _Total GWP = GWP production + GWP end of life_(to be added)
     •	   To calculate the embodied emission of retrofitted building, only the additional 
           emissions should be considered in the df[total_column]. Which means only 
           the retrofitted components should be involved in the equation and the 
           unchanged components should be set to 0. (to be added)
     •	   Note: The new input variable (Service_Life_Component) is already added for CH dataset and are extracted from: https://doi.org/10.1016/j.dib.2021.107062. (added in #3354)

martin-mosteiro:
@MMeshkinkiya Indeed, this is missing in CEA at the moment. The changes you proposed sound solid and would make a great first contribution to CEA if you're keen!

shizhongming:
Hi @martin-mosteiro 
@MMeshkinkiya is currently helping CEA improve the LCA Component. Soon, she will create a pull request accordingly.
I will add you the reviewer of that PR :D

PR #3354, opened 2023-07-25
Embodied update

MMeshkinkiya:
Following the suggestions in issue #3351, in this branch, a modified version of emissions model algorithm under the CityEnergyAnalyst \ cea \ analysis \ lca \ **embodied.py** is presented.

A new variable as _Service_Life_Component_ should add to building components database in cea, and to WINDOW , ROOF, WALL, FLOOR sheets in ENVELOPE.xlsx 

The changes are happened in the lines: _**116-151 and 245-259**._


martin-mosteiro:
Copying the description in the current version of the documentation, which is more relevant to the PR than to the script itself:
```
    # New changes are essential i.e. (1) The LCA of replacement which consider the specific Reference service life
    # for each component.
    # (2) Updating Envelop dataset to contain Service_Life_Component (Reference service life).
    # Service_Life_Component are extracted from: https://doi.org/10.1016/j.dib.2021.107062 .
    # The eq. for LCA of replacement is: ni=(((SERVICE_LIFE_OF_BUILDINGS/Service_Life_Component)-1)+1).
    # ni should be multiplied by GHG_components and component's area.
    # For Envelop dataset, GHG_component_KgCO2m2 refers to the sum of GWP layers of each component.
    # GHG_components_... returns the total GWP which is: Total GWP = GWP production + GWP end of life.
    # df[total_column] is LCA of whole building and can be calculated for both existing and new buildings.
    # To calculate the embodied emission of retrofitted building only the additional emissions
    # should be considered. To do that, only the retrofitted components should be involved in equations
    # and others components should be assumed as 0.
    # To avoid a zero result for df[total_column] in buildings over 60 years, SERVICE_LIFE_OF_BUILDINGS must
    # change to 120 years in workflows/constants.py
```

cea/analysis/lca/embodied.py
```diff
@@ -203,30 +220,56 @@ def calculate_contributions(df, year_to_calculate):
...
     ## if it was built less than 60 years before, the contribution from each building component is calculated
-    df[total_column] = ((df['GHG_WALL_kgCO2m2'] * (df['area_walls_ext_ag'] + df['area_walls_ext_bg']) +
-                         df['GHG_WIN_kgCO2m2'] * df['windows_ag'] +
-                         df['GHG_FLOOR_kgCO2m2'] * df['floor_area_ag'] +
-                         df['GHG_BASE_kgCO2m2'] * df['floor_area_bg'] +
-                         df['GHG_PART_kgCO2m2'] * (df['floor_area_ag']+df['floor_area_bg']) * CONVERSION_AREA_TO_FLOOR_AREA_RATIO +
-                         df['GHG_ROOF_kgCO2m2'] * df['footprint']) / SERVICE_LIFE_OF_BUILDINGS) * df['confirm']
 
-    df[total_column] += (((df['floor_area_ag'] + df['floor_area_bg']) * EMISSIONS_EMBODIED_TECHNICAL_SYSTEMS) / SERVICE_LIFE_OF_TECHNICAL_SYSTEMS) * df['confirm']
+    # New changes are essential i.e. (1) The LCA of replacement which consider the specific Reference service life
```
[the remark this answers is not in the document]
martin-mosteiro: Also: `ni=(((SERVICE_LIFE_OF_BUILDINGS/Service_Life_Component)-1)+1)`
The -1 and the +1 just cancel out, leaving: `ni=SERVICE_LIFE_OF_BUILDINGS/Service_Life_Component`
Is that the intended result?

cea/constants.py
```diff
@@ -29,7 +29,7 @@
 
 # grey emissions
 CONVERSION_AREA_TO_FLOOR_AREA_RATIO = 1.5  # conversion component's area to floor area
-SERVICE_LIFE_OF_BUILDINGS = 60  # service life of standard building components and materials SIA2032
+SERVICE_LIFE_OF_BUILDINGS = 120  # service life of standard building components and materials SIA2032
```
martin-mosteiro, line 32: Has this value been changed in a newer version of SIA 2032? If so, could you please add the year of the new version of the standard? And if not, could you add the new source?

Note:
```
To avoid a zero result for df[total_column] in buildings over 60 years, 
SERVICE_LIFE_OF_BUILDINGS must change to 120 years in workflows/constants.py
```

This is not a good reason to change a value in the constants. If SIA2032 recommends using 60 years, then our equations should match that assumption rather than changing the constant to match our equation. If this assumption is outdated (e.g., if there is a new version of SIA2032 recommends a value of 120 years), then the reference should be updated too.

cea/databases/CH/components/CONVERSION_NEW.xlsx
martin-mosteiro, line 1: This file needs to be renamed to **CONVERSION.xlsx**

cea/examples/embodiedUpdated.py
martin-mosteiro, line 1: What is this script used for?

cea/databases/CH/assemblies/ENVELOPE.xlsx
martin-mosteiro, line 1: Any changes to the CH database should be applied to the SG database as well. So if there's new databases for ENVELOPE, SUPPLY and CONVERSION in CH, please do the same for SG

cea/databases/CH/assemblies/SUPPLY_NEW.xlsx
martin-mosteiro, line 1: This file needs to be renamed to **SUPPLY.xlsx**

cea/databases/CH/components/ENERGY_CARRIERS.xlsx
martin-mosteiro, line 1: What is this file used for?

cea/databases/CH/assemblies/SUPPLY_NEW.xlsx
martin-mosteiro: Do the primary/secondary/tertiary components refer to the codes in **CONVERSION.xlsx**? If so, there are components missing there (e.g., BO4, BO5 and BO6) are listed in primary_components but there's no corresponding system on **CONVERSION.xlsx**.

martin-mosteiro: Also, all cooling supply systems have the same primary and tertiary components. Is that on purpose?

martin-mosteiro: Also, why are there no secondary components for any of the systems? And why are there no components for HOT_WATER and ELECTRICITY?

cea/constants.py
martin-mosteiro: Also, when would a value of 60 lead to a "zero result"? Since you're using `np.ceil` the value you get for `ni` is always >= 1

martin-mosteiro (changes requested):
@MMeshkinkiya congrats on your first contribution to CEA! I've made some minimal changes to your PR (formatting things and things like that). I have some comments and requests for change, just to summarize:
1. If you change things in the CH database, you need to make the same changes in the SG database
2. I don't think the reason given for the change in the service life is valid. Unless the new value is the new recommended value from SIA, I would revert to the old value
3. Your equation for the replacement of building components appears to be more complicated than necessary
4. The in-line documentation for your contribution is longer than necessary, I've made a suggestion on how to improve it
5. There are a number of changes to the databases and other unrelated scripts that don't seem to be related to your work. Are those intentional?

cea/analysis/lca/embodied.py
```diff
@@ -203,30 +220,56 @@ def calculate_contributions(df, year_to_calculate):
 
     # calculate the embodied energy/emissions due to construction
     total_column = 'saver'
+
     ## calculate how many years before the calculation year the building was built in
     df['delta_year'] = year_to_calculate - df['YEAR']
     ## if it was built more than 60 years before, the embodied energy/emissions have been "paid off" and are set to 0
```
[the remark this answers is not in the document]
MMeshkinkiya: The embodied emissions are per year. Please see embodied.py, line 259.

cea/analysis/lca/embodied.py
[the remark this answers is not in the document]
MMeshkinkiya: The equation has been corrected as below:
ni=(SERVICE_LIFE_OF_BUILDINGS/Service_Life_Component)-1

cea/constants.py
MMeshkinkiya: The payback period in SIA 2023 /2020 is still 60 years. The longer period is suggested by literature and DUREE project (https://www.aramis.admin.ch/Texte/?ProjectID=38626).

In cea the difference between “construction year“ and “year of calculation” is the basis for calculating building GHG. If the value is larger than 60, cea assumes GHG to be Zero. 
See lines 226-228 in embodied.py.

cea/examples/embodiedUpdated.py
MMeshkinkiya: Adding the script to Git was a mistake and should have been gitignore. It is removed from the latest version

cea/databases/CH/assemblies/SUPPLY_NEW.xlsx
MMeshkinkiya: These are not my commits.

cea/datamanagement/streets_helper.py
```diff
@@ -33,7 +33,7 @@ def calc_bounding_box(locator):
     return result
 
 
-def geometry_extractor_osm(locator, config):
+def geometry_extractor_osm(locator: object, config: object) -> object:
```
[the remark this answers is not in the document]
MMeshkinkiya: These are not my commits.

cea/databases/CH/components/CONVERSION_NEW.xlsx
MMeshkinkiya: These are not my commits.

cea/databases/CH/assemblies/ENVELOPE.xlsx
MMeshkinkiya: This is fixed in the latest commit.

cea/databases/CH/components/ENERGY_CARRIERS.xlsx
MMeshkinkiya: These are not my commits.

cea/analysis/lca/embodied.py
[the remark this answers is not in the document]
martin-mosteiro: Thanks for clarifying that. The output variables should then be renamed `GHG_sys_embodied_kgCO2m2yr` and `GHG_sys_embodied_tonCO2yr`.

cea/analysis/lca/embodied.py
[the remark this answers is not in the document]
martin-mosteiro: Hmmm but that's just the replacements, right? Otherwise if `SERVICE_LIFE_OF_BUILDINGS == Service_Life_Component` then the embodied emissions for that component will be 0.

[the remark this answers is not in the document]
martin-mosteiro: If we just want to calculate the construction emissions + replacements, the original equation makes sense.
If we want to calculate construction and replacements separately, the second equation makes sense, but then we need an additional step before that to calculate the construction emissions.

cea/constants.py
martin-mosteiro: Yes, I'm aware that after the SERVICE_LIFE_BUILDINGS is complete CEA assumes the embodied emissions to have been "paid off". But as currently written a building of more than 120 years of age (say, the ETH Hauptgebäude, for example) will also have 0 emissions. So I don't think that's a good reason to change the value for the service life of buildings in CEA.

An updated value in literature could indeed be a good reason, but then the source should be updated in the comment as well.

martin-mosteiro: Furthermore, I've quickly gone through the DUREE documentation and I'm not sure they _recommend_ a life span of 120 years.

They mention this value as the highest found in the literature:
> The national LCA methods generally uses conventional values for this parameter. In Switzerland, the LCA national method (SIA 2032 and SIA 2040 technical books [12]) proposes 60 years. The SIA 480 standard does not define an RSP but the service life of the building structure instead. The SIA 480:2004 standard [41] considers from 80 to 100 years while the revised 2016 version considers from 40 to 120 years with an intermediate value at 75 years [42]. In addition, the SNARC method, used in early design stages, considers 30 years [43]. Other LCA methods in Europe consider 50 years [44], 80 years [45] or even 120 years [46].

And then they considered a range of possible lifespans:
> Except for the baseline study, the influence of different methodological choices on the LCA and LCC result was studied. In the beginning, the RSP was varied from 30 to 120 years and the results showed that the contributions of the LCA and LCC stages change as a function of the chosen RSP, as well as the magnitude of the Sobol’ Indices. 

And in their recommendations for practitioners they seem to recommend such a scenario-based approach:
> The building RSP is an influential parameter on the LCA and LCC uncertainty. Different scenario analyses are recommended to capture more accurately the variability of the LCA and LCC.

cea/datamanagement/streets_helper.py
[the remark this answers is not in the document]
martin-mosteiro: Screen Shot 2023-07-28 at 5 23 52 PM

Gitblame disagrees 😛 No worries, probably an unintended change.

cea/databases/CH/components/CONVERSION_NEW.xlsx
martin-mosteiro: No worries, please just delete any files that are not an intended part of your PR

cea/databases/CH/components/ENERGY_CARRIERS.xlsx
martin-mosteiro: Ditto

cea/examples/embodiedUpdated.py
martin-mosteiro: Thanks!

cea/databases/CH/assemblies/SUPPLY_NEW.xlsx
martin-mosteiro: Ditto

cea/databases/CH/assemblies/ENVELOPE.xlsx
martin-mosteiro: Thanks!

cea/analysis/lca/embodied.py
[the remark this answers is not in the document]
MMeshkinkiya: The lable of final results are updated in the latest commit.

cea/analysis/lca/embodied.py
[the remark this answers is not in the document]
MMeshkinkiya: I double checked the standards and literature, and noticed that I misinterpreted the equation. To consider the replacement phase in LCA calculations, the original equation from DIN EN 15978 ((section 9.3.3) is proposed as:

ni=(SERVICE_LIFE_OF_BUILDINGS/Service_Life_Component)-1
 
However, all the emissions from production, replacement and end-of-life can be packed into one equation, the one with both ”+1” and "-1" that I wrote originally, which didn’t make sense. In fact, the point of adding the “+1”  was to avoid a value of zero for the components that have an identical service life with the building. Hence, with a bit of modification, the original nonsense equation can be considered as a reference, for calculating the "ni" value for both new and retrofitted buildings. I suggest rewriting the equation as: 

ni= np.max(((SERVICE_LIFE_OF_BUILDINGS/Service_Life_Component)-1),1)
 
Note from literature:
LCEI is composed of the environmental impacts (EI) associated with production of all building components (Module A1-A3), their replacement (module B4) and end of life (modules C3 and C4) as well as the environmental impacts related with building operation (B6). This is translated by the following equation: LCEI = b ∑ i=1 kproduction,i + kEoL, i )* ( ni + 1 ) + QF,a*kop.*RSP*ERA) [https://doi.org/10.1016/j.buildenv.2020.107143]. 
 
The last part of this equation (LCEI) is calculated separately in CEA in operation.py

I updated the equation in the latest commit.

MMeshkinkiya:
Additional note:
I think i forgot to mention this before:
I also have updated some data in FEEDSTOCKS. xlsx for CH database. Accordingly, the GHG_kgCO2MJ for all sheets are updated by the latest version of KBOB  dataset (KBOB / ecobau / IPB  2009/1:2022) .

cea/analysis/lca/embodied.py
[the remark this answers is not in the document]
MMeshkinkiya: Hi Martin,

I see that I forgot to push the above changes into Github from my branch.

shizhongming:
@martin-mosteiro
Today, Maryam and I checked the PR together. We have updated the equations and the annotations. Please take a look again? 
P.S. The Supply_new database is from Mathias' new optimization script.

shizhongming (approved):

cea/analysis/lca/embodied.py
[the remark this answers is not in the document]
martin-mosteiro: Thanks for clarifying that!

cea/constants.py
martin-mosteiro: Ok, looks like the value was reverted to 60. I'm resolving this conversation now.

cea/databases/CH/components/CONVERSION_NEW.xlsx
martin-mosteiro: Looks like this was indeed part of somebody else's commit, so I created a separate issue about that. Sorry for the mixup!

PR #3388, opened 2023-09-19
Saving additional info zone helper

shizhongming:
In addition to the internal CEA handle of `Building Name`, such as B1000, and B1001, this PR updates zone_helper to save additional information, including `house number`, `street name`, `house_name`, `postal code`, city, `residential_type` (to indicate Singapore's HDB), `city`, and `country`.

### DIFF
--- a/cea/datamanagement/zone_helper.py
+++ b/cea/datamanagement/zone_helper.py
@@ -190,6 +190,8 @@ def assign_attributes_additional(shapefile):
     This script fills the zone.shp file with additional information from OSM,
     including house number, street name, postcode, if HDB (for Singapore), city, country
     """
+    # TODO: include different terms used by OSM in different countries
+    #  As of 19 Sept 2023, this function works fine in Switzerland and Singapore
 
     # local variables
     no_buildings = shapefile.shape[0]
--- a/cea/datamanagement/zone_helper.py
+++ b/cea/datamanagement/zone_helper.py
@@ -213,10 +213,10 @@ def assign_attributes_additional(shapefile):
 
     # Assign the cea-formatted columns with attributes
     shapefile['house_no'] = shapefile['addr:housenumber']
-    shapefile['street_name'] = shapefile['addr:street']
+    shapefile['street'] = shapefile['addr:street']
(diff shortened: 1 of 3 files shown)


cea/datamanagement/zone_helper.py
MatNif, line 1: Most of the changes look good to be. The only question I have I about the removal of the string-to-float conversion for some of the parameters: What's the reasoning behind removing that section?

MatNif (commented):
I tested the script on a small residential area in Singapore Woodlands. The changes worked as intended, and the corresponding columns appeared in the Input Editor. The contents of the columns also made sense; only three of the buildings were missing an address - the first corresponded to a small shelter and the other two to an electrical substation. So all of that makes sense.

The only open question in my opinion is the one regarding the string-to-float conversion.

cea/datamanagement/zone_helper.py
```diff
@@ -72,12 +72,6 @@ def assign_attributes(shapefile, buildings_height, buildings_floors, buildings_h
               'if we do not find data in OSM for a particular building, we estimate it based on the number of floors, '
               'multiplied by a pre-defined floor-to-floor height (set to 3m by default)')
 
-        # Make sure relevant OSM parameters (if available) are passed as floats, not strings
-        OSM_COLUMNS = ['building:min_level', 'min_height', 'building:levels', 'height']
-        selected_columns = list(set(list_of_columns).intersection(set(OSM_COLUMNS)))
-        shapefile[selected_columns] = shapefile[selected_columns] \
-            .fillna(1).apply(lambda x: pd.to_numeric(x, errors='coerce'))
-
```
shizhongming, line 80: @MatNif 
Correct me if I am wrong. The original Line 75-80 and Line 81-85 are duplicated. Hence, I remove the Line75-80

MatNif: Sorry, I think you're right. I missed that.

In that case, I think this PR is good to go.

MatNif (approved):
I believe this PR can be merged. The code snippet converting the parameters has been removed since it was duplicated right after.

PR #3406, opened 2023-10-13
Revert "Update pythonocc to enable Mac ARM support"

ShiZhongming:
Reverts architecture-building-systems/CityEnergyAnalyst#3265

PR #3408, opened 2023-10-15
Enable Mac ARM support

reyery:
Reverts architecture-building-systems/CityEnergyAnalyst#3406

Restore commits for Mac ARM support


ShiZhongming:
This is to be merged in Jan 2024.

PR #3409, opened 2023-10-15
Fix network layout script

reyery:


reyery:
Will merge this first and fix mac dependencies after

PR #3458, opened 2024-01-18
Update network-file naming convention

MatNif:
This PR introduces a hashing function to encode the network-connectivity vector if the latter were to get too long.
This effectively shortens the network-file names. These changes are limited to the network script within the new optimisation.

To test:  It might make sense to lower the number of buildings where this change kicks in from 50 to e.g. 10 (connectivity.py line 206) and then run the optimisation script on any district with 11 or more buildings.


reyery:
@MatNif why not hash the name every time instead of just when it is more than 50 characters?

MatNif:
Hey @reyery,
the idea was that you could more easily identify the corresponding network file if part of the connectivity vector was still in its name (if for whatever reason you on might need to look at the details of the network files at some point). But I'm open to change and simplify that.

reyery:
Where are these files stored? I can't seem to find it in my scenarios. 

Hmm does it make sense for the connectivity vector to be part of the filename? You could probably try to encode it in hexadecimal instead of hashing it, since you can convert it back to binary easily.

MatNif:
These files actually only get stored to the temp-files as far as I know. So it's normal that you wouldn't have seen them in your foler structure. But i recall having had to access them during debugging before.

But, you're right. Unless in this very specific debugging situation the files don't really need to be accessed. So it might make more sense to just get rid of the connectivity vector values all together.

As for the hashing function, I'll have to double-check tomorrow, but I thought that particular function is actually reversible. But since it has the added benefit of outputting a fixed length code, I chose that over a hexcode conversion.

MatNif:
Hi @reyery.
I've made a few changes to this according to your suggestions. Can you have another look at it and merge if you're fine with the new version?

PR #3484, opened 2024-02-08
add create-new-scenario to scripts.yml

ddceruti:
Release v.3.35.0 broke the scenario creation for the dashboard, becuase the create-new-scenario script was removed from scripts. This causes the dashboard to fail when creating a new scenario from scratch with the option "Import input files".

To reproduce, use latest master branch and try to set up a new scenario with the option "Import input files".

Error Message:
```
Exception: create_new_scenario: module 'cea.api' has no attribute 'create_new_scenario'
```

### DIFF
--- a/cea/scripts.yml
+++ b/cea/scripts.yml
@@ -346,6 +346,13 @@ Utilities:
     module: cea.datamanagement.data_initializer
     parameters: ['general:scenario', data-initializer]
 
+  - name: create-new-scenario
+    label: Create new scenario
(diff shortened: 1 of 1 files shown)


reyery:
Hi @ddceruti, thanks for pointing this issue out. 
The initial idea was for us to remove the `create_new_scenario` script and replace it with something else in the future. Unfortunately we did not consider that it would break the dashboard now. We can merge this PR with a few changes to fix that.

Issue #3495, opened 2024-02-19
Surroundings helper behaves wrongly for custom building footprints


MatNif:
**Describe the bug**
When uploading .shp files to CEA with entirely new building footprints for an area (e.g. in a redevelopment scenario), the surroundings helper would include all buildings that don't precisely match the uploaded footprints. This can easily lead to overlaps of uploaded buildings with existing buildings. A behaviour that is likely unexpected for many users.

**To Reproduce**
1. Create new Scenario and select "Import input files"
2. Upload a .shp file with building-footprints that differ from the existing buildings in the area
3. Run surroundings helper
4. See error in Input Editor

**Expected behaviour**
I think the easiest workaround here would be to change the checks in 'surroundings helper' to eliminate buildings that partially overlap with selected buildings in the area.

**Screenshots**
![image](https://github.com/architecture-building-systems/CityEnergyAnalyst/assets/93093343/2926660b-346f-49bb-b995-ecd11cc640c9)

PR #3496, opened 2024-02-19
Fixing surroundings helper with uploaded building footprints.

MatNif:
Deleting surrounding buildings that overlap with the footprints of zone-buildings.

To test, refer to issue #3495 .


ShiZhongming:
I was able to replicate the issue with CEA v3.35.3.
This PR solves the issue, and it will be included in CEA v3.35.4.
Thank you @MatNif

PR #3510, opened 2024-02-29
Add workflow to build mac dmg

reyery:


reyery:
@ShiZhongming the .dmg for macos-arm can be downloaded [here](https://github.com/architecture-building-systems/CityEnergyAnalyst/actions/runs/8066319785/artifacts/1279068202)

Issue #3520, opened 2024-03-10
Browsing weather file with Weather Helper


Khayatian:
**Describe the bug**
When using the Weather Helper tool, the button "+ Browse for weather file" does not work.

**To Reproduce**
1. Go to 'Weather Helper'
2. Click on '+ Browse for weather file'
3. Does not open a new window for selecting an *.epw file.

**Expected behaviour**
In the previous versions, a window would pop-up for selecting a weather data from a local drive.

**Hardware (please complete the following information):**
 - OS and version: [Windows 10 Enterprise]
 - Version of CEA: [v. 3.34.2]

ShiZhongming:
Dear @Khayatian 
Thanks for rasing this. This is an known issue. It is affecting many users. We will fix it on the next release. @reyery has a provisional solution.

`

For the "+ Browse for databases path" button bug, we have fixed it for the GUI but you would have to manually update it on your end for now, without having to reinstall it.

If you are using Windows, you can extract the attached file and replace the app.asar file that is found in your CEA installation path:
e.g.

```
C:\Users\foobar\Documents\CityEnergyAnalyst\dashboard\resources\app.asar
```
[app.asar.zip](https://github.com/architecture-building-systems/CityEnergyAnalyst/files/14422643/app.asar.zip)

Let me know if that works for you.`

Khayatian:
Dear @ShiZhongming,

Thanks for the patch. Worked perfectly well. If you don't mind, I am closing this issue since a solution is in the pipeline for the next release.

PR #3525, opened 2024-03-14
Fix Windows setup bugs regarding developer version

reyery:


ShiZhongming:
@reyery 
Does this include the fix for this? #3520

reyery:
> @reyery Does this include the fix for this? #3520

It only fixes the Windows setup which does not include the gui. 
But building a new setup would include changes that have fixed the issue in the gui.

reyery:
@ShiZhongming anyway I have removed the shortcuts for CEA config, because it was creating conflicts for non-admin users when the "All user" option is chosen. The shortcuts were pointing to the config file in the admin location instead of the user

PR #3549, opened 2024-04-15
Handling zero interest rates

MatNif:
This introduces a simple fix to calculating annualised capex and opex with an interest rate of 0. This is necessary since the basic annuity formula is undefined for an interest rate of zero.

In addition, the documentation of these methods was updated and a source explaining how the annuity factor is derived has been added.

### DIFF
--- a/cea/analysis/costs/equations.py
+++ b/cea/analysis/costs/equations.py
@@ -26,7 +26,7 @@ def calc_opex_annualized(OpC_USDyr, Inv_IR_perc, Inv_LT_yr):
     :param Inv_LT_yr: Lifetime in years
     :return: Annuitized Operating Expenditure in USD per year
     """
-    if Inv_LT_yr == 0:
+    if Inv_IR_perc == 0:
         return OpC_USDyr
 
     Inv_IR = Inv_IR_perc / 100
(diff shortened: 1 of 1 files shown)


cea/analysis/costs/equations.py
```diff
@@ -17,6 +17,18 @@
...
+    """
+    Calculate Annuitized Operating Expenditure (OPEX) -
+    Calculations based on http://fahmi.ba.free.fr/docs/Courses/2012%20HEC/FBA_FE_Chap1_time_value_derivation.pdf
+
+    :param OpC_USDyr: Operating Expenditure in USD per year
+    :param Inv_IR_perc: Interest Rate in percentage
+    :param Inv_LT_yr: Lifetime in years
+    :return: Annuitized Operating Expenditure in USD per year
+    """
+    if Inv_LT_yr == 0:
```
reyery, line 29: Is this supposed to check for the interest rates instead of the lifetime variable?

MatNif: You're obviously right. Very bad oversight on my part. I just fixed it.

reyery (approved):

PR #3550, opened 2024-04-16
Optimisation - Calculate the baseline system

MatNif:
**Purpose**
Until now, the new optimisation script calculated supply system structures, operational characteristics and network layouts for all of the selected optimal energy systems, but did not offer any way to calculate these systems to the baseline of the district established in the input editor.

This introduces a calculation of the district's initial/base energy system, including the corresponding networks and supply systems.  These new changes will be relevant to allow the users to perform simple comparisons between optimal and baseline systems when corresponding graphics will be introduced to the Dashboard.

**To test**
1. Select a simple district use case & run all scripts found under the `Data management` and `Demand forecasting` sections under `Tools`.
2. Edit the default supply systems of the buildings found under the `supply-systems` section of the input editor
3. Run the new optimisation script of a district of your choice.
4. Check the newly added `current_DES` folder that is output in the `optimisation_new` folder found under `outputs`.


reyery:
@MatNif would that mean that the `current_DES` files could be used to show the initial network in the input editor map? But the user has to run the optimization script first for the files to be available is that right.

MatNif:
Hi @reyery, 

Good question. Currently the input editor displays a network connecting all buildings that have a 'DISTRICT'-scale supply system (even before the changes made in this branch).  Now with the new changes, if these supply systems differ (e.g. B1000 has a `SUPPLY_COOLING_AS3`-system and B1001 has a `SUPPLY_COOLING_AS4`-system) the optimisation script would place them in two different networks and create corresponding .geojson-files for the networks.

So, yes, the new results could be used as a basis for a more accurate representation of the current network connections. What I'm not too sure about is what type of file the dashboard currently creates as a basis for the network display. I could very well imagine that that file differs from the optimisation script's current outputs.

PR #3567, opened 2024-04-29
Fixing oudated excel writer

MatNif:
Updating Excel writer to work with the latest version of the `pandas` and `OpenpyxlWriter` libraries.

### DIFF
--- a/cea/utilities/reporting.py
+++ b/cea/utilities/reporting.py
@@ -33,12 +33,11 @@ def full_report_to_xls(tsd, output_folder, basename):
     #timestamp = datetime.datetime.now().strftime('%Y-%m-%d-%H-%M-%S')
     #output_path = os.path.join(output_folder,"%(basename)s-%(timestamp)s.xls" % locals())
     output_path = os.path.join(output_folder, "%(basename)s.xls" % locals())
-    writer = pd.ExcelWriter(output_path, engine='xlwt')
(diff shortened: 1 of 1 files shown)


cea/utilities/reporting.py
```diff
@@ -33,12 +33,11 @@ def full_report_to_xls(tsd, output_folder, basename):
     #timestamp = datetime.datetime.now().strftime('%Y-%m-%d-%H-%M-%S')
     #output_path = os.path.join(output_folder,"%(basename)s-%(timestamp)s.xls" % locals())
     output_path = os.path.join(output_folder, "%(basename)s.xls" % locals())
-    writer = pd.ExcelWriter(output_path, engine='xlwt')
+    writer = pd.ExcelWriter(output_path, engine='openpyxl')
 
     df.to_excel(writer, na_rep='NaN')
 
     # Close the Pandas Excel writer and output the Excel file.
-    writer.save()
```
reyery, line 41: Actually, I think it would be better to use a context manager for the writer instead.
So it would close the file handle automatically.

e.g.
```python
with pd.ExcelWriter("path_to_file.xlsx", engine="openpyxl") as writer:
    df.to_excel(writer, na_rep="NaN")  
```

https://pandas.pydata.org/docs/reference/api/pandas.ExcelWriter.html

PR #3570, opened 2024-05-02
removing _new.xlsx in the databases

ShiZhongming:
This PR is part of the task to phase out the older optimisation script with the new optimisation script.

The beta version of the new optimisation script introduced two _NEW.xlsx files into the database. They served the new optimisation scripts well yet caused confusion. This PR merges the _NEW.xlsx files into the original databases.

All the SG, CH, and DE databases have been updated.


ShiZhongming:
@MatNif Thanks for fixing this!
Will you have time to take a look at the Tab Boiler with the few newly added boiler types in the DE database? ; )

MatNif:
Hi @ShiZhongming,
Sure, not sure I'll have time this weekend but I can check this out early next week at the latest.

Issue #3555, opened 2024-04-18
Adiabatic surfaces are apparently not considered?


martin-mosteiro:
**Describe the bug**
Per issue #443 and pull request #2416, adjacent buildings in CEA are _supposed to_ be considered as adiabatic surfaces. However, I tested simulating the demand for a building surrounded on three sides by other buildings compared to the same building after removing all surrounding buildings:
![Screenshot 2024-04-18 at 16 20 05](https://github.com/architecture-building-systems/CityEnergyAnalyst/assets/18548065/6cd34d47-eb24-4bb5-a90d-9c3ac6c96505) ![Screenshot 2024-04-18 at 16 20 30](https://github.com/architecture-building-systems/CityEnergyAnalyst/assets/18548065/536ff4d0-5374-4d69-857b-afe8232e99e4)

After running the solar radiation and the energy demand simulations for both cases, I get the following results:
![Screenshot 2024-04-18 at 16 18 45](https://github.com/architecture-building-systems/CityEnergyAnalyst/assets/18548065/30f89f27-2d4f-4179-9fe4-fb81056e96bc)
The plot on the top shows the results for the building surrounded on three sides, the bottom one shows the isolated one. As expected, the solar gains increase significantly when the building is isolated. However, the transmission losses through the envelope barely change – for example, the difference in transmission losses through the walls in January is 1.6% higher when the building is fully detached. 

Here's a scatter plot for the three most relevant parameters for both cases:
![image](https://github.com/architecture-building-systems/CityEnergyAnalyst/assets/18548065/e9675a38-24b3-4e99-a4a5-e09a6864d0f5)
The solar gains for the detached case go up by up to 8 times, but the transmission losses are basically linear.

Having investigated a bit more, the parameter implemented in PR #2416 appears to have been made optional for the sake of speed (`config.radiation.consider_intersections`) but as a result buildings are modeled as though they were fully surrounded by outdoor air.

**To Reproduce**
1. Create a scenario with adjacent buildings
2. Run the solar and demand simulations
3. Create a new scenario with the same buildings but remove attached buildings
4. Run the solar and demand simulations for the new scenario
5. Compare results

**Expected behaviour**
Surfaces attached to adjacent buildings should be treated as adiabatic.

reyery:
Does having `config.radiation.consider_intersections` set to true solve this issue? I'm not too sure how much of a speed loss it has on runs.

martin-mosteiro:
[...]

PR #3576, opened 2024-05-09
Consider adjacent walls by default

martin-mosteiro:
As discussed in issue #3555, adjacent walls were not considered by CEA by default. This means that there was artificial insolation on attached walls and adiabatic surfaces were neglected (i.e., walls were assumed to be in contact with exterior air). This is not consistent with the SIA and ISO standards on which the calculation is based, but it does have the advantage of somewhat lower simulation times.

Therefore, the parameter has been redefined: rather than users whether they would like to `consider-intersections`, they are now asked whether they would like to `neglect-adjacent-buildings`. The default value is `False`, as that is the more correct simulation method, but users may set this parameter to `True` if they are okay with less accurate simulations in exchange for somewhat faster simulation times.


reyery (approved):
Looks good to me.

ShiZhongming:
Looks good to me too. Thank you @martin-mosteiro 
Will merge when we fix the reference-case tests.

Issue #3573, opened 2024-05-08
questions about the DE Database


ShiZhongming:
Hi @ddceruti 

Thank you again for leading the German (DE) Database.
I am opening this issue to check with you about two points in the DE database.

1. Under `assmblies/supply.xlsx, Tabs Heating and Hot_water`, multiple new entries of technologies have been added, however, they are not linked under `archetypes/construction_standard.xlsx, Tab Supply`. Is this intended? If not, do you think it is possible to update them accordingly?
2. Under `assmblies/supply.xlsx, Tabs Heating, BO4,5,6,7` are added yet defined under `components/conversion.xlsx, Tab Boiler`. Could define `BO4, 5, 6, 7` and add them to `components/conversion.xlsx, Tab Boiler`?

These two points are affecting Optimization script.

Thanks and regards, Zhongming

ddceruti:
Hi,
I can update them, it will take a while because I will have to go through all TABULA standards manually and try to assign a modeled supply technology. I will try to get to it during this week.

ShiZhongming:
Thank you! @ddceruti

ddceruti:
Hi @ShiZhongming,
I have  two questions regarding the supply and conversion components. 

1) How is the total capex calculated? Because both have investment costs and different sizes available. For example:
Boiler section of conversion:
![grafik](https://github.com/architecture-building-systems/CityEnergyAnalyst/assets/52672045/9e4b46ca-d997-4794-9e97-cffe20b2a4b6)
Supply section of heating:
![grafik](https://github.com/architecture-building-systems/CityEnergyAnalyst/assets/52672045/bd1c48c3-219d-41ad-bb96-b55f33e1ec75)
What are I am supposed to define in these costs and efficiencies?

2) Can we define two conversion technologies for heating? I suppopse yes?
![grafik](https://github.com/architecture-building-systems/CityEnergyAnalyst/assets/52672045/8f9c40a6-aad0-4d4b-b6d0-e3d3bdfec3ed)


>     1. Under `assmblies/supply.xlsx, Tabs Heating and Hot_water`, multiple new entries of technologies have been added, however, they are not linked under `archetypes/construction_standard.xlsx, Tab Supply`. Is this intended? If not, do you think it is possible to update them accordingly? 
I have already implemented the heating and hot water supply technologies for the TABULA categories.

MatNif:
Hi @ddceruti 

[...]

PR #3601, opened 2024-06-04
DE database supply technolgies fix

ddceruti:
I assgined all TABULA supply technologies and addded all required technologies with costs and efficienceies to the BOILER and HEATPUMP sheets in the CONVERSION excel file as discussed in the issue #3573. I also updated the documentation accordingly. As the CAPEX can be calulcated in two places at the same time, I have left the SUPPLY file cost parameters untouched and just assigned the technologies. 

@MatNif can you provide some feedback if the changes are enough and if you require some more changes?

Disclaimer: the main cost source of KEA Baden Württenberg is quite a lot cheaper than the current CH database costs - mainly due to the very high fixed costs defined in CONVERSION. For a 60kW gas boiler: old CAPEX = 40 800 USD; new CAPEX = 6400 USD.
Maybe we can discuss some approaches to homogenize the costs. 

@ShiZhongming, @ whole CEA dev team: these changes are very difficult to follow since excel files are not supported properly and you cannot see the changes line by line. If more countries want to contribute their databases, this should be changed urgently to allow proper collaborative workflows.


MatNif:
Hi @ddceruti, as far I can tell so far, the changes look good and the databases look complete.
As you correctly mentioned though, it is a bit difficult to review the details of the changes made in the databases. As such I still need a bit more time for testing since in the process of testing the new databases I have run into a related issue but that doesn't concern your changes directly. 

The issue is more with how the developersversion of CEA loads the databases for me. While the DE databases of my CEA installation were changed according to your latest updates within the CEA installations database folders, when I created a new test case the databases in the inputs did not adequately reflect these changes, but loaded the outdated versions of the databases instead. As expected, this lead to the previously encountered crash of the optimisation script.

@reyery or @ShiZhongming, do you have an idea of how to change/fix this behaviour? As far as I could determine when I ran the data-initialiser, the databases actually are loaded in from the dependencies instead of the CityEnergyAnalst folder, specifically from:
`- data-initializer:databases-path = C:\Users\nimathia\Documents\GitHub\CityEnergyAnalyst\dependencies\micromamba\envs\cea\lib\site-packages\cea\databases\DE`
I assume the intended behaviour would be that the database in the dependencies gets updated adequately when the github branch that the installation points to changes. That did not happen in my case.

ShiZhongming:
Thanks @ddceruti and @MatNif !
Rey and I will take a look at this next week.

reyery:
@MatNif are you running it via pycharm or the CEA console. It should change accordingly when CEA is installed as "editable" in the python environment. i.e. `pip install -e .`

Could you try running `pip list` and see if `cityenergyanalyst` has a `Editable project location`?

MatNif:
Hi @reyery.

I ran the majority of the scripts through the dashboard (including the `Database initializer`).
And, yes, you were right. I seem to have mad a mistake in the installation process indeed.

After reinstalling in 'editable mode', the path in the database initialiser now points to the correct location as well, i.e.:
`- data-initializer:databases-path = C:\Users\nimathia\Documents\GitHub\CityEnergyAnalyst\CityEnergyAnalyst\cea\databases\DE`

So, in my opinion, the database updates are now functional and good to go. The layout of the `CONVERSION` database is a bit messed up in parts (e.g., the boilers tab), but that doesn't prevent it from working as intended.

ShiZhongming:
Thank you @ddceruti for providing this. Also many thinks to @Mar-Ges @ululele.
Thanks to @MatNif for reviewing it.

PR #3621, opened 2024-06-25
Optimisation: Making base DES part of initial population

MatNif:
Conversion of the optimisation algorithm is still a bit slow. Consequently, it can happen that the identified 'optimal' solutions are sometimes worse than the initial district energy system in a district. (Especially when the number of generations is set at a very low number).

To avoid that in the future, this change makes the original DES part of the initial population for the genetic algorithm. This guarantees, that all of the identified 'optimal' solutions are at least as good as the original system. This improved starting point should hopefully also accelerate conversion of the algorithm as a whole.

**To test**:
- Run optimisation on your district of choice (set `ga-number-of-generations` to 1)
- Print the Pareto front of the solution (using the `visualisations_for_new_optimisation_script` branch)
- Compare the points in the Pareto front with the objective-function values of the current district energy system.

If the current district energy system lies in or above the Pareto-front, the change likely works as intended (you can run a second case to double-check)


ShiZhongming:
Hi @MatNif 

Thanks for the PR!
I am in the process of testing it. I have some questions and let's chat during the monthly assembly.

ShiZhongming:
This PR has been tested both in cases sg and ch. Worked as intended.

PR #3634, opened 2024-07-17
New operations for improved network generation in genetic algorithm

MatNif:
The optimization algorithm previously managed to find converged solutions for thermal supply systems rather quickly, while the networks of the non-dominated solutions often looked unrefined at best and completely counterintuitive at worst.

These changes add a series of operations to the algorithm that can lead to better network structures faster. More explicitly, these new methods are:

1. ConnectivityVector mutation and crossover algorithms that are based on a clustering of the buildings in the domain (done at the start of the run):
- The new **ClusterSwitch** mutation by switches all buildings in the same cluster to a random other connectivity value with a probability of mut_prob.
- The new **ClusterSwap** crossover method recombines two connectivity vectors by exchanging the connectivity values of some of their clusters. Outliers remain unchanged by this operation.
- The new **ClusterAlignment** crossover operation recombines two connectivity vectors by exchanging the connectivity values of the buildings where (a.) they deviate from the most prevalent connectivity value of the cluster they belong to and (b.) a swap of the connectivity value between the two connectivity vectors would align them with the most prevalent connectivity value of their cluster.

Preclustering of the buildings either follows:
- The [Louvain method for community detection](https://en.wikipedia.org/wiki/Louvain_method) or
- The [DBSCAN algorithm](https://en.wikipedia.org/wiki/DBSCAN) 

2. New ConnvectivityVector correction methods for network overlaps are also introduced. The user gets to select if ConnectivityVectors should be altered so that:
- One of the overlapping networks gets **deleted** 
- Overlapping networks get **merged**
- One of the networks gets **cut** on overlapping nodes (one randomly selected remaining piece of the network will be kept)
- **None** of the above happens (this is mostly introduced to check for potential biases these operations might introduce)

Or if one of these four operations should be chosen at **random**.

The parameters for these new operations have been added to the config. Default parameters have been set to **ClusterSwitch**, **ClusterSwap**, **Louvain** and **Random** based on a first series of tests.

**To test this branch**
1. Select a district of choice (ideally more than 10 buildings)
2. Run the optimisation with an old set of settings (e.g. "ShuffleIndexes", "Uniform", "Louvain" and "None", the clustering method will be ignored if the crossover and mutation algorithms are not based on a clustering)
3. Run the optimisation with a set of new operations
4. Compare the networks-graphs and objective function values of the two runs. While improvements in objective function values will likely be minor there should be a clear visual improvement in the network-graphs.


MatNif:
Sorry. Looks like I made a mistake when I initially created this branch here and changes from a branch for visualisations for the optimisation script somehow made their way into this branch. I'll sort out the situation tomorrow (will probably have to create a new PR).

PR #3636, opened 2024-07-19
Introduce result-retention toggle

MatNif:
Introducing a toggle to let the user decide if they want to prevent the results of a specific run from being overwritten by the next run.

- If the toggle is **off** (default) results will be saved in 'outputs/data/optimisation/centralised', which can be overwritten by new runs.
- If the toggle is **on**, the results will be saved in 'outputs/data/optimisation/centralised_run_X', where X=1,2,3,... depending on how many previous runs have been saved. I intentionally didn't go for the timestamp solutions because I wanted to keep naming as simple as possible, and the timestamp can already be found in the file explorer.

**To test**
Run the optimisation script multiple times over, with the toggle activated or deactivated.


ShiZhongming:
This PR has been tested. It run as decribed.

reyery (approved):

PR #3650, opened 2024-08-28
Fix 'None'-setting for 'network-overlap-correction'

MatNif:
A new 'networks-overlap-correction-method' setting has been introduced for the centralized optimization in the last release v3.38.0.

The 'None' option for this setting was not properly handled and let to a crash of the script.

This change introduces a proper handling of that option and therefore resolves the issue.

**To Test**:

1. Select a district case study of your choosing.
2. Run all scripts until up to and including 'Demand forecasting' > 'Building energy demand'
3. Open the settings page of 'Optimization' > 'Supply system part II (centralized)'
4. Under 'Optimization algorithm parameters' set the 'networks-overlap-correction-method' setting to 'None'
5. Run the optimization script

If the script runs without failing, that means the fix works as intened.


ShiZhongming:
It worked as described. Thank you @MatNif

PR #3664, opened 2024-10-01
fixing the messing columns for pipe properties

ShiZhongming:
This fixes Issue #3562 

How to test?
Use a smaller scenario and run network part 2 (detailed)
Compare results `DH__velocity_edges_mpers` and `DH__massflow_edges_kgs.csv`
The results (v) in the former .csv and results (m˙) in the latter .csv:

v=ρ⋅Am˙​

Here, v (velocity) depends on m˙ (mass flow rate, check DH__massflow_edges_kgs.csv`) and A (cross-sectional area of the pipe, which depends on the pipe diameter, check `Distribution.xlsx` in the Scenario Input database) and ρ equals 998kg/m3.

Can you be the reviewer of this PR?  @KFSchlager 

Thanks, Zhongming

### DIFF
--- a/cea/technologies/thermal_network/thermal_network.py
+++ b/cea/technologies/thermal_network/thermal_network.py
@@ -14,6 +14,7 @@ import geopandas as gpd
 import networkx as nx
 import numpy as np
 import pandas as pd
+import re
 
 import cea.config
 import cea.inputlocator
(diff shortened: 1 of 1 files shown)


ShiZhongming:
This PR functioned as intended after being tested by @KFSchlager

Issue #3619, opened 2024-06-24
Energy Demand too low TABULA Germany


pnguyen-dh:
**Describe the problem**
In the newest version, CEA already implements the data for Germany from TABULA. So I use this database for my simulation, I would expect the energy demand (specific heating demand) should be roughly the same as in TABULA, or maybe about 10-20% different. But after the simulation the building's energy demand are too low, I expect 150 kWh/(m2a) ([source](https://webtool.building-typology.eu/#vares)), but get only about 15-65 kWh/(m2a). I would say that the simulation demand were too good to be true. 

**To Reproduce**
If you need the input/output data, I can zip it and send it. 

**Expected behaviour**
The energy demand should be then higher? Or I actually might miss something. 

**Screenshots**
![image](https://github.com/architecture-building-systems/CityEnergyAnalyst/assets/119336539/9a73559f-b40f-41bd-bee9-799580edc784)

ddceruti:
Hi! Thanks for your feedback and using the database! Can you provide an input file to reproduce (maybe 1 or 2 synthetic buildings)? I would look into it and see if it depends on the TABULA values within the database or some other database parameter.

pnguyen-dh:
[inputs.zip](https://github.com/user-attachments/files/16014890/inputs.zip)

Hi, here is the input. I also tested the same buildings with the database from CH, and the energy demands are much higher. I also tried to look into the DE database but couldn't find anything special. Feel free to tell me if you need anything else.

ShiZhongming:
Hi @pnguyen-dh 

Thanks for raising this issue. It is very much appreciated as the DE Database is new and needs time to stabilise.

I have also used a random single-family house neighbourhood near Berlin. 
For buildings using SFH_F, the heating demand I got was ranging from ~90 kWh/m2 - ~120 kWh/m2, which is rather similar to the reference value of TABULA at 139.5. Please note that in the graph, the heating energy demand calculated is based on a floor area equaling 0.85 X gross floor area, while energy end-use intensity equals this heating energy demand divided by 100% of the gross floor area. This could be one of the reasons that lead to the low normalised results in the graph.

For buildings using SFH_F_NR, the heating demand I got was much lower than that of TABULA, which is as what you had described.

We will need a bit more investigation over this. I am also tagging @ddceruti here. Please share your opinion on this.

[...]

PR #3676, opened 2024-11-04
De fix

ddceruti:
Fixed a bug where the hot water profiles were equal to the occupancy for the single and multi residential types. 
Addionally updated use types with new statistical information on setpoints and hot water consumption according to BDEW (see documentation with updated sources).

This alleviates but does not solve the problem where the heating need is underestimated in issue #3619.


ShiZhongming:
Thank you @ddceruti for maintaining the German Database.
I just tested and it worked well.

PR #3736, opened 2024-12-10
DE_Ns_CONSTRUCTION_STANDARD.xlsx_fix

kleme66:
**German Database fix:**
Small fix regarding the Fraction of net gross floor area (Ns) inside the file CONSTRUCTION_STANDARD.xlsx.
The fix only applies to NWG type, setting **Ns** to the correct value and **Hs_bg** to 0. Previously the two columns were swapped.

**Previously behaviour**
Previously, when Ns = 0, the occupancy was estimated as 0, causing other calculations to be incorrect.

Ns_fix


ShiZhongming:
Hi @kleme66 Thanks for the PR.

Hi @ddceruti Could you take a look at this PR? Thanks!

ddceruti:
> Hi @kleme66 Thanks for the PR.
> 
> Hi @ddceruti Could you take a look at this PR? Thanks!

Works as intended, discussed with @kleme66 and checked the excel file.

Issue #3282, opened 2023-01-27
How can a horizontal panel produce more per m2 of panel than a tilted panel (latitude value)?


RicardoGomesIST:
### Discussed in https://github.com/architecture-building-systems/CityEnergyAnalyst/discussions/3281

<div type='discussions-op-text'>

<sup>Originally posted by **RicardoGomesIST** January 27, 2023</sup>
Hi,
I am using CEA to calculate PV panel potential. I am testing a very simple case for Lisbon (latitude =38º) - just one building with surrounding buildings lower than it. 
![image](https://user-images.githubusercontent.com/52715280/215099185-bb38b332-4842-42e0-a453-f215ea843a20.png)

I created two scenarios one with:

- tilt=1º - (as 0 is not possible) so horizontal

and other with: 

- tilt= 30º

Assuming that the tilted panels face south, I was expecting that the PV production (kWh/m2) would be higher for the tilted than for the horizontal. 

What I get was:
- tilt=1º - (as 0 is not possible) so horizontal
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/Ricardo/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/Ricardo/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

Name | PV_roofs_top_E_kWh | E_PV_gen_kWh | radiation_kWh | PV_roofs_top_m2 | Area_PV_m2
-- | -- | -- | -- | -- | --
B1000 | 134821.9 | 134821.9 | 1046941 | 655.66 | 655.66

</body>

</html>

Doing the calculation:
kWh/m2=E_PV_gen_kWh /Area_PV_m2= 205.6

--

</body>

</html>

- tilt= 30º
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
[...]

PR #3763, opened 2025-01-21
Fix PV potential calculation based on tilt angle

martin-mosteiro:
This PR fixes issue #3282. As discussed there (and in #3281), tilted PV panels were producing _less_ electricity than flat ones, even at the optimal tilt angle. After investigating, I realized there was an inconsistent sign convention in the azimuth angle between the geographic coordinates and the ones used by the PV model (from Duffie and Beckman, 2013). After fixing this issue, you indeed get the expected improvement in PV performance as a function of tilt angle. In the process of fixing this issue, I have also found an incorrect constant (_K=4 m^-1_ from Duffie and Beckman was incorrectly coded as `K=0.4  #m^-1`) and one equation that inexplicably didn't match Duffie and Beckman (2013) (eq. 5.4.2 had a factor of `59.68` instead of _59.7_), with marginal effect on the results. This PR fixes those as well. Finally, I have also changed the solar equations so an angle of 0° (flat placement) is possible.

Before fixing the issue, here are the results for different PV tilt angles for one case study building:
- 0°: yearly 150.21 kWh/m2, peak 121.78 kW/m2
- 10°: yearly 135.53 kWh/m2, peak 111.63 kW/m2
- optimal (~34°): yearly 97.22 kWh/m2, peak 73.65 kW/m2

After fixing the issue, here are the results for the same PV tilt angles for the same case study building:
- 0°: yearly 148.99 kWh/m2, peak 120.85 kW/m2
- 10°: yearly 161.27 kWh/m2, peak 128.49 kW/m2
- optimal (~34°): yearly 175.15 kWh/m2, peak 135.84 kW/m2

I have also fixed the azimuth issue on the PVT calculation. I am not sure why the calculation for PVT used a legacy function for angle of incidence (previously used for PV until @shanshanhsieh changed it in 2021), so I've changed it to `pvlib.irradiance.aoi` to match the PV calculation. After doing so, the PVT potential goes from 134.4 kWh/m2 at 10° and 101.22 kWh/m2 at 30° to 158.27 kWh/m2 at 10° and 171.2 kWh/m2 at 30°.

I have done the same for the angle of incidence for the SC calculation.

**How to test this PR**
1. Use any case study building (or buildings) and run the radiation script (if necessary)
2. Run the PV potential calculation for different tilt angles (e.g., a flat PV panel vs. the optimal angle) using the master branch; the electricity potential per m2 of PV panel should _decrease_ instead of **increasing** when going from a small tilt angle (e.g., flat placement) compared to the optimal tilt angle.
3. Repeat the procedure for the branch in this PR. The electricity potential per m2 should instead **increase** when going from a smaller to an optimal tilt angle.

## Release Notes

- **New Features**
  - Enhanced solar technology calculations with more precise angle and radiation computations.
  - Integrated `pvlib` library for improved solar panel performance modeling.
  - Introduced new function for calculating air mass based on solar zenith angle and geographical coordinates.

- **Bug Fixes**
  - Corrected glazing extinction coefficient in solar constants.
  - Fixed function naming inconsistencies across solar technology modules.

- **Documentation**
  - Added reference citations for solar technology parameters.
  - Updated documentation for the German Building Stock Database, including new naming conventions and sections on non-residential types.
  - Clarified parameter descriptions in configuration files, including the `custom-tilt-angle` parameter.

- **Refactor**
  - Updated solar angle calculation methods.
  - Streamlined solar panel property retrieval processes.

- **Chores**
  - Removed section for installing a developer version of the City Energy Analyst application to simplify the installation process.


martin-mosteiro:
Thanks so much @MatNif for the review! He caught (and fixed) some important issues that only popped up when looking at higher latitudes (such as Trondheim) that caused negative spacing between panels and negative array areas. In the process, I have also added another fix to an equation that only applied to zenith angles of less than 70°, so that now we have a working equation for higher zenith angles.

This PR looks mostly okay now, pending the tests (which are currently broken due to the update in the input file format in PR #3766). After the tests pass, I suggest we run a few more edge cases (e.g., far north, far south, far east and far west) to make sure the PV calculations are accurate and then we'll be ready to merge!

reyery:
@martin-mosteiro unit tests should be fixed now

ShiZhongming:
Hi @martin-mosteiro 
Thanks for the PR again!
Grea the test is working again and I just did some final test of the PR.

The "0" degree flat all ran through for four cases: pv_north (very north in Sweden), pv_ch (zurich), pv_sg (singapore), pv_south (very south in Chile, even more south than argentina hehe).

However, except for the Singapore case worked as intended. All the three had some issues.

For Scenario pv_north, your script triggered "'GB not in expected range'". I had some hacky fixed as the lastest commit and it worked. The results are as intended.

For Scenario pv_ch, "optimal" yields much less electricity than "flat".

For Scenario pv_south, 1 degree tilted yields less than 0 degree and optimal is much less than 0-degree.

I am attaching the scenario I was using via a drop box link, which I sent to you via Slack. Feel free to take a look. Or am I testing them wrongly? Just drop me a message if you need a quick chat over this.

Thank you.
Best, Zhongming

martin-mosteiro:
@ShiZhongming Thanks for your review! Indeed, the "optimal" angle in CEA means maximum kWh/m2 PV *NOT* maximum total PV production (that's the name it had before already btw – not really my preference ;)). Counterintuitive – I believe @MatNif rephrased the description in the GUI to make the naming a little more sane.

I'll check all the other cases you mentioned over the coming days!

ShiZhongming:
Hi @martin-mosteiro 
I have re-tested all scenarios. All results are as intended. Sorry for the confusion.
Just one thing before we merge. Do you think the commit makes sense that I added?

martin-mosteiro (commented):
@ShiZhongming Yes! This makes sense. Not a hack. Please proceed 👍

PR #3791, opened 2025-03-11
Update PV functions to use vectorized operations

reyery:
`np.vectorize` does not make the code any faster as it is still using loops under the hood. 

The functions are updated to use numpy math operations that actually does vectorization. This decreases PV calculation for reference case from ~25s per building to <1s per building.

To test, run PV script without this PR and one using this PR, then compare results to make sure it is the same.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined solar energy and radiation computation routines for improved efficiency and consistency.
  - Enhanced type handling to boost reliability and flexibility in solar simulation models.
  - Simplified processing logic by removing redundant operations for faster, more accurate performance.
  - Updated mathematical operations to utilize direct NumPy calls, improving performance.
  - Removed unnecessary vectorization for several calculations, allowing functions to handle array inputs directly.

### DIFF
--- a/cea/technologies/solar/photovoltaic.py
+++ b/cea/technologies/solar/photovoltaic.py
@@ -202,18 +202,16 @@ def calc_pv_generation(sensor_groups, weather_data, date_local, solar_properties
 
         teta_ed_rad, teta_eg_rad = calc_diffuseground_comp(tilt_rad)
 
-        absorbed_radiation_Wperm2 = np.vectorize(calc_absorbed_radiation_PV)(radiation_Wperm2.I_sol,
-                                                                             radiation_Wperm2.I_direct,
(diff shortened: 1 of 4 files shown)


ShiZhongming:
Worked as intended.
This is really great! For the same results for PV and PVT.
What used to be 4 minutes and 16 minutes for PV and PVT are now 40 seconds and 10 minutes.
Hoping to see more such improvement on othe features!

Really nicely done! @reyery

martin-mosteiro:
Indeed, good catch. I usually avoid `np.vectorize` since it doesn't really have any advantages over a normal `for` loop, but it's a good reminder to try to write functions using vector operators whenever possible. Thanks @reyery!

PR #3804, opened 2025-03-30
Rename to radiationCRAX and add binary for macOS

reyery:
## Summary by CodeRabbit

- **New Features**
	- Introduced an updated solar radiation simulation module now labeled as CRAX, featuring enhanced configuration options such as increased grid resolution and new geometry parameters.
	- Distributed additional resource binaries required for CRAX.
	- Improved dashboard logging to better capture errors during user logout.
	- Updated solar radiation analysis feature with a new name, status, and module path reflecting the transition to CRAX.

- **Refactor**
	- Updated naming conventions and function interfaces for a more consistent simulation workflow aligned with the new CRAX approach.
	- Transitioned from the USR model to the CRAX model in various components, ensuring compatibility with new requirements.
	- Updated Dockerfile to reflect changes in base images and resource paths for CRAX.
	- Added a method to ensure parameter order during script initialization to prevent unexpected results.


ShiZhongming:
I am so happy that this is in place now!
What used to take 6.6 min for a Zurich case study is now 1.17 min.
Looking forward to seeing the new sensor feature to further reduce the simulation time.

ShiZhongming:
Caught into errors when running a larger case with 300+buildings in Sydney.
I think we should merge and fix later.
`Grid generation of CEA finished in 4.51 mins
Running CRAX radiation calculation

--------------------------------------
Script failed. Execution time: 270.69s

ERROR: Error running radiation.exe:
libc++abi: terminating due to uncaught exception of type std::invalid_argument: stod: no conversion`

PR #3861, opened 2025-07-31
Move void_deck to zone after meeting

yiqiaowang-arch:
The goal of this pull request is to move the definition of void_deck to all three types of users of CEA:
- CEA-3 users that wish to continue using CEA-4;
- Current CEA-4 users;
- future CEA-4 users.

The goals are defined as such:

### General migration:
- [x] existence of `void_deck` should be checked in `zone.shp`. `void_deck` in `envelope.csv` should just be ignored.
- [x] All databases should NOT have `void_deck` definitions (in CH, DE and SG), even though SG definition originally has two archetypes with `void_deck == 1`.

### For CEA-3 users: 
- [x] `void_deck` in migration should exist in zone instead of in `envelope.csv`. Its value should be copied from the old `architecture.dbf` or `architecture.csv` file (which exists only in CEA3 structure) and merged into zone.

### For existing CEA-4 users: 
- [x] `void_deck` must be read from file in the following scenarios: 
    - geometry generation for DAYSIM
    - `BuildingProperties` initilization
    - occupancy calculation
    - embodied emission calculation
  
  In each scenario, if `void_deck` column does not exist in zone, this column will be **CUT** from `envelope.csv` to `zone.shp`. Later, both updated files will be saved. If `void_deck` already exists, nothing happens.

### For upcoming CEA users: 
- [x] During scenario creation from OpenStreetMap, `void_deck` should be initialized as 0.

## Summary by CodeRabbit

* **New Features**
  * Introduced automatic migration and validation of the "void_deck" attribute in building geometry data, ensuring data consistency across workflows.
  * Added "void_deck" support to zone geometry schema, with validation for correct values.

* **Bug Fixes**
  * Improved handling of missing or outdated "void_deck" data by initializing or migrating values as needed.

* **Refactor**
  * Removed "void_deck" from building architecture and construction standards schemas and code, consolidating its use within zone geometry.
  * Streamlined internal logic to ensure "void_deck" is only processed where relevant.
  * Removed all references to "void_deck" from building property calculations and envelope handling.

* **Documentation**
  * Updated schema documentation to reflect changes in "void_deck" attribute handling.

* **Chores**
  * Added `.vscode/` to `.gitignore` to exclude Visual Studio Code configuration files from version control.


cea/datamanagement/format_helper/cea4_migrate.py
```diff
@@ -526,6 +525,7 @@ def migrate_cea3_to_cea4(scenario, verbose=False):
                         typology_df.rename(columns=columns_mapping_dict_typology, inplace=True)
                         zone_df_4 = pd.merge(zone_df_3, typology_df, left_on=['name'], right_on=["Name"], how='left')
                         zone_df_4.drop(columns=['Name'], inplace=True)
+                        zone_df_4['void_deck'] = 0
```
yiqiaowang-arch: I have problems with copying the cea3 `void_deck` in `architecture.dbf` column to `zone` in cea4. The `cea4_migrate.py` file seems to discuss a lot of possible scenarios, and I do not fully understand them. Therefore, currently a simple initialization with 0 is implemented.

@ShiZhongming Could you please help me with the copying?

ShiZhongming: Just pushed my edits! thanks @yiqiaowang-arch

reyery:
@yiqiaowang-arch you can run `cea/tests/create_unittest_data.py` to replace the test data if the new result is the expected one. I am guessing it is correct to assume that heating demand now is lower due to lesser occupied floors being considered.

yiqiaowang-arch:
> @yiqiaowang-arch you can run `cea/tests/create_unittest_data.py` to replace the test data if the new result is the expected one. I am guessing it is correct to assume that heating demand now is lower due to lesser occupied floors being considered.

I believe the test function has more issues than just inaccurate results in B1014. Please see #3866 for further clarification.

reyery:
> I believe the test function has more issues than just inaccurate results in B1014. Please see #3866 for further clarification.

I have "fixed" it in #3869, there is definitely more issues with CEA tests but this hotfix should be sufficient for now. You might need to update the test results again.

yiqiaowang-arch:
> > I believe the test function has more issues than just inaccurate results in B1014. Please see #3866 for further clarification.
> 
> I have "fixed" it in #3869, there is definitely more issues with CEA tests but this hotfix should be sufficient for now. You might need to update the test results again.

Thanks for the changes! The master branch now has the same result as my branch. This is expected since the moving of `void_deck` should not affect any actual simulations.

reyery:
Nice, I will merge once all tests pass.

PR #3862, opened 2025-07-31
Move not merge test

yiqiaowang-arch:
this is a test against github automatic tests, because I suspect that the recent merge from master caused problems in #3861 , that the `Qhs_sys_kWh` is much higher than the ground truth:

```
  File "/Users/runner/work/CityEnergyAnalyst/CityEnergyAnalyst/cea/tests/test_calc_thermal_loads.py", line 105, in test_calc_thermal_loads_other_buildings
    self.assertAlmostEqual(expected_qhs_sys_kwh, qhs_sys_kwh,
AssertionError: 2459012.494 != 603387.053 within 3 places (1855625.441 difference) : qhs_sys_kwh for B1014 should be: 603387.05300, was 2459012.49400
```

I created another branch just before the merge, and cherry-picked the latest two commits: (https://github.com/architecture-building-systems/CityEnergyAnalyst/commit/5ac59fa773e6fd8b69bec5b2039cfc8fc3fd13eb) and (https://github.com/architecture-building-systems/CityEnergyAnalyst/commit/a27793952f03ade1c1e8b858a22c408c723328d0). I expect the tests now to be passed.


reyery:
It seems that B1014 has 5 `floors_ag` and 4 `void_deck`, could the heating demand difference be because now we only consider 1 floor now instead of 5. Now that `void_deck` is properly considered 😅.

image

image

reyery:
@yiqiaowang-arch I think if this is expected behaviour, we can update the test data to the new values to make it pass.

reyery:
https://github.com/yiqiaowang-arch/CityEnergyAnalyst/blob/a27793952f03ade1c1e8b858a22c408c723328d0/cea/tests/test_calc_thermal_loads.py#L105-L107

I just realized that the error message from the test is very misleading. `2459012.49400` is the expected saved value, but `603387.05300` is what was calculated from the current branch. (see above lines in the code)

So the error message should be something like `Expected 2459012.49400 but got 603387.05300` instead.

PR #3873, opened 2025-08-07
a prototype of construction mapping helper

yiqiaowang-arch:
@ShiZhongming the preview of the interface is available here. I will add some functional code later.
image


yiqiaowang-arch:
A functional construction mapping helper is available. Documentations of functions still needed to be added. 
User would only need to set three layers of construction once, and all five types of construction (roof, exterior wall, interior wall, interior ceiling, floor) will be automatically added to scenario's database.

yiqiaowang-arch:
> @yiqiaowang-arch I guess we can close this PR since it is not required anymore?

Yes, I will close it with this comment.

PR #3894, opened 2025-08-29
Update database classes and clean up schemas

reyery:


reyery:
> Hey @reyery, are you planning to merge this change soon? I am also trying to access databases and found that the different naming of same quantity (for example, `U_wall`, `U_roof`, etc.) across envelope databases is very annoying, for example when I try to access U value, or emission across components. As a workaround, I have to create dicts to map a universal key to database-specific values. Does your update include any database-agnostic access? Thanks!

Yeah I am merging it now. RIght now these classes reads the whole database into classes using Dataframes, it does not do any validation yet, which I intend to do once I have time. You should be able to use it to create a CEA database object and excess the properties of the whole database, not sure if it would help with what you need though.

PR #3897, opened 2025-08-31
Preliminary boiler data

MatNif:
Added data for coal-fired, electrical and wood-fired boilers. If possible, this data should be improved based on an extensive component pricing study at some point in the future.


reyery:
@MatNif when I view the current database and validate it with the schemas, it shows me this error.
I am just wondering if it is correct and the schema is wrong i.e. `d` can be negative.

image

MatNif:
Hey @reyery , yeah sorry. I didn't rotic tgat the schema was showing this error.

d can definitely be negative. What shouldn't be negative is the resulting cost value. But I've checked and the cost values of these regressions are all positive.

So in short, yes. The schemas oversimplify this a bit at the moment.

PR #3901, opened 2025-09-03
Rename "occupancy_helper" for consistency and clarity

martin-mosteiro:
On CEA, all the `_helper`s are about retrieving data, except the `occupancy_helper`, which includes occupancy models. I think this is a bit confusing, especially because it makes it harder to differentiate `schedule_helper` from `occupancy_helper`.

Why not just call it `occupancy`, since that is what the script is about? It's basically the same as `thermal_loads` (which isn't `thermal_loads_helper`) or `thermal_network` (which isn't `thermal_network_helper`).

I think this makes things a bit clearer with no downsides!

## Summary by CodeRabbit

* **Refactor**
  * Renamed public entry from "occupancy_helper" to "occupancy" across schemas and public APIs, aligning data producer/consumer declarations.

* **Chores**
  * Script and workflow steps updated to invoke the renamed occupancy entry; runtime logging and script wiring adjusted; script inputs expanded to include building architecture, zone geometry, and weather files.

* **Tests**
  * Unit-data generation and test suites updated to call the new occupancy entry point; test behavior unchanged.

* **Documentation**
  * Config section renamed to "occupancy"; occupancy-model now has choice options and expanded help text; docstrings updated.

### DIFF
--- a/cea/schemas.yml
+++ b/cea/schemas.yml
@@ -1915,7 +1915,6 @@ get_building_internal:
         min: 0.0
   used_by:
   - demand
-  - occupancy
 get_building_supply:
   created_by:
   - archetypes_mapper
@@ -2110,7 +2109,6 @@ get_building_weekly_schedules:
         type: float
   used_by:
   - demand
-  - occupancy
 
 get_costs_operation_file:
   created_by:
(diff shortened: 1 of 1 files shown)


ShiZhongming:
Hi @martin-mosteiro 
Thanks for spotting this and making the changes!
You are right about the naming convention issue, and it does open many more issues for the terms we have been using in CEA. Let's discuss in the next meeting?

PR #3919, opened 2025-10-03
replace py4design in radiation_daysim with compas

yiqiaowang-arch:
this PR replaces the out-of-service py4design package used in the radiation calculation with a currently maintained geometry framework, COMPAS (https://compas.dev/), including `compas`, `compas_occ` and `compas_libigl`. For debugging purposes, one would also need to install `compas_viewer`.
Changes:
1. in geometry_generator, use `compas_occ.brep.bOCCBrep` and `compas.geometry.Polygon` to define building geometries.
2. in `daysim` and in `radiance`, use `compas.geometry.Point` and `compas.geometry.Vector` to define sensor points and its direction.
3. refactor the class `BuildingGeometry` into `BuildingGeometryForRadiation` to avoid naming duplication with the `BuildingGeometry` class for thermal calculation. Also changed it into a `dataclass` for better type hinting. 
4. introduce a new way of generating sensor points on building surfaces (see below for comparison) in `sensor_area_calculator`.

Comparison and consequences of sensor point generation methods:
| branches | window generation | sensor distribution|
|--------|--------|--------|
| master | points are distributed related to world XY image | each wall is turned into a window in the middle, and 8 triangles around it. In total 9 parts are generated. image |
| this PR | instead of triangulation, only 4 trapezoids are generated around the window. image | sensors are distributed related to local XY of the building. This has advantages when the building footprint is close to a rectangle, because it avoids sensor being too close to the edge of the building. image |

Currently, there are a few problems that needs to be addressed:
1. There might be significant differences in radiation result, if a roof plane of the building is partially shaded, because of different distribution of sensor points, especially when another building is floating on top of the roof.
2. as a general problem, underside sensors that's touching the ground receives unexpectedly high radiation (which should not happen at all).
3. it might not work with multi-processing.
4. for very small grid size, the `sensor_area_calculator` is unstable and need more testing.


yiqiaowang-arch:
When the sensor density is very high (1m for both roof and wall), this is what it looks like (each building takes ~80s on my computer with single processor): 
image

yiqiaowang-arch:
@reyery the tests did not pass because of lack of compas environments (including `compas`, `compas_occ` and `compas_libigl`). Could you let me know how to configure the test environments?

reyery:
@yiqiaowang-arch I added it using `pixi add compas compas_occ`, which updates the lock file that the test environment uses. I had to add `compas_libigl` as pip dependency since conda does not provide any target for linux ARM architecture.

yiqiaowang-arch:
> @yiqiaowang-arch I added it using `pixi add compas compas_occ`, which updates the lock file that the test environment uses. I had to add `compas_libigl` as pip dependency since conda does not provide any target for linux ARM architecture.

I am using compas_occ 1.3.0 where the boolean union is possible. However, in the pyproject.toml its version is not specified. I could try to use raw occ.core to union breps if an upgrade to 1.3.0 is not possible. What do you suggest?

compas packages I am using:
```
  compas                         2.14.1                 pyhd8ed1ae
  compas_libigl                  0.7.6                  py312hbb8e
  compas_occ                     1.3.0                  pyhd8ed1ae
```

yiqiaowang-arch:
@reyery I removed `compas_occ` as a dependency by:
1. simplifying point-in-solid check (assuming all buildings are vertical extrusions, might not work for buildings with slope rooftops or other overhangs, see commit e9db222);
2. directly construct polygons with correct normals (see commit 7270933). 

Please help me remove compas_occ from dependency if the tests passed. We can also discuss about it later this week :)

reyery:
I have a feeling that this might be a very specific case where more sensors on edge benefits this particular shading pattern (i.e., floating building right above, only areas close to edge have higher irradiation). So even doing the rotation strategy, as long as it does not have enough sensors on the edges, the results will not match the high resolution results. 

The only way to "fix" this would be to always line the edges with sensors, but this might be hard to reason with since it might introduce logic to solve what I would consider an edge case. So on my end the 2 questions to be considered is:
-  Is introducing the edge sensors logic difficult to implement and how much does it increase the complexity and computational cost.
- Since it is only for one particular building in the scenario, how much of the total/overall irradiation is affected by this difference.

PR #3938, opened 2025-10-21
Add demand calculation checks

reyery:
Add guards for division by zero and checks for physical limitations.


reyery:
@martin-mosteiro could you help check if these validations and their error messages are correct.

PR #3941, opened 2025-10-23
Ensure building nodes are not merged in network

reyery:
## Summary by CodeRabbit

* **API Changes**
  * Renamed parameter/attribute: `tolerance` → `coord_precision`.

* **New Features**
  * Graph corrections now run earlier (before terminal creation) and accept protected nodes (e.g., building terminals).
  * Networks enforce a canonical CRS and reproject grids when needed.

* **Bug Fixes**
  * Avoid merging or connecting two protected nodes.
  * Use dynamic minimum segment length when splitting lines; warn on disconnected graphs after correction.

* **Tests**
  * Updated tests to reflect `coord_precision` rename and protected-node behavior.

### DIFF
--- a/cea/technologies/network_layout/connectivity_potential.py
+++ b/cea/technologies/network_layout/connectivity_potential.py
@@ -1,10 +1,34 @@
 """
-This script uses libraries in shapely to create connections from
-a series of points (buildings) to the closest street
+Network connectivity module for connecting buildings to street networks.
+
+This module creates a graph of potential thermal network paths by connecting building centroids to the nearest
+points on a street network, preparing the data for Steiner tree optimization.
+
+WORKFLOW:
+=========
+1. Load and validate street network
+2. Apply GraphCorrector to street network (fixes connectivity, intersections, merges close nodes)
+3. Create building terminal connection lines (geometries from building → nearest street point)
+4. Topologically merge and connect geometries:
+   - one_linestring_per_intersection: Merge touching lines, split at intersections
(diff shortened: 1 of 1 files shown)


reyery:
2. I think `protected-node` was added because there was an error regarding missing terminal nodes I saw from student error logs.
```text
Graph validation failed before Steiner tree: 14 terminal nodes not found in graph. This should not happen after corrections in _load_pot_network(). Please report this issue.
```
But looking at it now, it could be mainly due to the CRS differences instead. I can double check that again if remove it if it is not needed.

3. Yes, ideally we should standardise how we read files, and not just for shapefiles. But because all the scripts were written in silo, there was never a standardised reading standard. Daren did try to do it before by making the input locator do the reading/writing as well, but he never completed it. I have some ideas to continue this, but I do not have the time yet.

But for shapefiles, I think the rule of thumb now is to always follow the CRS of the zone buildings.

reyery:
@MatNif I just check and during the steiner tree validation it is checking if the building nodes exist in the network. So I am assuming that building nodes are part of the network?

https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/a3141ce7b29f144899ff0556457ddeaebe30e704/cea/technologies/network_layout/steiner_spanning_tree.py#L118-L128

https://github.com/architecture-building-systems/CityEnergyAnalyst/blob/9c00a0a053303835d14a4a167e3f58a9892ddb2d/cea/datamanagement/graph_helper.py#L871-L875

PR #3942, opened 2025-10-23
Optimisation part 2 to work on a subset of buildings by user choice

ShiZhongming:
## Summary by CodeRabbit

* **New Features**
  * Large expansion of optimization settings: select buildings, cooling/heating/heat-rejection technologies, energy sources/potentials, network temperature, model complexity, and detailed output/retention options
  * New algorithm and GA controls for tuning optimization (algorithms, mutation/crossover methods, population, generations, probabilities)

* **Bug Fixes**
  * Reset of optimization state between runs to prevent carryover
  * More robust coordinate handling and network graph construction for improved network accuracy and building/potential loading

### DIFF
--- a/cea/optimization_new/containerclasses/energyPotential.py
+++ b/cea/optimization_new/containerclasses/energyPotential.py
@@ -156,7 +156,7 @@ class EnergyPotential(object):
         else:
             return None
 
-    def _get_building_potentials(self, energy_potential_files, building_codes, main_potential_column_name,
+    def _get_building_potentials(self, energy_potential_files, building_codes: list[str], main_potential_column_name,
                                  temperature_column_name=None, auxiliary_potential_column_name=None):
         """
         Gets main and auxiliary potentials from the stored energy potential files and stores them in the corresponding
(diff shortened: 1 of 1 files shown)


cea/optimization_new/network.py
```diff
@@ -393,6 +393,13 @@ def calculate_operational_conditions(self, connectivity_string_for_files):
 
     @staticmethod
     def initialize_class_variables(domain):
+        # Clear class variables before reinitializing to prevent state persistence across runs
+        Network._domain_potential_network_graph = nx.Graph()
+        Network._domain_potential_network_terminals_df = pd.DataFrame()
+        Network._domain_buildings_flow_rate_m3pers = pd.DataFrame()
+        Network._domain_buildings_supply_temp_K = pd.DataFrame()
+        Network._domain_buildings_return_temp_K = pd.DataFrame()
```
reyery: @ShiZhongming were the students experiencing problems with repeated values?

cea/optimization_new/network.py
MatNif, line 1: I think the changes in this file are actually all entirely unnecessary. I don't think they will break the code per say, either. So it really depends on what style people prefer. I prefer the former, but I guess I'm biased since I implemented it originally.

cea/optimization_new/domain.py
MatNif, line 1: Not sure, why the conversion to a pd.Series is necessary for the buildings_in_domain object. But the handling of the new optimization_new.buildings parameter makes sense.

cea/optimization_new/districtEnergySystem.py
MatNif, line 1: The way we run scripts in CEA, the state never persists across runs. So this changes is not necessary. But I think an argument can be made that this might be good for "future-proofing" our code.

cea/default.config
MatNif, line 1: Good and essential change for what this branch is trying to achieve.

MatNif (commented):
The new buildings parameter for the centralized optimization script makes sense and is implemented well.

However, there are also quite a few unnecessary changes in this branch. I think these are mostly based on the assumption that CEA retains information in memory between script runs. With how CEA launches scripts that is actually never the case.

The changes made as a consequence don't really impact the functionality of the code, but I would still remove them since I think they're mostly just "clutter". It's really dependent on personal preference in the end though. @reyery What do you think of these code changes?

cea/optimization_new/network.py
reyery: Though it is not entirely necessary for it to work, design wise it is better to use classmethod instead of staticmethod when the method modifies the class state. This has clearer intent compared to using staticmethod, which are mainly for utility methods.

I think eventually I would move it from using class variables to using composition since mutable class variables are hard to keep track. And that would also solve the "reinitialise state" issue.

PR #4011, opened 2026-03-16
Stage bundled Windows runtimes before launching CRAX binaries

yiqiaowang-arch:
## What

Make the CRAX launcher stage `mesh-generation.exe` / `radiation.exe` with the Pixi-bundled runtime DLLs before launching on Windows.

## Why

`radiation-crax` failed on my machine during mesh generation with exit code `3221225477` (`0xC0000005`).
Windows Error Reporting showed the crash came from `C:\Windows\System32\MSVCP140.dll`.

This makes the external CRAX executable depend on machine-specific DLL resolution instead of the bundled environment runtime.

## How

- add a Windows execution-context helper in `CRAXModel.py`
- copy the CRAX executable and matching bundled runtime DLLs into a temporary runtime folder before launch
- use the same helper for both mesh generation and radiation
- add a regression test for the staging logic

## Summary by CodeRabbit

* **Documentation**
  * Added API documentation for CRAX operations and runtime configuration patterns.

* **Bug Fixes**
  * Improved Windows runtime dependency handling for CRAX executables to ensure proper library resolution and execution.
  * Enhanced error reporting for mesh generation and radiation computations.

* **Tests**
  * Added test coverage for Windows runtime staging functionality.

### DIFF
--- a/cea/resources/radiationCRAX/AGENTS.md
+++ b/cea/resources/radiationCRAX/AGENTS.md
@@ -1,0 +1,32 @@
+# radiationCRAX
+
+## Main API
+- `CRAX.run_mesh_generation(json_file: str) -> str` - Run `mesh-generation` with staged native runtimes.
+- `CRAX.run_radiation(json_file: str) -> str` - Run `radiation` with the same runtime-staging rules.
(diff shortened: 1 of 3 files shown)


reyery:
Closing PR as static linking for CRAX binaries should have solved this.

PR #4033, opened 2026-04-22
Fix zone helper issue

martin-mosteiro:
Certain buildings are imported with a floor-to-floor height of 1 (or even as a 1-floor building with the height as the number of floors). See the screenshots below: OSM correctly lists the building as 7 floors tall, but the building height is also 7m in the CEA model.
image
image

This appears to be caused by the following line:
```
shapefile[[c for c in OSM_COLUMNS if c in list_of_columns]] = \
            shapefile[[c for c in OSM_COLUMNS if c in list_of_columns]].fillna(1) \
                .apply(lambda x: pd.to_numeric(x, errors='coerce'))
```

Not sure why we'd need to fill in values with 1's instead of just leaving them as NaN...

I have also changed the following line:
```
shapefile["floors_ag"] = [int(x) if not np.isnan(x) else data_osm_floors_joined for x in
                                  data_osm_floors_sum_with_nan]
```
since it just filled in all floor NaN values with the median – but sometimes we have the height and not the floors, so you could get an unreasonable building height/number of floors combination. I have therefore changed it as follows:
```
        # fill in missing data in OSM files with assumed floor-to-floor height
        shapefile['building:levels'] = shapefile['building:levels'].where(
            ~shapefile['building:levels'].isna(), shapefile['height'].astype(float) / constants.H_F)
        shapefile['height'] = shapefile['height'].where(
            ~shapefile['height'].isna(), shapefile['building:levels'].astype(float) * constants.H_F)

        shapefile["floors_ag"] = [int(x) if not np.isnan(x) else data_osm_floors_joined for x in
                                  shapefile['building:levels'] + shapefile['roof:levels'].fillna(0)]
```
Where you first fill in missing OSM data where available, then fill in missing floors based on OSM data

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing building height and level data from OpenStreetMap sources with enhanced numeric field processing and bidirectional imputation between floor counts and height values.

### DIFF
--- a/cea/datamanagement/zone_helper.py
+++ b/cea/datamanagement/zone_helper.py
@@ -58,6 +58,7 @@ def parse_building_floors(floors):
 
 def assign_attributes(shapefile, buildings_height, buildings_floors, buildings_height_below_ground,
                       buildings_floors_below_ground, key):
+    N_FLOORS_ASSUMPTION = 3     # number of floors assumed when no information is available
     # local variables
     no_buildings = shapefile.shape[0]
     list_of_columns = shapefile.columns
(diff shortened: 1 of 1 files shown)


ShiZhongming:
It worked as intended, fixed the problem described. Thanks @martin-mosteiro 
We should merge it when the test passed

Issue #4074, opened 2026-09-21
PV yield differs across macOS/Windows: environment-dependent sensor mesh amplified by the radiation threshold


ShiZhongming:
## Summary

Three separate effects, of very different sizes. The first caused the original report; the other two are inherent and remain.

| # | Effect | Size | Status |
|---|---|---|---|
| 1 | Different `annual-radiation-threshold` on the two machines (0 vs 800) | **+49% yield** | User config, resolved |
| 2 | Sensor mesh differs by environment (OCC tessellation) | ~115 kWh/m²/yr on affected sensors | Dependency-level, partly fixable |
| 3 | Hard threshold turns small radiation differences into whole-sensor area jumps | −0.4% district, up to −4.4% per building, −24% on one facade | Design choice, worth discussing |

Scenario: `idp26_STU0921_zhishuang_tutorial`, 7 buildings, 1727 sensors, PV1–PV4.

## Finding 1 — the 49% gap was config, not code

`~/cea.config` on the Mac had `annual-radiation-threshold = 0.0`; the Windows run used 800. The setting is per-machine (user config), not stored with the scenario, so a value set in the GUI on one machine never reaches another.

Evidence: minimum `total_rad_Whm2` among selected sensors was 0 on Mac (1537 sensors kept) and 804,195 Wh/m² on Windows (301 kept). Roofs were identical in both (58 sensors, 832.24 m²); the entire gap was walls, 1479 vs 243.

After re-running the Mac at 800, district totals agree to **−0.37% / −0.39% / −0.39% / −0.38%** for PV1/PV2/PV3/PV4, with installed area 2411.0 vs 2419.2 m².

## Finding 2 — the sensor mesh is environment-dependent

The `*_geometry.csv` sensor grids are not identical between environments. Invariants are preserved exactly — same sensor count, same total area (B6096: 2486.815 m² everywhere), identical sorted patch-area list — but individual patches sit in different places.

Regenerating the grid locally (pythonocc 7.9.0) and comparing point sets:

| Building | Local vs Mac | Local vs Windows |
|---|---|---|
| B2019 (611) | 611/611 identical | 0/611 |
| B2021 (116) | 116/116 | 113/116 |
| B4027 (223) | 219/223 | 207/223 |
| B6096 (422) | 407/422 | 397/422 |

**It is deterministic per environment:** two consecutive local runs on the same machine are bit-identical across all 7 buildings. So this is not randomness — it is the geometry kernel.

[...]

PR #4078, opened 2026-09-21
Add radiation determinism experiment (within and across OSes)

reyery:
## Summary

Experiment for #4074, separate from regular CI. It separates DAYSIM binary differences from OCC mesh differences, following the findings in #4075 (meshes differ across OSes, and identical-position sensors still differ by about 0.3% in radiation).

`.github/workflows/radiation-determinism.yml` builds the `zug_heating` scenario up to radiation, then:

- **A. Same OS, full radiation script twice** (OCC geometry + DAYSIM): compares intermediate files (radiance geometry/material, sensor points, sensor grid CSVs) and every sensor's hourly results bit for bit.
- **B. Same OS, DAYSIM twice on identical input files**: ubuntu generates the DAYSIM inputs once (`prepare`); each OS replays `epw2wea`, `radfiles2daysim`, `gen_dc`, `ds_illum` on them.
- **C. Different OSes, DAYSIM on those same inputs**, compared against ubuntu. Converted DAYSIM inputs and `.dc` daylight coefficient files are hashed to show which step first differs (`.dc` hashes are best effort, the files may embed paths).

The `compare` job writes a summary table to the run page. Code is in `.github/scripts/radiation_determinism/`; the truncated scenario workflow lives there too so the integration tests do not pick it up.

## How to read the result

- A or B not bit-identical: noise exists within one OS, so it is not a cross-OS effect.
- A and B identical, C not: the DAYSIM binaries differ across OSes. The stage hashes show whether it is `radfiles2daysim`, `gen_dc` (`.dc`) or `ds_illum`.
- A and B identical, C identical: DAYSIM is fine and the cross-OS difference comes from the scene/mesh that OCC produces.

## Testing

Ran the whole flow locally on macOS (single OS): A and B were bit-identical (2662/2662 sensors), C empty as expected. The workflow itself has not run on GitHub yet, so runner paths and the Windows steps are unverified. It runs on this PR because of the `pull_request` path trigger (`workflow_dispatch` needs the file on the default branch); expect roughly 60-90 minutes for the slowest OS.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Added automated checks to verify consistent radiation-processing results across repeated runs and operating systems.
  * Added comparisons for intermediate outputs, sensor results, and annual radiation values.
  * Added replay testing for prepared DAYSIM inputs.

* **Chores**
  * Added a workflow to prepare the Zug baseline scenario and run determinism checks on Ubuntu, macOS, and Windows.
  * Added reporting of comparison results in the CI summary.


reyery:
Closing without merging: this was an experiment and its findings are written up in #4080. The branch `ci-radiation-determinism-experiment` is kept so the workflow can be revived (rebased and merged) if we test a DAYSIM build with deterministic floating-point flags.

